PR #10, opened 2022-01-17
update great_ape_safe from main repo, add badger system

jorijnsmit:
first step towards adding badger system as a greatapesafe class


petrovska-petro (commented):
I think that in all the classes inside of the directory `ape_api/..` will be nice in their init section to have define the contracts they are pointing to grab directly from the address book as you had done for badger class

```
def __init___
```

jorijnsmit:
> I think that in all the classes inside of the directory `ape_api/..` will be nice in their init section to have define the contracts they are pointing to grab directly from the address book as you had done for badger class
> 
> ```
> def __init___
> ```

i agree this makes sense for the multisig repo, considering we only want 0x vars in the addresses.py.

lets create a ticket for that though, since it falls outside of the scope of this pr.

petrovska-petro:
> > I think that in all the classes inside of the directory `ape_api/..` will be nice in their init section to have define the contracts they are pointing to grab directly from the address book as you had done for badger class
> > ```
> > def __init___
> > ```
> 
> i agree this makes sense for the multisig repo, considering we only want 0x vars in the addresses.py.
> 
> lets create a ticket for that though, since it falls outside of the scope of this pr.
> 
> edit: created ticket at #15

Sounds good to handle it into its own ticket 👍

petrovska-petro (dismissed):

scripts/issue/9/claim_ibbtc_rewards.py
```diff
@@ -0,0 +1,21 @@
...
+    bcvx = interface.ISettV4h(registry.eth.treasury_tokens.bCVX)
+    bvecvx = interface.ISettV4h(registry.eth.treasury_tokens.bveCVX)
+    bcvxcrv = interface.ISettV4h(registry.eth.treasury_tokens.bcvxCRV)
+
+    ibbtc_msig.take_snapshot(tokens=[
+        bcvx.address, bvecvx.address, bcvxcrv.address
+    ])
+
+    ibbtc_msig.init_badger()
+    ibbtc_msig.badger.claim_all_from_tree()
```
sajanrajdev, line 19: There doesn't seem to be a `claim_all_from_tree()` method on Badger. Shouldn't it be `claim_all()`?

scripts/issue/9/claim_ibbtc_rewards.py
```diff
@@ -0,0 +1,21 @@
...
+
```
sajanrajdev, line 20: Didn't you meant to add a printout for the snapshot here `ibbtc_msig.print_snapshot()`?

scripts/issue/9/claim_ibbtc_rewards.py
jorijnsmit: good catch, i renamed that to match the signatures we use for the other classes, forgot to change it here.

scripts/issue/9/claim_ibbtc_rewards.py
jorijnsmit: `post_safe_tx` does a printout as well at the end.

scripts/issue/9/claim_ibbtc_rewards.py
jorijnsmit: #d107eef

jorijnsmit:
converted to draft until #22 is merged (will conflict)

jorijnsmit:
ok merged main, conflicts resolved.

jalbrekt85 (approved):
`api_url` seems to work well, expected token balances are increased after calling `claim_all`

petrovska-petro (approved):
Confirmed the right naming of the method called and tested also with the dev_msig for another sample.

```
snapshot result for 0xB65cef03b9B89f99517643226d76e286ee999e77:
                   balance_before              balance_after              balance_delta
symbol                                                                                 
bCVX         0.000000000000000000     816.419200067112707272     816.419200067112707272
bveCVX     852.098087588164863839   3,626.959065662560452333   2,774.860978074395588494
bcvxCRV  4,957.223934273965294772  22,950.316753449797467040  17,993.092819175832172268 
```

Also includes the PR which brought to the classes the feature of pulling all the addresses directly from the book instead of hardcoding them

Issue #13, opened 2022-01-17
week3: process expired unlocks on bvecvx vault


jorijnsmit:
1: https://github.com/GalloDaSballo/vested-cvx/blob/b475e7fab0913f89d8406b02b8f1caaf5afc3f37/tests/test_roleplay_unlock.py#L84
2: `strat_proxy.manualSendCVXToVault({"from": governance})`

-> Earn every Wednesday at 23:47 -> All funds in vault are locked
-> Unlock on Thursday (manual operation may not be able to estimate by the minute) -> All funds unlocked are sent to vault
-> This means we give the 6 days + unlock window we agreed on with users

jorijnsmit:
think deadline is next week thursday, but having the script tested and reviewed would be nice to have this week.

jorijnsmit:
posted nonce 987, https://gnosis-safe.io/app/eth:0xB65cef03b9B89f99517643226d76e286ee999e77/transactions/0x0c0ef1b818633a5958dc8ea6dad3e0b79d0fa1524bdc78c3a67435829e6eb84b :
```
snapshot result for 0xfd05D3C7fe2924020620A8bE4961bBaA747e6305:
              balance_before              balance_after              balance_delta
symbol                                                                            
CVX     0.006295640935761957  70,778.570845258333457529  70,778.564549617397695572 
```

Tritium-VLK:
**Note that this ticket should happen at a predictable time.  @wasabiboat should put out comms in the necxt hour stating this will happen between Noon and 1PM EST on Friday.**

Please try to ensure it happens at this time.  I am working with @GalloDaSballo to reduce the security requirements on calling this function, but for now we will need to time this to happen predictably at the same time each week (fridays at noon for now)

jorijnsmit:
https://etherscan.io/tx/0x2f8351174a10f19156b88b6674c7030336b5f4cf49a947aaca7836fb6e4f6ad9

PR #16, opened 2022-01-18
call vested cvx strat to process expired vlcvx

jorijnsmit:
script for #13 

we should consider integrating this into greatape instead of having it as a script, since it is a recurring action (will have to run every week, see https://dune.xyz/tianqi/Convex-Locked-CVX for upcoming unlockings).


petrovska-petro (approved):
Snap on vault:

```
snapshot result for 0xfd05D3C7fe2924020620A8bE4961bBaA747e6305:
              balance_before          balance_after          balance_delta
symbol                                                                    
CVX     0.008809586569433787  20.008809586569433787  20.000000000000000000 
```

Unlocks 20 CVX which as per the dune query is correct, ref:

Screenshot 2022-01-18 at 14 28 26

GalloDaSballo (approved):
The script will:
-> Unlock all expired locks
-> Move all unlocked CVX from the Strategy to the Vault

NOTE: Running the simulation now shows 20 unlocked CVX, but on Thursday the same simulation will unlock 70758.6 + the 20

Issue #23, opened 2022-01-21
write general purpose script that grabs all balances for all tokens...


jorijnsmit:
... from a msig and dumps it to a airdrop csv format.

this can help in sweeping a wallet or grouping multiple transfers into one.

jorijnsmit:
solved by #25

PR #25, opened 2022-01-21
scrape all token balances for multi sig wallets

jalbrekt85:
**script for #23**

~~- `brownie-config` updated so that api keys can be referenced from `brownie.config` inside scripts~~
~~- uses ethplorer api to get all tokens and their balances~~
~~- `token_address` and its `value` are saved into csv. not sure if any other data is needed~~

- scrape function now adheres to the airdrop format
- queries token balances from `treasury_tokens` instead of external api
- specify receiving address
- progress bar and table

script can be run through command line to scrape balances with supplied address:
`brownie run scripts/gnosis/scrape_tokens.py scrape <address> <receiver>`


scripts/issue/23/scrape_tokens.py
```diff
@@ -0,0 +1,39 @@
+from brownie import config
+import requests
+import pandas as pd
+
+def main():
+    """
+    Get all ERC-20 tokens and their balances in a csv for a given address with ethplorer api
+    
+    `scrape` can be called directly from command line with an address:
+    brownie run scrape_tokens.py scrape <address>
```
jorijnsmit, line 10: since this is a nice tool we will reuse, i suggest to create a folder such as `scripts/gnosis/` and place it there. it will be very difficult to find again in `scripts/issue/.../` in a couple of weeks/months.

scripts/issue/23/scrape_tokens.py
```diff
@@ -0,0 +1,39 @@
...
+    for token in res.json()['tokens']:
+        try:
+            if int(token['tokenInfo']['decimals']) > 0:
+                token_data['token_address'].append(token['tokenInfo']['address'])
+                token_data['value'].append(token['rawBalance'])
+        except KeyError:
+            # Non-ERC20 token
+            pass
+        
+    pd.DataFrame(token_data).to_csv(f'scripts/issue/23/{address}_token_balances.csv')
```
jorijnsmit, line 38: i probably should have specified; the gnosis airdrop csv has the following (strict) format:
```
token_type,token_address,receiver,value,id
```
> Transfer files are expected to be in CSV format with the following required columns:
>
>    token_type: The type of token that is being transferred. One of erc20,nft or native. NFT Tokens can be either ERC721 or ERC1155.
>    token_address: Ethereum address of ERC20 token to be transferred. This has to be left blank for native (ETH) transfers.
>    receiver: Ethereum address of transfer receiver.
>    value: the amount of token to be transferred. This can be left blank for erc721 transfers.
>    id: The id of the collectible token (erc721 or erc1155) to transfer. This can be left blank for native and erc20 transfers.
>
> Important: The CSV file has to use "," as a separator and the header row always has to be provided as the first row and include the described column names.

have a look at 6d6a95fe981f880391dadd68e28de6dce9e91309

my idea was then to leave `receiver` empty (or optional parameter for the func?).

simple use case is sending all dust on techops to treasury ops for processing, all in one go.

scripts/issue/23/scrape_tokens.py
```diff
@@ -0,0 +1,39 @@
...
+
+def scrape(address):
+    with requests.session() as session:
+        ethplorer_key = config['keys']['ETHPLORER_API_KEY']
+        api_key = ethplorer_key if ethplorer_key else 'freekey'
+        res = session.get(
+            f'https://api.ethplorer.io/getAddressInfo/{address}?showETHTotals=false', 
+            params={"apiKey": api_key}
+            )
+        assert res.status_code == 200
```
jorijnsmit, line 26: we rather not rely on external apis if not needed; a `interface.ERC20(token_addr).balanceOf(addr)` would be a lot more reliable (as balances are all taken verifiable directly from one and the same forked block). personally i would loop over `registry.eth.treasury_tokens`, although i know that that is not fully waterproof in the sense of being all inclusive..

jorijnsmit (changes requested):

jorijnsmit (approved):
- good use of `Decimal`
- i asked a dev on gnosis discord ([message url](https://discord.com/channels/907968493134155807/908129717507461130/935105166460477450)) about the `decimals` column being needed in the csv, but from testing current output in the app it doesnt seem like it poses a problem to include it.
- i didnt know about `Progress`; nice!

jorijnsmit:
@jalbrekt85

in response to https://discord.com/channels/907968493134155807/908129717507461130/935105166460477450 :
Screenshot 2022-01-24 at 15 05 32

PR #48, opened 2022-01-26
port gpv2 class from badger-ape using rebranded urls

jorijnsmit:
gpv2 is now cow protocol, edited api urls and naming to reflect this.


petrovska-petro (approved):
- Porting of cowswap class from prev repo, adding new urls ✅ 

- Script do multi approve for all tokens claimable from bribes to allow the relayer to do their task ✅ 

- Locks cvx and deposits directly into tree ✅

sajanrajdev (approved):
Ports over gpv2 with cow swap urls.

Issue #26, opened 2022-01-24
week 4: claim bribes and process into bvecvx emissions


jorijnsmit:
- claim votium bribes from strat
- claim convex bribes from strat
- swap all for cvx
- lock/swap for bvecvx (depending on health of https://curve.fi/factory/52)
- calc proper amounts and send to tree (refer to calcs from last round: https://docs.google.com/spreadsheets/d/1an-LWDguHSXF1D_OlhJTANRnYZlkpl6HwCYEkYpPNNk/edit#gid=820692909)

jorijnsmit:
notes from @GalloDaSballo :

Claim CVX bribes from Votium

## Why
CVX Bribes that couldn't be claimed in the past can now be claimed

## Technical
See Code from this test:
https://github.com/GalloDaSballo/vested-cvx/blob/6c5cbc84464fbfef4646bd702bb2e9b211668a1d/tests/test_claim_bribes.py#L150

## Specifically
```
TOKEN = "0x4e3FBD56CD56c3e72c1403e103b45Db9da5B9D2B"
INDEX = 684
AMOUNT = "0x6e4a3cd203a5a40000"
PROOF = [
    "0xde24992959a35e41f92dc070000fb59ade0b47f2399b2cbe7ec0126accc9f832",
    "0xe965287182073f2db7f8507458c9a1e168200f7973c155f5432e49dc51d80ed2",
    "0x53fba95706bb8bcfeb337fffb943c4ea708425185e2931139fdfe0e3f96ae49a",
    "0x9817ff39ca612d9b62f61f8e266df6a8d67b5406e339be58dd832e324851479e",
    "0x9fc0b7dc83afc55158b60d5709091348c4bc6f9a5263af5d42c5e70437a6db9e",
    "0xcbff70570ca71e83dbd2bdbd254c916a629092a73fd148675f55fdfb1ae85c4d",
    "0x0f53cd12926a48df89913df901b24def5d7b74f90b1e862f0103b575fcb7b0d2",
    "0x78266a424a8faf8964c0ad8ef984a56bb23daa585ce45ce549c72891f67c5619",
    "0x0395908c13baf6cf4dfb9ddba61bc7e02567fc186a4f8243ac970af4c290c634",
    "0x93fe0d6bdabadd863667aaa36cdbd1c48483b71ef903088ac2fca69964f1291e",
    "0x92313245368194a1e13309d50a3b382ea5185547fddd2651159a85936c8a7182"
]

[...]

PR #49, opened 2022-01-26
bveCVX yield redirection 26.01.2022

Tritium-VLK:
Sets a 15 day emissions schedule for bveCVX.  Amount still needs to be set based on sale.  This 15 days is one time meaning that going forward the emissions will run from 00:00 UTC FRIDAY giving us more time.

See #26


scripts/emissions/bveCVX_special_emissions.py
```diff
@@ -0,0 +1,29 @@
+from great_ape_safe import GreatApeSafe
+from helpers.addresses import registry
+
+DEVELOPMENT = False
+
+
+
+# UPDATE THESE VARIABLES WITH THE NEEDS FOR A NEW GUARD LAUNCH OR SIMILAR SCENARIO
+STARTING_TIME = 1643241600
+ENDING_TIME = 1644534000
```
jorijnsmit, line 10: this translates to `datetime.datetime(2022, 2, 10, 23, 0)`?

scripts/emissions/bveCVX_special_emissions.py
```diff
@@ -0,0 +1,29 @@
...
+AMOUNT_TO_EMIT_DURING_PERIOD = 100
+DURATION = ENDING_TIME - STARTING_TIME
+
+
+def main():
+    safe = GreatApeSafe(registry.eth.badger_wallets.dev_multisig)
+    safe.init_badger()
+    rewards_logger = registry.eth.rewardsLogger
```
jorijnsmit, line 18: think this should be set from techops, no?

also `rewards_logger` here is just an address string, still needs to be initiated as a contract object (`safe.contract(...)` is fastest for now).

jorijnsmit (changes requested):

scripts/emissions/bveCVX_special_emissions.py
jorijnsmit: or is the idea here to have 1 hour where nothing happens..?

Tritium-VLK: No.  Just forgot to adjust for timezones at it was midnight CET.

scripts/emissions/bveCVX_special_emissions.py
Tritium-VLK: correct ser.

Tritium-VLK:
@jorijnsmit discord is down, but needs to be set to amount of bveCVX in the poli multisig after swapping all swapped cvx for bveCVX.

Tritium-VLK:
https://docs.google.com/spreadsheets/d/1R3dBig3f13jNgjTuNhLXcGVle1gxERNNKJD-2TAWgbE/edit
image

jorijnsmit (approved):
amount is from bvecvx which was in politician + newly locked:
```
>>> from decimal import Decimal
>>> Decimal(7151.824281447517145778) + Decimal(4862.422279384771517847)
Decimal('12014.24656083228819625219330')
```
timestamps; adds one more day for buffer in case bribes are delayed:
```
>>> from datetime import datetime
>>> datetime.fromtimestamp(1643241600)
datetime.datetime(2022, 1, 27, 0, 0)
>>> datetime.fromtimestamp(1644537600)
datetime.datetime(2022, 2, 11, 0, 0)
```

petrovska-petro (approved):
LGTM!

- confirmed timestamps ending on 11th Feb 00:00 UTC
- confirmed figure from `depositFor` txs: 

https://etherscan.io/tx/0x1f15288f7b0d1dd355e7379e430fbd466f434c2b2e4635a59bc76684fc43370e
https://etherscan.io/tx/0x17d98ee2a40d670f509d67b07e2a83d059bc946eeffa63ec060a925d3d268232

also tx history from politician in cow: 

https://gnosis-protocol.io/address/0x6F76C6A1059093E21D8B1C13C4e20D8335e2909F

petrovska-petro:
@Tritium-VLK looks like not all commits were signed 💀

jorijnsmit:
forced merge instead of creating new pr.

Issue #31, opened 2022-01-24
upgrade bveCVX to 1.4


jorijnsmit:
Upgrade bveCVX Strategy to Version 1.4

## Why
Allows claiming of CVX bribes from both governance and strategist

## Technical
https://github.com/GalloDaSballo/vested-cvx/blob/main/tests/test_upgrade_1_4_strategist_bribes_claim.py

## Specifically
`proxy_admin.upgrade(strat_proxy, new_strat_logic, {"from": proxy_admin_gov})`

Logic to be deployed and verified shortly

## Urgency

The earlier the better so bribes can be claimed, I think Thursday this week is great

GalloDaSballo:
Verified Code:
https://etherscan.io/address/0x56df16192D586b288F53852466560b445E21939b#code

GalloDaSballo:
@sajanrajdev @shuklaayush Please verify the code is as expected 

Security Review was already passed:
https://github.com/Badger-Finance/badger-strategies/issues/33

sajanrajdev:
@GalloDaSballo Compared verified deployed logic to latest version on the repo and they are identical. LGTM

![image](https://user-images.githubusercontent.com/25463788/151223165-020d147b-2346-43aa-a801-1fdf82b3bbbf.png)

jorijnsmit:
https://etherscan.io/tx/0xca5530fc4ce7fc27a67714a877296622ee2ef7db41c3b36a0e170a408cfaae7b#eventlog

sajanrajdev:
Timelock execution call trace: 
```
Running 'scripts/issue/31/upgrade_vestedcvx_strat.py::main'...
Executing tx with parameters: {'data':
'0000000000000000000000003ff634ce65cdb8cc0d569d6d1697c41aa666cea900000000000000000000000056df16192d586b288f53852466560b445e21939b', 'eta': 1643640073,
'eth': 0, 'signature': 'upgrade(address,address)', 'target': '0x20Dce41Acca85E8222D6861Aa6D23B6C941777bF'}
Transaction sent: 0x820a1eeed822dc8c703b3c63df2b283b4c33f285f339c391b75aafeb40f19719
  Gas price: 0.0 gwei   Gas limit: 12000000   Nonce: 1
  GovernanceTimelock.executeTransaction confirmed   Block: 14114691   Gas used: 35811 (0.30%)

[...]

PR #53, opened 2022-01-27
feat: vestedcvx upgrade payload data

sajanrajdev:
This PR relates to issue #31 

Adds the timelock queue payload data for execution.


jorijnsmit (approved):
calldata checks out, timestamp too.

petrovska-petro (approved):
Confirmed calldata and timestamp against nonce 990 queued in dev msig

Issue #34, opened 2022-01-25
increase depth of our curve v2 factory pool


jorijnsmit:
steps:
- get 3.5 wbtc + ~15k badger to techops. withdrawing from bibbtc position on the vault might already be done by #32 . can use `MintableForkToken` for testing first
- deposit fraction of this, ie 1 wbtc + equal parts badger
- monitor diff between price oracle and price scale
- swap or withdraw at unequal parts?
- repeat until calibrated

dev: not sure on func sigs for the v2 pool compared to the old style. prob best to get interfaces for the new pool/token and maybe add `curve_v2` class to great ape if needed.

deadline: the pool should consistently trade within a few percent of the market price by the next gauge vote that starts on February 2nd.

ref: https://curve.fi/factory-crypto/4

jorijnsmit:
https://gnosis-safe.io/app/eth:0x042B32Ac6b453485e357938bdC38e0340d4b9276/transactions/0x5108a3b55f7814b3c4bcf6322bff960c8ac571d82531a3567b3a6b0a6e1aaea0
```
Events In This Transaction
--------------------------
├── Curve.fi Factory BTC Metapool: ibBTC (0xFbdCA68601f835b27790D98bbb8eC7f05FDEaA9B)
│   └── (unknown)
│       ├── topic1: 0x5ad056f2e28a8cec232015406b843668c1e36cda598127ec3b8c59b8c72773a0
│       ├── topic2: 0x0000000000000000000000007abdbaf29929e7f8621b757d2a7c04d78d633834
│       └── data: 0x0000000000000000000000000000000000000000000000001b0b9895519b8d1a0000000000000000000000000000000000000000000000001acada6862d3015f000000000000000000000000000000000000000000000228c177e37e86122bfa
│   
├── IStableSwap (0x7fC77b5c7614E1533320Ea6DDc2Eb61fa00A9714)
│   └── (unknown)
│       ├── topic1: 0x9e96dd3b997a2a257eec4df9bb6eaf626e206df5f543bd963682d143300be310
│       ├── topic2: 0x0000000000000000000000007abdbaf29929e7f8621b757d2a7c04d78d633834
│       └── data: 0x0000000000000000000000000000000000000000000000001acada6862d3015f000000000000000000000000000000000000000000000000000000000ba10584
│   
├── ICurvePoolV2 (0x50f3752289e1456BfA505afd37B241bca23e685d)
│   └── (unknown)
[...]

PR #63, opened 2022-01-28
curve factory deposit script, curvev2, univ2, sushi classes

jalbrekt85:
- script addressing #34 to deposit into equal parts badger/wbtc into curve factory pool and monitors pool data. let me know if I have the right idea
- `greatapesafe` classes for curveV2, uniV2, and sushi
- ~~no withdraw functionality for curveV2 yet~~
- simple curvev2 class
- curve class updated to work with v2 interfaces


great_ape_safe/ape_api/curve.py
```diff
@@ -31,56 +32,75 @@ def deposit(self, lp_token, mantissas, asset=None):
         # TODO: find and route through zaps automatically
         # TODO: could pass dict to mantissas with {address: mantissa} and sort
         #       out proper ordering automatically?
-        pool_addr = self.registry.get_pool_from_lp_token(lp_token)
-        pool = self.safe.contract(pool_addr)
+        
+        # pool must be provided; cannot look up v2 factory pools from lp token
```
jorijnsmit, line 36: looks like `lp_token.minter()` should do the trick:
```
>>> from helpers.addresses import registry
>>> lp_token = interface.ICurveLP(registry.eth.treasury_tokens.badgerWBTC_f)
>>> lp_token.minter()
'0x50f3752289e1456BfA505afd37B241bca23e685d'
>>> registry.eth.crv_factory_pools.badgerWBTC_f
'0x50f3752289e1456BfA505afd37B241bca23e685d'
```

jorijnsmit (changes requested):
good use of the interfaces!!!

minor comments added inline.

great_ape_safe/ape_api/curve.py
jalbrekt85: forgot that method existed, very helpful

scripts/issue/34/issue34.py
```diff
@@ -0,0 +1,47 @@
...
+    safe.init_curve_v2()
+
+    if 'fork' in network.show_active():
+        WBTC = MintableForkToken(registry.eth.treasury_tokens.WBTC)
+        BADGER = MintableForkToken(registry.eth.treasury_tokens.BADGER)
+        WBTC._mint_for_testing(safe.address, 3.5 * 10**WBTC.decimals())
+        BADGER._mint_for_testing(safe.address, 15000 * 10**BADGER.decimals())
+    else:
+        WBTC = interface.ERC20(registry.eth.treasury_tokens.WBTC)
+        BADGER = interface.ERC20(registry.eth.treasury_tokens.BADGER)
```
jorijnsmit, line 35: this will need some adjusting

jorijnsmit (approved):
looking good @jalbrekt85 ! the issue34.py file itself we will have to adjust a bit (we always post from fork for example, so L28-L35 isnt going to work) but lets merge this since it provides lots of helpful functions already!

jorijnsmit:
@petrovska-petro your old code here being merged into greatape :)

petrovska-petro (approved):
Lots of goodies added into this PR!

Thanks for porting some of the methods of the old repo into here 🔥 

Talked with Gos, that probably the issue script itself will need some changes, rather than adding in one whole lump sum

PR #64, opened 2022-01-28
[CSV-generation]: Add script to be able to generate our pendant yields/fees/rewards

petrovska-petro:
Tackles #65 

- Working out a way to drop our claimable assets directly from the fork actual chain status to a CSV file for inspection

- Start portion univ3 into great ape, after this PR will bring the rest of the methods, wanted to keep it narrow down for this PR purpose

Testing command:

```
brownie run claimable_asset_csvs/claims_csv
```

CSV output for vault msig, minor problem with tree endpoint, reverting on `invalid cycle`:

```
token,claimable_amount
CVX,222.671360708843997871
CRV,1287.117691958635825851
BADGER,6961.242769653434709273
WBTC,2.08018343
```


great_ape_safe/ape_api/univ3.py
```diff
@@ -0,0 +1,50 @@
...
+    def __init__(self, safe):
+        self.safe = safe
+
+        # contracts
+        self.NonfungiblePositionManager = safe.contract(
+            registry.eth.uniswap.NonfungiblePositionManager
+        )
+        self.pool = interface.IUniswapV3Pool(
+            registry.eth.uniswap.v3pool_wbtc_badger, owner=self.safe.address
+        )
```
jorijnsmit, line 17: i suggest to rename this to `self.v3pool_wbtc_badger`

great_ape_safe/ape_api/univ3.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+class UniV3:
+    def __init__(self, safe):
+        self.safe = safe
+
+        # contracts
+        self.NonfungiblePositionManager = safe.contract(
+            registry.eth.uniswap.NonfungiblePositionManager
+        )
```
jorijnsmit, line 14: maybe introduce interface file for this?

great_ape_safe/ape_api/univ3.py
```diff
@@ -0,0 +1,50 @@
...
+        to allow us to claim the fees earned on each range over time
+        """
+        path = os.path.dirname(f"scripts/TCL/positionData/")
+        directory = os.fsencode(path)
+
+        token0 = Contract(self.pool.token0())
+        token1 = Contract(self.pool.token1())
+
+        wbtc_bal_init = token0.balanceOf(self.safe.address)
+        badger_bal_init = token1.balanceOf(self.safe.address)
```
jorijnsmit, line 34: if renaming these to `token0_bal_init` and `token1_bal_init`, the function makes more sense to be used on any uni v3 nfp.

great_ape_safe/great_ape_safe.py
```diff
@@ -90,7 +94,7 @@ def take_snapshot(self, tokens):
         self.snapshot = pd.DataFrame(df)
 
 
-    def print_snapshot(self):
+    def print_snapshot(self, csv_destination):
```
jorijnsmit, line 97: needs to be declared with a default value, eg `csv_destination=None`. otherwise all calls in existing code base (eg `safe.print_snapshot()` will fail since they do not provide a value for this (optional) argument.

great_ape_safe/great_ape_safe.py
```diff
@@ -45,6 +46,7 @@ def init_all(self):
         self.init_convex()
         self.init_curve()
         self.init_opolis()
+        self.init_univ3()
```
jorijnsmit, line 49: can i suggest `init_uni_v3` here, i just made the same comment in jalbrekt's pr.

i think it is important to be consistent with this, so always snake case `protocol_something_v5`, unless it is a class; `ProtocolSomethingV5`. this is common python style.

(i would even say the same for `self.NonfungiblePositionManager`; this is adopting solidity style in python. imo should be `self.nonfungible_position_manager` or `self.nfp_manager`)

jorijnsmit (changes requested):
smart use of print_snapshot, i see more possibilities there :)

great_ape_safe/ape_api/univ3.py
jorijnsmit: ah also `owner=self.safe.account`. sometimes this is needed in case some native (eth) calls are made to owner.

petrovska-petro:
Now the tree issue had being resolved, so it can grab the amount claimable from there for the vault and dev msigs 🥳 

dev:

```
token,claimable_amount
bcvxCRV,18062.326547035254292244
bveCVX,2795.791591385240677916
bCVX,816.419200067112707272
BADGER,164.244633607133173548
```

vault:

```
token,claimable_amount
stkAAVE,3.434529077081003375
CVX,242.084574499684314181
CRV,1424.026908821672436361
bcvxCRV,4339.377515491378419716
bveCVX,1086.577996858142440472
BADGER,7172.387202883565664455
WBTC,2.13733426

```

great_ape_safe/ape_api/univ3.py
petrovska-petro: good suggestion, added in last commit

great_ape_safe/ape_api/univ3.py
petrovska-petro: included the interface, from now on when creating the classes we are leaning towards always adding the interface?

great_ape_safe/ape_api/univ3.py
petrovska-petro: good one, just in case that we included more pools later on

great_ape_safe/great_ape_safe.py
petrovska-petro: pushed in last commit

great_ape_safe/great_ape_safe.py
petrovska-petro: noted, added your suggestion in last commit

great_ape_safe/ape_api/univ3.py
jorijnsmit: yea i think this is more robust and will prevent these 'invisible' brownie errors where a contract is loaded with the wrong interface. also i like researching the contracts and realising a bit better what is what.

jorijnsmit (approved):
🚀

jorijnsmit (approved):

sajanrajdev (approved):
- Adds a UNIV3 class to great ape with a fee collector function
- Adds a feature to the snapshot function to store a given snapshot in a CSV
- Adds a general script that claims all fees accumulated from all Vault/Dev positions and snapshots the results into a csv for each safe.

Vault:
```
token,claimable_amount
stkAAVE,3.653167803959258285
CVX,252.939266605397236735
CRV,1505.590872651174028189
bcvxCRV,4344.619396391233269182
bveCVX,1142.527037313226070071
BADGER,7359.794536637746973948
WBTC,2.17561379
```

Dev:
```
token,claimable_amount
bcvxCRV,18075.465897141760730306
bveCVX,2809.22100429784574711
bCVX,816.419200067112707272
BADGER,176.536786204805883766
```

NOTE: Any new position added to the stack will require us to modify the claims_csv.py script to ensure that it is covered.

LGTM

Issue #69, opened 2022-01-31
airdrop the 2000 minted rembadger tokens to 83 remaining victims


jorijnsmit:
https://docs.google.com/spreadsheets/d/14TGqwWn1pRr3px5ri5YbfLd-obiWuVIMERng9XflFQ0/edit#gid=109805477

jorijnsmit:
https://etherscan.io/tx/0x7ee58d63583924fb477dee567e1fc54343a162ce2b61f902fe2fc5cdf3c9fd91

PR #74, opened 2022-02-01
calc rembadger airdrop based on post mortem gsheet

jorijnsmit:
@Hal69k would be great if you could do a sanity check on the resulting csv vs your sheet.

issue #69


Hal69k:
Looks good. final row sums to 2000.0 and matches the addresses I have in my sheet.

Hal69k (commented):
Confirmed rows match and last column adds to 2000

Hal69k (dismissed):
Addresses match and final column sums to 2000.0

scripts/issue/69/calc_rembadger_airdrop.py
```diff
@@ -0,0 +1,63 @@
...
+
+    wbtc_rate = dff.at[1, 'value_usd_2dec'] / dff.at[1, 'value_wbtc']
+
+    # exception: convert $cvx value of single row to wbtc
+    cvx_idx = dff[dff['value_cvx'] > 0].index
+    dff.at[cvx_idx, 'value_wbtc'] = dff['value_usd_2dec'] / wbtc_rate
+
+    # exception: convert $weth value of single row to wbtc
+    eth_idx = dff[dff['value_weth'] > 0].index
+    dff.at[eth_idx, 'value_wbtc'] = dff['value_usd_2dec'] / 2 / wbtc_rate
```
petrovska-petro, line 33: @jorijnsmit I am unsure of this step, perhaps i do not understand properly

but the row is for victim **0xB1C120957a5b5C45A15fd6e5E17f5A2B70bF49d0**, i think in this way will end up with a `value_wbtc` of **2.16677859094074**, while i think it should be **4.333557181881488**

since we are converting 26.55771576 weth to wbtc, the operation should look something like:

```
wbtc_rate=55923.6062958284
```

```
1. 242348.1457 (value_usd_2dec) - 119007.4910677624(2.12803678 * 55923.6062958284) = 123340.6546322376 (weth part)

2. 123340.6546322376/55923.6062958284=2.2055204018814885 (converting the remaining of usdc value belonging to weth using the rate of wbtc)

3. 2.2055204018814885  + 2.12803678 = 4.333557181881488 (this i think should  be its value_wbtc)
```

let me know if i am mixing stuff

Hal69k: on Dec 1st ETH was worth between $4400 and $4600 which is about 2.2 BTC

jorijnsmit: good catch @petrovska-petro, i am overwriting the other half of the wbtc value in L33; thus ignoring half of the user's value.

will fix this by *not* dividing by 2 (essentially L33 becoming `value_wbtc = value_usd_2dec / wbtc_rate`).

Hal69k (approved):
the amount sent to 0xB1C120957a5b5C45A15fd6e5E17f5A2B70bF49d0  now looks correct

petrovska-petro (approved):
- Getting same csv output locally ✅ 

- Sorted issue with 0xB1C120957a5b5C45A15fd6e5E17f5A2B70bF49d0 ✅

Issue #78, opened 2022-02-01
Grant contract access to Treasury multisigs on ibBTC


sajanrajdev:
## Overview

Grant contract access to the Treasury Ops and Treasury Vault multisigs on all ibBTC peaks to enable minting. This needs to be done from the Peak owner multi (dualGovernance).

## Technical Details

Call:
`peak.approveContractAccess(multi_address)`

From: dualGovernance: 0xCF7346A5E41b0821b80D5B3fdc385EEB6Dc59F44

## Urgency and reason

Allow these multisig to perform operations with ibBTC directly without the need to delegate these steps to other multisigs. Mid priority,

gosuto-inzasheru:
nonce 25: https://etherscan.io/tx/0xea96d23f87051163e102bbd85f7e09fd69437c3262ecd99fc9f997b5b66dc6bb
nonce 26: https://etherscan.io/tx/0xe6832e0ca172bf747c31753f5dc91382db42d1c1d582994ab0a7b2ae78870aca

PR #79, opened 2022-02-01
feat: approve contract access on ibbtc peaks

sajanrajdev:
This PR closes issue #78 

Grants contract access to the treasury_ops multi on both ibBTC peaks. The peaks are owned by the DefiDollar-Badger dual multi, so the txs are to be loaded there.

**NOTE:**The atomic tx had to be split as there was an `ExecutionFailure` when running the two together - couldn't figure out the reason. Txs go through when loaded independently. 

Run the following to approve on the Badger peak;
```
brownie run scripts/issue/78/grant_access_ibbtc_peak.py
```

And with the following to approve on the byvWBTC peak;
```
brownie run scripts/issue/78/grant_access_ibbtc_peak.py main byvwbtc
```

### DIFF
--- a/scripts/issue/78/grant_access_ibbtc_peak.py
+++ b/scripts/issue/78/grant_access_ibbtc_peak.py
@@ -4,16 +4,17 @@ from brownie import interface
 
 TREASURY_OPS = registry.eth.badger_wallets.treasury_ops_multisig
 
-def main():
+def main(peak="badger"):
     safe = GreatApeSafe(registry.ibbtc.dfdBadgerShared)
     safe.init_badger()
 
-    badger_peak = interface.IPeak(registry.ibbtc.badgerPeak, owner=safe.account)
-    byvWbtc_peak = interface.IPeak(registry.ibbtc.byvWbtcPeak, owner=safe.account)
-
-    badger_peak.approveContractAccess(TREASURY_OPS)
-    assert badger_peak.approved(TREASURY_OPS)
-    byvWbtc_peak.approveContractAccess(TREASURY_OPS)
-    assert byvWbtc_peak.approved(TREASURY_OPS)
+    if peak == "badger":
+        badger_peak = interface.IPeak(registry.ibbtc.badgerPeak, owner=safe.account)
+        badger_peak.approveContractAccess(TREASURY_OPS)
+        assert badger_peak.approved(TREASURY_OPS)
+    else:
+        byvWbtc_peak = interface.IPeak(registry.ibbtc.byvWbtcPeak, owner=safe.account)
+        byvWbtc_peak.approveContractAccess(TREASURY_OPS)
+        assert byvWbtc_peak.approved(TREASURY_OPS)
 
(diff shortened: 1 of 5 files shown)


petrovska-petro:
> **Note:** @petrovska-petro @jorijnsmit, for some reason I get an `ExecutionFailure` when running the script with both peaks together in an atomic tx. Running them independently doesn't produce this error and I can't figure why this is. Would you mind running this to see if you also get this error?

Getting also locally `ExecutionFailure()`, looking at the contract action cannot make sense on why atomicly will fail

ref on method: https://etherscan.io/address/0xa9499b50c4a74000687435a895b55974f079ecaa#code#F5#L26

probably separating them will be a quick workaround

sajanrajdev:
> > **Note:** @petrovska-petro @jorijnsmit, for some reason I get an `ExecutionFailure` when running the script with both peaks together in an atomic tx. Running them independently doesn't produce this error and I can't figure why this is. Would you mind running this to see if you also get this error?
> 
> Getting also locally `ExecutionFailure()`, looking at the contract action cannot make sense on why atomicly will fail
> 
> ref on method: https://etherscan.io/address/0xa9499b50c4a74000687435a895b55974f079ecaa#code#F5#L26
> 
> probably separating them will be a quick workaround

I agree, will go ahead and do that

sajanrajdev:
@petrovska-petro, changes made.

**NOTE:**The atomic tx had to be split as there was an `ExecutionFailure` when running the two together - couldn't figure out the reason. Txs go through when loaded independently. 

Run the following to approve on the Badger peak;
```
brownie run scripts/issue/78/grant_access_ibbtc_peak.py
```

And with the following to approve on the byvWBTC peak;
```
brownie run scripts/issue/78/grant_access_ibbtc_peak.py main byvwbtc
```

scripts/issue/78/grant_access_ibbtc_peak.py
```diff
@@ -4,16 +4,17 @@
 
 TREASURY_OPS = registry.eth.badger_wallets.treasury_ops_multisig
 
-def main():
+def main(peak="badger"):
     safe = GreatApeSafe(registry.ibbtc.dfdBadgerShared)
     safe.init_badger()
```
petrovska-petro, line 9: minor comment, in this case it is not necessary, just slows down the script

jorijnsmit (approved):
does what it says on the box

cant figure out why it reverts on combining both into one tx either.

scripts/issue/78/grant_access_ibbtc_peak.py
sajanrajdev: Ah, I had added it because I thought for a second I could use the whitelisting functions from it. Turns out I couldn't and forgot to remove it. Thanks for the heads up!

sajanrajdev:
Removed unnecessary badger class initialization.

jorijnsmit (approved):
redundant line removed

petrovska-petro (approved):

Issue #33, opened 2022-01-25
swap EPS buffer for BNB 


petrovska-petro:
For prev distribution we extracted a small percentage from the EPS distribution with the purpose of topping up signers wallets so we can keep it up with the monthly distribution

- buffer of 2117 EPS will result in ~0.94 BNB at current rates 

- should be sufficient gas to keep distributions for at least 26 tx of 400 recipients. ref tx cost: https://bscscan.com/tx/0x7fc87c323b0dd135f827d85cbcbdac590cd3f7d353ae8ac566a366ce7b72795b

petrovska-petro:
```
{
│   'ethereum_client': <gnosis.eth.ethereum_client.EthereumClient object at 0x7faad33daa30>,
│   'safe_address': '0x329543f0F4BB134A3f7a826DC32532398B38a3fA',
│   'to': '0x40A2aCCbd92BCA938b02010E17A5b8929b49130D',
│   'value': 0,
│   'data': HexBytes('0x8d80ff0a000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000001f200a7f552078dcc247c2684336020c03648500c6d9f00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b300000000000000000000000010ed43c718714eb63d5aa57b78b54704e256024e000000000000000000000000000000000000000000000072c6f3dbf65aae66420010ed43c718714eb63d5aa57b78b54704e256024e0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000010418cbafe5000000000000000000000000000000000000000000000072c6f3dbf65aae66420000000000000000000000000000000000000000000000000de3f4d24248a1fc00000000000000000000000000000000000000000000000000000000000000a0000000000000000000000000329543f0f4bb134a3f7a826dc32532398b38a3fa0000000000000000000000000000000000000000000000000000000061fdf8330000000000000000000000000000000000000000000000000000000000000002000000000000000000000000a7f552078dcc247c2684336020c03648500c6d9f000000000000000000000000bb4cdb9cbd36b01bd1cbaebf2de08d9173bc095c0000000000000000000000000000'),
│   'operation': 1,
│   'safe_tx_gas': 0,
│   'base_gas': 0,
│   'gas_price': 0,
│   'gas_token': '0x0000000000000000000000000000000000000000',
[...]

PR #80, opened 2022-02-02
Add pancakeswap class and script to swap eps -> bnb

petrovska-petro:
Tackles #33 

Command:

```
brownie run --network bsc-main-fork scripts/issue/33/swap_eps_for_bnb
```

**Note:** I am getting timeout while testing locally, so i am unable of getting till the end rn. Would appreciate if @jorijnsmit @sajanrajdev can run it and let me know if you have same issue

Err:

```
ReadTimeout: HTTPConnectionPool(host='127.0.0.1', port=8545): Read timed out. (read timeout=10)
```

### DIFF
--- a/great_ape_safe/ape_api/pancakeswap_v2.py
+++ b/great_ape_safe/ape_api/pancakeswap_v2.py
@@ -1,0 +1,15 @@
+from brownie import web3
+from great_ape_safe.ape_api.uni_v2 import UniV2
+from helpers.addresses import registry
+
+
+class Pancakeswap(UniV2):
+    def __init__(self, safe):
+        self.safe = safe
+
+        self.router = safe.contract(registry.bsc.pancakeswap.router_v2)
+        self.factory = safe.contract(registry.bsc.pancakeswap.factory_v2)
+
+        self.max_slippage = 0.02
+        self.max_weth_unwrap = 0.01
+        self.deadline = 60 * 60 * 12
--- a/great_ape_safe/ape_api/pancakeswap_v2.py
+++ b/great_ape_safe/ape_api/pancakeswap_v2.py
@@ -1,9 +1,8 @@
-from brownie import web3
 from great_ape_safe.ape_api.uni_v2 import UniV2
 from helpers.addresses import registry
 
 
-class Pancakeswap(UniV2):
+class PancakeswapV2(UniV2):
     def __init__(self, safe):
         self.safe = safe
 
--- a/helpers/addresses.py
+++ b/helpers/addresses.py
@@ -466,6 +466,7 @@ ADDRESSES_BSC = {
     "pancakeswap": {
         "router_v1": "0x05fF2B0DB69458A0750badebc4f9e13aDd608C7F",
         "router_v2": "0x10ED43C718714eb63d5aA57B78B54704E256024E",
+        "factory_v2": "0xcA143Ce32Fe78f1f7019d7d551a6402fC5350c73"
     },
 }
 
--- a/great_ape_safe/great_ape_safe.py
+++ b/great_ape_safe/great_ape_safe.py
@@ -19,6 +19,7 @@ from great_ape_safe.ape_api.convex import Convex
 from great_ape_safe.ape_api.curve import Curve
 from great_ape_safe.ape_api.curve_v2 import CurveV2
 from great_ape_safe.ape_api.opolis import Opolis
+from great_ape_safe.ape_api.pancakeswap_v2 import Pancakeswap
 from great_ape_safe.ape_api.sushi import Sushi
 from great_ape_safe.ape_api.uni_v2 import UniV2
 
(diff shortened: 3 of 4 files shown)


great_ape_safe/ape_api/pancakeswap_v2.py
```diff
@@ -0,0 +1,15 @@
+from brownie import web3
```
jorijnsmit, line 1: redundant..?

great_ape_safe/ape_api/pancakeswap_v2.py
```diff
@@ -0,0 +1,15 @@
...
+from great_ape_safe.ape_api.uni_v2 import UniV2
+from helpers.addresses import registry
+
+
+class Pancakeswap(UniV2):
```
jorijnsmit, line 6: would stick to convention of `PancakeswapV2` or `PancakeV2`.

i say this also because i know we will need to add v1 at one point, since some of our lps on bsc will need v1 and some will need v2 router to withdraw from: https://bscscan.com/tokenholdings?a=0x6DA4c138Dd178F6179091C260de643529A2dAcfe

jorijnsmit: really smooth use of inheritance here btw 🤩

great_ape_safe/ape_api/pancakeswap_v2.py
```diff
@@ -0,0 +1,15 @@
...
+    def __init__(self, safe):
+        self.safe = safe
+
+        self.router = safe.contract(registry.bsc.pancakeswap.router_v2)
+        self.factory = safe.contract(registry.bsc.pancakeswap.factory_v2)
+
+        self.max_slippage = 0.02
+        self.max_weth_unwrap = 0.01
+        self.deadline = 60 * 60 * 12
```
jorijnsmit, line 15: these will get inherited from UniV2 if left undeclared. so if using same values as univ2 these could be left out.

helpers/addresses.py
```diff
@@ -466,6 +466,7 @@
     "pancakeswap": {
         "router_v1": "0x05fF2B0DB69458A0750badebc4f9e13aDd608C7F",
         "router_v2": "0x10ED43C718714eb63d5aA57B78B54704E256024E",
+        "factory_v2": "0xcA143Ce32Fe78f1f7019d7d551a6402fC5350c73"
```
jorijnsmit, line 469: confirmed via bscscan.

jorijnsmit (commented):
left minor comments

great_ape_safe/ape_api/pancakeswap_v2.py
petrovska-petro: i tried removing them, but while running got errors

```
  File "./great_ape_safe/ape_api/uni_v2.py", line 153, in swap_exact_tokens_for_eth
    amountOut * (1 - self.max_slippage),
AttributeError: 'PancakeswapV2' object has no attribute 'max_slippage'
```

great_ape_safe/ape_api/pancakeswap_v2.py
petrovska-petro: last commit adapts better naming for the versions all around

great_ape_safe/ape_api/pancakeswap_v2.py
jorijnsmit: ah yea sorry my bad, that is because they are not part of the global class like the defined funcs, but only initialised through __init__. guess we could move them outside the __init__, but some other time :)

jorijnsmit (approved):
all good here
```
snapshot result for 0x329543f0F4BB134A3f7a826DC32532398B38a3fA:
                  balance_before         balance_after              balance_delta
symbol                                                                           
ETH         0.000000000000000000  0.924017489420419844       0.924017489420419844
EPS     2,117.264868292869121602  0.000000000000000000  -2,117.264868292869121602 
```

sajanrajdev (approved):
Script to swap all the existing EPS on the Multisig for BNB. Snapshot results:
```
snapshot result for 0x329543f0F4BB134A3f7a826DC32532398B38a3fA:
                  balance_before         balance_after              balance_delta
symbol
ETH         0.000000000000000000  0.917225199039372808       0.917225199039372808
EPS     2,117.264868292869121602  0.000000000000000000  -2,117.264868292869121602
```

LGTM

NOTE: We could look into refactoring the snapshot function later on so that it displays the correct symbol for each Chain's native token. It is hardcoded to ETH which is not bad but could be misleading.

petrovska-petro:
> Script to swap all the existing EPS on the Multisig for BNB. Snapshot results:
> 
> ```
> snapshot result for 0x329543f0F4BB134A3f7a826DC32532398B38a3fA:
>                   balance_before         balance_after              balance_delta
> symbol
> ETH         0.000000000000000000  0.917225199039372808       0.917225199039372808
> EPS     2,117.264868292869121602  0.000000000000000000  -2,117.264868292869121602
> ```
> 
> LGTM
> 
> NOTE: We could look into refactoring the snapshot function later on so that it displays the correct symbol for each Chain's native token. It is hardcoded to ETH which is not bad but could be misleading.

Good idea, we can leverage there the `network.chain.id` to decide which native token to show up

Issue #45, opened 2022-01-26
Process bDIGG to DIGG to cover future emissions


petrovska-petro:
Withdraw from the bDIGG sett held in the treasury ops(0x042B32Ac6b453485e357938bdC38e0340d4b9276) to cover future emissions demand of DIGG

Needs to cover till Q1 ends:

```
0.01544 * 12 * 7 * 9 = 11.67264 digg to cover, including this week
```

jorijnsmit:
might be handy to also withdraw all from the bdigg/eth position there for even more (b)digg.

jorijnsmit:
```
Events In This Transaction
--------------------------
├── SushiSwap LP Token (0xF9440fEDC72A0B8030861DcDac39A75b544E7A3c)
│   ├── Approval
│   │   ├── owner: 0x042B32Ac6b453485e357938bdC38e0340d4b9276
│   │   ├── spender: 0xd9e1cE17f2641f24aE83637ab66a2cca9C378B9F
│   │   └── value: 26688069502695397918
│   ├── Transfer
│   │   ├── from: 0x042B32Ac6b453485e357938bdC38e0340d4b9276
│   │   ├── to: 0xF9440fEDC72A0B8030861DcDac39A75b544E7A3c
│   │   └── value: 26688069502695397918
│   ├── Transfer
│   │   ├── from: 0x0000000000000000000000000000000000000000
│   │   ├── to: 0x5ad6211CD3fdE39A9cECB5df6f380b8263d1e277
│   │   └── value: 7871248168945486
│   └── Transfer
│       ├── from: 0xF9440fEDC72A0B8030861DcDac39A75b544E7A3c
│       ├── to: 0x0000000000000000000000000000000000000000
│       └── value: 26688069502695397918
│   
├── Badger Sett Digg (0x7e7E112A68d8D2E221E11047a72fFC1065c38e1a)
│   └── Transfer
│       ├── from: 0xF9440fEDC72A0B8030861DcDac39A75b544E7A3c
│       ├── to: 0x042B32Ac6b453485e357938bdC38e0340d4b9276
│       └── value: 17924943022053909055
│   
├── Wrapped Ether (0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2)
│   └── (unknown)
│       ├── topic1: 0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef
[...]

Issue #18, opened 2022-01-19
refactor all IERC20* interface declarations to interface.ERC20


jorijnsmit:
see https://github.com/Badger-Finance/badger-utils/issues/16

(i just found out about this too)

PR #91, opened 2022-02-02
process bdigg to digg from sett and slp position for emissions

jalbrekt85:
**#18** **#45**

- script to withdraw bdigg from sett for digg
- script to withdraw from eth/bdigg sushi position for even more digg
- could be improved by calc withdraw exact amount of digg needed before withdrawing from bdigg positions
- IERC20s are now use brownie's ERC20


scripts/issue/45/process_bdigg_sushi.py
```diff
@@ -0,0 +1,31 @@
...
+    weth = interface.ERC20(registry.eth.treasury_tokens.WETH, owner=safe.account)
+    wbtc = registry.eth.treasury_tokens.WBTC
+
+    safe.take_snapshot(tokens=[slp, bdigg, digg])
+    
+    bdigg_before = bdigg.balanceOf(safe)
+    weth_before = weth.balanceOf(safe)
+    safe.sushi.remove_liquidity(False, slp, SLP_TO_WITHDRAW * 10** slp.decimals())
+    bdigg.withdraw(bdigg.balanceOf(safe) - bdigg_before)
+    safe.sushi.swap_tokens_for_tokens(weth, weth.balanceOf(safe) - weth_before, [weth, wbtc, digg], safe.account)
```
jorijnsmit, line 28: we are not looking to buy digg. (also, in this case, i think we couldve gone for withdraw_to_one_coin no?). anyway just leave the weth as weth, thats fine.

and just to note: careful here with 'stringing' balanceOf funcs. the weth we get onfork from the remove_liquidity might be different from the amount of weth we will get onchain (due to pool composition changing in between). the script however will then still call the swap with the weth amount hardcoded by the simulation onfork---which could result in a balance too low.

the only time you can call balanceOf correctly in these scripts is in the first action, after that it will not be precise...

jorijnsmit (changes requested):
good stuff!!

can you consolidate them into one script/func/tx that withdraws *all*?

order is important here; so first sushi then sett.

also you can then use the handy `ISettV4h.withdrawAll`, which calls a balanceOf onchain and solves our problem of not being able to call balanceOf later into the multicall ;)

scripts/issue/45/process_bdigg_sushi.py
jalbrekt85: good to know about how `balanceOf` behaves in that case. Gathering all the bdigg and using `withdrawAll` at the end definitely makes more sense. For `withdraw_to_one_coin`, I believe that is for curve liquidity unless I am missing something

jorijnsmit: sorry, you are right, sushi does not support one sided withdrawal.

scripts/issue/45/process_bdigg.py
```diff
@@ -0,0 +1,26 @@
...
+    safe = GreatApeSafe(registry.eth.badger_wallets.treasury_multisig)
+    safe.init_sushi()
+    
+    slp = interface.IUniswapV2Pair(registry.eth.treasury_tokens.slpEthBDigg, owner=safe.account)
+    bdigg = interface.ISettV4h(registry.eth.treasury_tokens.bDIGG, owner=safe.account)
+    digg = interface.ERC20(registry.eth.treasury_tokens.DIGG)
+
+    safe.take_snapshot(tokens=[slp, bdigg, digg])
+    
+    safe.sushi.remove_liquidity(False, slp, SLP_TO_WITHDRAW * 10** slp.decimals())
```
jorijnsmit, line 22: perfect, except for here we want to withdraw slp.balanceOf(safe). calling balanceOf here is fine, since it is the first action, and we know that there is no way the balance will change between running simulation onfork (and thus getting hardcoded amount), and executing onchain.

jorijnsmit (commented):
if you can alter L22 we can merge and run it!

jorijnsmit (approved):
smooth as butter

sajanrajdev (approved):
Removes liquidity from Sushi LP into DIGG and withdraws all from treasury ops bDIGG position. Snapshot results:
```
snapshot result for 0x042B32Ac6b453485e357938bdC38e0340d4b9276:
               balance_before         balance_after           balance_delta
symbol
SLP     26.688069502695397918  0.000000000000000000  -26.688069502695397918
bDIGG   20.886782838869027490  0.000000000000000000  -20.886782838869027490
DIGG     0.335341938000000000  7.693524775000000000    7.358182837000000000
```

jorijnsmit:
@jalbrekt85 merging this for you right away so i can try and post it.

Issue #67, opened 2022-01-31
Set perf fee gov 2000 & perf fee strat 0 in native.sushiDiggWbtc


petrovska-petro:
- Currently, it is set to 1000 perf fee strat also and some % of fees are heading to **0xda25ee226e534d868f0dd8a459536b03fee9079b** to be consistent with rest of set up

- Process all fees thru dev_msig, setting up perf fee strat to 0 and perf fee gov to 2000

petrovska-petro:
note from meeting: 

```
performanceFeeStrategist not 0 for native.sushiWbtcEth
performanceFeeStrategist not 0 for native.sushiBadgerWbtc
performanceFeeStrategist not 0 for native.sushiDiggWbtc
performanceFeeStrategist not 0 for experimental.sushiIBbtcWbtc
```

petrovska-petro:
Output while posting, note it does not emit events while updates the fee values:

```
Events In This Transaction
--------------------------
└── Gnosis Safe (0xB65cef03b9B89f99517643226d76e286ee999e77)
    └── ExecutionSuccess
        ├── txHash: 0x046e40f7d2a063fa5381d0e5e865336f9604b619379c9e58e2ae1a819d1355b4
        └── payment: 0

{
│   'ethereum_client': <gnosis.eth.ethereum_client.EthereumClient object at 0x7fac7d217a90>,
│   'safe_address': '0xB65cef03b9B89f99517643226d76e286ee999e77',
│   'to': '0x40A2aCCbd92BCA938b02010E17A5b8929b49130D',
│   'value': 0,
[...]

PR #93, opened 2022-02-02
Script to modify perf fee strat & gov in few strategiest where is sending to deployer wallet

petrovska-petro:
Tackles #67 

Command:

```
brownie run issue/67/set_fees_strat_and_gov
```

### DIFF
--- a/interfaces/badger/IStrategy.sol
+++ b/interfaces/badger/IStrategy.sol
@@ -41,4 +41,6 @@ interface IStrategy {
     function tend() external;
 
     function harvest() external;
+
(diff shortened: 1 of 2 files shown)


jorijnsmit (approved):
looks good to me, will make all perf fees on these four strats go 100% to gov instead of 50% strat 50% gov

sajanrajdev (approved):
Runs like charm: Sets perf fee strategy to 0 and performance fee gov to 2000 for all those strats that have any other configuration. In this case it does this for all the sushi strats.

LGTM

Issue #94, opened 2022-02-03
week 5: run gas refiller


jorijnsmit:

jorijnsmit:
posted nonce 608: https://gnosis-safe.io/app/eth:0xD4868d98849a58F743787c77738D808376210292/transactions/0x0b80b00517c5e7c8148fe66b3851b288b8532a76ad732677925fd5f82ad777ab
```
ops_deployer3
snapshot result for 0x283C857BA940A61828d9F4c09e3fceE2e7aEF3f7:
              balance_before         balance_after         balance_delta
symbol                                                                  
ETH     1.231121475773451807  2.000000000000000000  0.768878524226548193 

ops_deployer7
snapshot result for 0x7140B5254d80154f9Fc5F86054efB210f3a1e2c6:
              balance_before         balance_after         balance_delta
symbol                                                                  
ETH     1.847184429400000000  2.047184429400000000  0.200000000000000000 

ops_executor3
snapshot result for 0xC69Fb085481bC8C4bfF99B924076656305D9a25D:
              balance_before         balance_after         balance_delta
symbol                                                                  
ETH     0.989663564668420088  1.189663564668420088  0.200000000000000000 

ops_executor6
snapshot result for 0x66496eBB9d848C6A8F19612a6Dd10E09954532D0:
              balance_before         balance_after         balance_delta
symbol                                                                  
ETH     0.233339737888087393  1.000000000000000000  0.766660262111912607 

ops_root-validator-new
snapshot result for 0x1318d5c0C24830D86Cc27Db13Ced0CED31412438:
              balance_before         balance_after         balance_delta
symbol                                                                  
ETH     1.389475612343025426  2.000000000000000000  0.610524387656974574 

ops_botsquad
snapshot result for 0xF8dbb94608E72A3C4cEeAB4ad495ac51210a341e:
               balance_before          balance_after         balance_delta
symbol                                                                    
ETH     12.934791141354470078  15.000000000000000000  2.065208858645529922 

ops_botsquad_cycle0
snapshot result for 0x1a6D6D120a7e3F71B084b4023a518c72F1a93EE9:
              balance_before         balance_after         balance_delta
symbol                                                                  
ETH     1.531175076848271948  2.000000000000000000  0.468824923151728052 

ops_multisig
snapshot result for 0xD4868d98849a58F743787c77738D808376210292:
                balance_before           balance_after          balance_delta
symbol                                                                       
[...]

PR #95, opened 2022-02-03
add refill script, clean up addresses, move unlocker

jorijnsmit:
- solve #94. since we will run it weekly i placed this in scripts/badger
- same for cvx unlocker, weekly script so moved to scripts/badger
- cleaned up addresses.py a bit


jorijnsmit:
@Tritium-VLK still not sure about all these comments in addresses.py, if we want to share this publicly as our contract list it might be better to remove all that and have it be a clean dict.

petrovska-petro (dismissed):
Like the idea of introducing the scripts into badger for usual runs

Lots of juicy snaps added in the refill_gas script, useful! 🔥

petrovska-petro (approved):

hmmmason (approved):

PR #102, opened 2022-02-04
arb script for bvecvxcvx factory pool

jorijnsmit:
- minor improvements in curve class
- introduces new dependency

arb script is reusable and figures out exact amount to arb automatically.

if treasury ops' balance is too low it will arb with its full balance.

in case the arb amount is under the threshold (in this case 5k) it will stop.

### DIFF
--- a/scripts/badger/arb_bvecvxcvx_pool.py
+++ b/scripts/badger/arb_bvecvxcvx_pool.py
@@ -44,9 +44,9 @@ def main():
         mantissa_to_add = solve(
             (int(bal_bvecvx) + x) / (int(bal_total) + x) - TARGET, x
         )[0]
-        assert mantissa_to_add > THRESHOLD * 1e18
         if mantissa_to_add >= bvecvx.balanceOf(safe):
             mantissa_to_add = bvecvx.balanceOf(safe)
+        assert mantissa_to_add > THRESHOLD * 1e18
         safe.curve.deposit(bvecvxcvx, mantissa_to_add, bvecvx)
 
     bal_cvx = cvx.balanceOf(bvecvxcvx)
--- a/scripts/badger/arb_bvecvxcvx_pool.py
+++ b/scripts/badger/arb_bvecvxcvx_pool.py
@@ -1,0 +1,63 @@
+from brownie import interface
+from sympy import Symbol
+from sympy.solvers import solve
+
+from great_ape_safe import GreatApeSafe
+from helpers.addresses import registry
+
+
+TARGET = .55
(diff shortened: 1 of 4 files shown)


petrovska-petro (approved):
Works beautifully, smart usage of `sympy` was not familiar with that library

@sajanrajdev just curious on this script, are you getting timeout as well by any chance?

sajanrajdev (dismissed):
Script LGTM - Arbs the bveCVX/CVX Curve Pool as needed according to certain thresholds. Running it at this moment it would execute the arb operation with the following result:
```
cvx              61727.23952792606 0.5460027034014117
bvecvx           51325.75296347792 0.4539972965985884
cvx+bvecvx       113052.99249140397
Transaction sent: 0x3044c0e30c91c09fbb57a92e6d8a247d1256e5f69b0348c8afcef3d5c16b5500
  Gas price: 0.0 gwei   Gas limit: 12000000   Nonce: 1
  ERC20.approve confirmed   Block: 14140147   Gas used: 46898 (0.39%)

Transaction sent: 0x6a04aeda271101ef7a40947bc44eafaad3af9da805bbd1c0681a24fa2bff41d7
  Gas price: 0.0 gwei   Gas limit: 12000000   Nonce: 2
  IStableSwap2Pool.add_liquidity confirmed   Block: 14140148   Gas used: 126220 (1.05%)

cvx              61727.23952792606 0.45
bvecvx           75444.40386746518 0.55
cvx+bvecvx       137171.64339539123
```

@petrovska-petro, I also go the timeout error...

scripts/badger/arb_bvecvxcvx_pool.py
```diff
@@ -0,0 +1,63 @@
...
+    )
+    bvecvxcvx = interface.IStableSwap2Pool(
+        registry.eth.treasury_tokens['bveCVX-CVX-f'], owner=safe.account
+    )
+
+    safe.take_snapshot(tokens=[bvecvx.address, bvecvxcvx.address])
+
+    bal_cvx = cvx.balanceOf(bvecvxcvx)
+    bal_bvecvx = bvecvx.balanceOf(bvecvxcvx)
+    bal_total = bal_cvx + bal_bvecvx
```
mitche50, line 35: Here we're assuming that bveCVX will never increase in PPFS.  We can note this in a comment as it likely will not during the expected use of this script, or we can take into account PPFS, but should be apparent if used for reference down the line.

sajanrajdev (approved):
Added note about potential ppfs change.

mitche50 (approved):

Issue #104, opened 2022-02-04
 leverage `network.chain.id` to decide which native token to show up in snapshot


jorijnsmit:
see https://github.com/Badger-Finance/badger-multisig/pull/80#issuecomment-1029079241

petrovska-petro:
Closing after merging #105

Issue #96, opened 2022-02-03
port curve swap function


jorijnsmit:
old source code:
```
    def _swap(self, registry, pool_address, asset_in, asset_out, mantissa):
        # helper for common functionalities despite of the registry/pool route
        initial_asset_out_balance = asset_out.balanceOf(self.safe)
        asset_in.approve(pool_address, mantissa)
        i, j, underlying_coin_boolean = registry.get_coin_indices(
            pool_address, asset_in, asset_out
        )
        pool = self.safe.contract(pool_address)
        expected = pool.get_dy(i, j, mantissa) * (1 - self.max_slippage_and_fees)
        # L139 docs ref
        pool.exchange(i, j, mantissa, expected)
        assert asset_out.balanceOf(self.safe) >= initial_asset_out_balance + expected


    def swap(self, asset_in, asset_out, mantissa):
        # swap `asset_in` (amount: `mantissa`) for `asset_out`
        # https://curve.readthedocs.io/factory-pools.html?highlight=exchange#StableSwap.exchange
        # https://curve.readthedocs.io/registry-exchanges.html?highlight=get_best_rate#finding-pools-and-swap-rates
        pool = self.generalised_rates_info.get_best_rate(asset_in, asset_out, mantissa)[0]
        registry = self._get_registry(self.safe.contract(pool))
        self._swap(registry, pool, asset_in, asset_out, mantissa)
```

jorijnsmit:
@jalbrekt85 you mind seeing how easy it is to pr this into the current class? will have to happen after merge of #89 .

also i think changes to curve class in that pr will allow you to clean up calls such as `pool.add_liquidity[f'uint[{len(mantissas)}],uint']` since the pool interface will be correctly set automatically.

have a look, maybe you can help review #89 as well.

jalbrekt85:
> @jalbrekt85 you mind seeing how easy it is to pr this into the current class? will have to happen after merge of #89 .
> 
> also i think the pending changes to the curve class in that pr will allow you to clean up calls such as `pool.add_liquidity[f'uint[{len(mantissas)}],uint']` since the pool interface will be correctly set automatically.
> 
> have a look, maybe you can help review #89 as well.

 👍

Issue #87, opened 2022-02-02
dogfood 3pool into 3eur, cleanup from last withdrawal


jorijnsmit:
- convert last bit of mim3pool to 3pool or send to tr.ops
- dogfood 3pool into 3eur instead of mim
- send crv and cvx to tr.ops

jorijnsmit:
clean was done already in #88 

only thing left is to get whole 3pool balance into staked cvx3eur

jorijnsmit:
https://etherscan.io/tx/0xf1e631aa4b8c48c4f3390f21fc3d9bb180a5d26160dc47b1314a013a2ad54726

PR #105, opened 2022-02-07
curve swap, snapshot native currency

jalbrekt85:
**for #96 & #104**

- swap functionality added to curve class. got very inconsistent results using the [registry exchange](https://etherscan.io/address/0xD1602F68CC7C4c7B59D686243EA35a9C73B0c6a2#readContract), so resorted to providing `lp_token`/`pool` and just calling `exchange`. would be nice if the swap could be done without having to provide pool. ran into a similiar shortcoming when calling `get_coins` from factory contract.
- `tests/curve/test_curve_swap.py` to play around with
- native currency symbol now prints in snapshot, chain id look up from data from `helpers/chaindata.json` loaded in `addresses`


tests/curve/test_curve_swap.py
```diff
@@ -0,0 +1,14 @@
+from brownie import interface
+from helpers.addresses import registry as registry_addr
+
+def test_deposit_given_amounts(safe, curve, USDC):
+    TOKEN = interface.ICurveLP(registry_addr.eth.treasury_tokens.crv3pool, owner=safe.account)
+    DAI = interface.ERC20(registry_addr.eth.treasury_tokens.DAI)
+
```
jorijnsmit, line 7: make sure to mint the token using `MintableForkToken` so that the test doesnt rely on actual onchain balances (and thus cannot fail for that reason).

helpers/addresses.py
```diff
@@ -659,6 +659,10 @@ def checksum_address_dict(addresses):
     return checksummed
 
 
+with open('helpers/chaindata.json') as chaindata:
+    chain_ids = json.load(chaindata)
+
+
```
jorijnsmit, line 665: really clean solution, i like it a lot!

the only thing im wondering here is why put this in addresses.py? id suggest to move this to chaindata.py, and then `from helpers.chaindata import labels` or something.

also maybe instead of str: str set the json up as int: str?

jorijnsmit (commented):
good thing to pick up the testing here. i think curve swapping is the toughest nut to crack so far, with its different protocol versions, token interfaces and registries.

i was wrestling with this on friday as well for #87. i dont think we are there yet, but what would really help is a couple more test methods here that swap the different varieties of tokens/pools. once i have a better view of this i will spec a ticket if needed.

- i will pick up the curve swap func and curve interfaces from here, so i suggest to merge what you have here so i can continue with it.
- please have a look at some of the my other review comments and adjust where needed.

jorijnsmit (dismissed):
as discussed, will tinker a bit more with curve interfaces and registries.

other additions look good; test functions are good to have and chain label json is neat solution.

petrovska-petro (approved):
- Great addition for each native token for each chain, cross-checked most relevant chains for us: Arb, Mainnet and BSC and worked as expected ✅ 

- Curve stuff will be handle in another PR

jorijnsmit (approved):

petrovska-petro (approved):

Issue #71, opened 2022-01-31
consolidate all $badger into treasury vault


jorijnsmit:
see https://etherscan.io/token/0x3472A5A71965499acd81997a54BBA8D852C6E53d#balances

badger.rewards ("geyser", `0xBD9c69654B8F3E5978DFd138B00cB0Be29F28cCf`):
- turn off bbadger strat
- move all $badger in bbadger strat to bbadger vault
- this prevents withdrawals from calling badger.rewards, ie: https://etherscan.io/tx/0x99aa4a6593444eb38cdd4c452f9e01f4006b3ec2e3d6babaee5a2e51e95fc3eb
- calc total $badger that should be in vault for users to withdraw
- move surplus to treasury vault

badger.treasury (`0x4441776e6A5D61fA024A5117bfc26b953Ad1f425`):
- holds 35% of total supply; move all to treasury vault

after that all uncirculating supply will be consolidated into treasury vault.

sajanrajdev:
After further investigation, the following was found:
- The StakingRewards contract (aka Badger Geyser) is on a BADGER deficit. The balance of the pool for bBADGER is the following:
![image](https://user-images.githubusercontent.com/25463788/152573838-6d1df5bf-1bb2-4474-8b96-80603daed2cf.png)
And the geyser contains the following:
![image](https://user-images.githubusercontent.com/25463788/152573921-01bbcff4-9d11-443a-bd37-d23d43b897b8.png)

The remaining must be injected to the Geyser before we can call `controller.withdrawAll(badger_token)` to bring all BADGER from the strategy to the vault.

- In order to move the 35% of the BADGER supply sitting on the DAO_terasury (0x4441776e6A5D61fA024A5117bfc26b953Ad1f425) this contract needs to be upgraded to point the beneficiary address to the devMultisig. After this, calling `DAO_treasury.release()` will transfer the balance to the multisig.

sajanrajdev:
New logic contract deployed: https://etherscan.io/address/0xb7acd34643181c879437c2967538d5c0ea42b5d9#code

jorijnsmit:
posted nonce 1006: https://gnosis-safe.io/app/eth:0xB65cef03b9B89f99517643226d76e286ee999e77/transactions/0xa44dc361b37b36bfec093331393532398c10d49d332d76f6178645ecbdc858bf
```
[...]

PR #110, opened 2022-02-07
feat: script to consolidate badger

sajanrajdev:
This PR tackles issue #71 

The script is split into two parts:

### PART 1: Upgrade DAO_treasury contract
- Upgrades the logic of the DAO_treasury contract to the newly deployed logic which changes the beneficiary of the release() function.
- Upgrade performed on the old_proxyAdmin directly from the the [Ops_Multisig_old](https://gnosis-safe.io/app/eth:0x576cD258835C529B54722F84Bb7d4170aA932C64/transactions/queue).

Run Part 1 with:
```
brownie run scripts/issue/71/consolidate_badger.py main false true
```

### PART 2: Releases BADGER and redirects
- Releases all the BADGER from the DAO_treasury into the devMulti
- Checks for the deficit between the bBADGER strat staked balance and the Geyser balance and fills out the Geyser to this amount
- Calls `controller.withdrawAll(BADGER)` to move all BADGER from the strat to the vault (Reverts unless there is enough balance on the Geyser to withdraw, hence the previous step).
- Transfer the remaining amount of BADGER obtained from the DAO_treasury  into the Treasury Vault multi.

All of the above is performed atomically for better security.

You can simulate the upgrade of the DAO_treasury and the execution of PART 2 with the following:
```
brownie run scripts/issue/71/consolidate_badger.py main false true
```
Run Part 2 with:
```
brownie run scripts/issue/71/consolidate_badger.py main false
```

**NOTE:** PART 1 must execute in order to be able to load PART 2.

### DIFF
--- a/helpers/addresses.py
+++ b/helpers/addresses.py
@@ -15,6 +15,7 @@ ADDRESSES_ETH = {
     "guardian": "0x6615e67b8B6b6375D38A0A3f937cd8c1a1e96386",
     "GatedMiniMeController": "0xdDB2dfad74F64F14bb1A1cbaB9C03bc0eed74493",
     "GlobalAccessControl": "0x9c58B0D88578cd75154Bdb7C8B013f7157bae35a",
+    "badger_geyser": "0xBD9c69654B8F3E5978DFd138B00cB0Be29F28cCf",
     # the wallets listed here are looped over by scout and checked for all treasury tokens
     "badger_wallets": {
         "fees": "0x8dE82C4C968663a0284b01069DDE6EF231D0Ef9B",
@@ -68,6 +69,7 @@ ADDRESSES_ETH = {
         "devUngatedProxyAdmin": "0x9215cBDCDe25629d0e3D69ee5562d1b444Cf69F9",
         "testProxyAdmin": "0xB10b3Af646Afadd9C62D663dd5d226B15C25CdFA",
         "techOpsProxyAdmin": "0x7D0398D7D7432c47Dffc942Cd097B9eA3d88C385",
+        "opsProxyAdmin_old": "0x4599F2913a3db4E73aA77A304cCC21516dd7270D",
         "badgerHunt": "0x394DCfbCf25C5400fcC147EbD9970eD34A474543",
         "rewardsEscrow": "0xBE838aE7f6Ba97e7Eb545a3f43eE96FfBb3184DC",
     },
@@ -263,6 +265,7 @@ ADDRESSES_ETH = {
         "SettV1h": "0x9376B47E7eC9D4cfd5313Dc1FB0DFF4F61E8c481",
         "SettV1_1h": "0x25c9BD2eE36ef38992f8a6BE4CadDA9442Bf4170",
         "SettV4h": "0x0B7Cb84bc7ad4aF3E1C5312987B6E9A4612068AD",
+        "SimpleTimelockWithVoting": "0xb7AcD34643181C879437c2967538D5c0eA42b5D9", # V1.1 -> Beneficiary: devMulti
     },
     "guestlists": {
         "bimBTC": "0x7feCCc72aE222e0483cBDE212F5F88De62132546",
--- a/scripts/issue/71/consolidate_badger.py
+++ b/scripts/issue/71/consolidate_badger.py
@@ -45,8 +45,8 @@ def main(upgrade="true", simulation="false"):
         # Beneficiary is currently the DAO_Agent
         assert dao_treasury.beneficiary() == registry.eth.badger_wallets.fees
 
-        # Nonces are messed up on this multi - next in line is 78
-        old_multi.post_safe_tx(replace_nonce=78)
+        # Nonces are messed up on this multi - replace 76 (on-chain rejection)
+        old_multi.post_safe_tx(replace_nonce=76)
 
     else:
         # Simulates upgrade of contract
@@ -115,28 +115,27 @@ def main(upgrade="true", simulation="false"):
         balance_checker.verifyBalance(
             badger_token.address,
             BADGER_GEYSER,
-            geyser_deficit
+            geyser_deficit + geyser_balance
         )
 
         # Once the Geyser has enough BADGER within, we can
         # withdrawAll to transfer all BADGER from strat to the vault
-        assert bBadger_vault.available() == 0
-
         controller.withdrawAll(bBadger_vault.token())
 
-        assert bBadger_vault.available() == strategy_balanceOf
+        assert bBadger_vault.available() >= strategy_balanceOf
         assert bBadger_strat.balanceOf() == 0
         assert badger_token.balanceOf(BADGER_GEYSER) == 0
 
         # Transfer remaining of amount received from the DAO_treasury to
         # the treaury vault multisig
         amount = treasury_balance - geyser_deficit
+        treasury_vault_balance = badger_token.balanceOf(TREASURY_VAULT)
         badger_token.transfer(TREASURY_VAULT, amount, {"from": safe.address})
 
         balance_checker.verifyBalance(
             badger_token.address,
             TREASURY_VAULT,
-            amount
+            amount + treasury_vault_balance
         )
 
         safe.print_snapshot()
--- a/scripts/issue/71/consolidate_badger.py
+++ b/scripts/issue/71/consolidate_badger.py
@@ -1,0 +1,145 @@
+from great_ape_safe import GreatApeSafe
+from helpers.addresses import registry
+from brownie import interface
+from rich.console import Console
+
+CONSOLE = Console()
+
+DEV_MULTISIG = registry.eth.badger_wallets.dev_multisig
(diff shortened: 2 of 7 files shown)


scripts/issue/71/consolidate_badger.py
```diff
@@ -0,0 +1,145 @@
...
+        balance_checker.verifyBalance(
+            badger_token.address,
+            DEV_MULTISIG,
+            treasury_balance + gov_balance
+        )
+
+        ## == Transfer missing BADGER to Geyser and withdrawAll to vault == ##
+        geyser_balance = badger_token.balanceOf(BADGER_GEYSER)
+        strategy_balanceOf = bBadger_strat.balanceOf()
+        geyser_deficit = strategy_balanceOf - geyser_balance
```
jorijnsmit, line 105: i would add some extra room here, in case of activity on the strat. like `* 1.1` or something?

scripts/issue/71/consolidate_badger.py
```diff
@@ -0,0 +1,145 @@
...
+        controller.withdrawAll(bBadger_vault.token())
+
+        assert bBadger_vault.available() == strategy_balanceOf
+        assert bBadger_strat.balanceOf() == 0
+        assert badger_token.balanceOf(BADGER_GEYSER) == 0
+
+        # Transfer remaining of amount received from the DAO_treasury to
+        # the treaury vault multisig
+        amount = treasury_balance - geyser_deficit
+        badger_token.transfer(TREASURY_VAULT, amount, {"from": safe.address})
```
jorijnsmit, line 134: this cannot be done precise onfork, i would opt for sending 8_500_000 and getting the remainder in another tx.

helpers/addresses.py
```diff
@@ -263,6 +265,7 @@
         "SettV1h": "0x9376B47E7eC9D4cfd5313Dc1FB0DFF4F61E8c481",
         "SettV1_1h": "0x25c9BD2eE36ef38992f8a6BE4CadDA9442Bf4170",
         "SettV4h": "0x0B7Cb84bc7ad4aF3E1C5312987B6E9A4612068AD",
+        "SimpleTimelockWithVoting": "0xb7AcD34643181C879437c2967538D5c0eA42b5D9", # V1.1 -> Beneficiary: devMulti
```
jorijnsmit, line 268: i confirmed this is the new logic with dev msig as beneficiary of `release()`

jorijnsmit (changes requested):
crazy to think these addresses werent in the list yet! good additions of interfaces.

only thing i would note is that this is onfork, not onchain, so some numbers wont be precise. there are still withdrawals occurring every couple of hours: https://etherscan.io/address/0x19d97d8fa813ee2f51ad4b4e04ea08baf4dffc28

one last thing; do we have to make sure no earns are called anymore on the vault? it is important that the strat and geyser remain disabled from now on.

scripts/issue/71/consolidate_badger.py
```diff
@@ -0,0 +1,145 @@
...
+
+        assert treasury_balance == badger_token.balanceOf(dao_treasury.address)
+        assert token == dao_treasury.token()
+        assert releaseTime == dao_treasury.releaseTime()
+        assert beneficiary == dao_treasury.beneficiary()
+        # Beneficiary is currently the DAO_Agent
+        assert dao_treasury.beneficiary() == registry.eth.badger_wallets.fees
+
+        # Nonces are messed up on this multi - next in line is 78
+        old_multi.post_safe_tx(replace_nonce=78)
```
petrovska-petro, line 49: I am thinking that in this msig if the first nonce is going to be anyway rejected (*76*), perhaps is worthy to directly replace nonce 76, instead of paying for a rejection, thoughts?

helpers/addresses.py
petrovska-petro: +1

scripts/issue/71/consolidate_badger.py
```diff
@@ -0,0 +1,145 @@
...
+            treasury_balance + gov_balance
+        )
+
+        ## == Transfer missing BADGER to Geyser and withdrawAll to vault == ##
+        geyser_balance = badger_token.balanceOf(BADGER_GEYSER)
+        strategy_balanceOf = bBadger_strat.balanceOf()
+        geyser_deficit = strategy_balanceOf - geyser_balance
+
+        # Ensure geyser is in deficit
+        assert geyser_deficit > 0
```
petrovska-petro, line 108: why this deficit occurred? 🤔

scripts/issue/71/consolidate_badger.py
```diff
@@ -0,0 +1,145 @@
...
+        
+        # Transfer missing BADGER to Geyser
+        badger_token.transfer(BADGER_GEYSER, geyser_deficit, {"from": safe.address})
+
+        CONSOLE.print(f"{geyser_deficit / 1e18} BADGERs were transfer to the Geyser!\n")
+
+        balance_checker.verifyBalance(
+            badger_token.address,
+            BADGER_GEYSER,
+            geyser_deficit
```
petrovska-petro, line 118: should not this be `geyser_deficit + geyser_balance`?

scripts/issue/71/consolidate_badger.py
```diff
@@ -0,0 +1,145 @@
...
+        )
+
+        # Once the Geyser has enough BADGER within, we can
+        # withdrawAll to transfer all BADGER from strat to the vault
+        assert bBadger_vault.available() == 0
```
petrovska-petro, line 123: This line in case that there is idle badger in the sett will revert

ref: https://etherscan.io/address/0xe4ae305b08434bf3d74e0086592627f913a258a9#code#L1254

for example now it is greater than 0, there is 668 badger

Screenshot 2022-02-08 at 14 16 07

scripts/issue/71/consolidate_badger.py
petrovska-petro: 2 txs approach sounds safest to avoid reverts from miscalc off-chain till signing/executing

scripts/issue/71/consolidate_badger.py
```diff
@@ -0,0 +1,145 @@
...
+
+        # Transfer remaining of amount received from the DAO_treasury to
+        # the treaury vault multisig
+        amount = treasury_balance - geyser_deficit
+        badger_token.transfer(TREASURY_VAULT, amount, {"from": safe.address})
+
+        balance_checker.verifyBalance(
+            badger_token.address,
+            TREASURY_VAULT,
+            amount
```
petrovska-petro, line 139: vault also already has some badger, more strict will be its balance  + amount transfer

scripts/issue/71/consolidate_badger.py
sajanrajdev: Very good call, I missed that.

scripts/issue/71/consolidate_badger.py
sajanrajdev: Good idea, I don't see the harm. I confirmed that 76 is obsolete and should be rejected. Will make the change.

scripts/issue/71/consolidate_badger.py
sajanrajdev: Yes, you are right, I should avoid writing scripts like this towards the end of the day... good catch!

scripts/issue/71/consolidate_badger.py
sajanrajdev: You are right, specially if we turn earn() off, BADGER could start accumulating. Will remove the assertion.

scripts/issue/71/consolidate_badger.py
```diff
@@ -0,0 +1,145 @@
...
+            geyser_deficit
+        )
+
+        # Once the Geyser has enough BADGER within, we can
+        # withdrawAll to transfer all BADGER from strat to the vault
+        assert bBadger_vault.available() == 0
+
+        controller.withdrawAll(bBadger_vault.token())
+
+        assert bBadger_vault.available() == strategy_balanceOf
```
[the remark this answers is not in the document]
sajanrajdev: Good call

scripts/issue/71/consolidate_badger.py
sajanrajdev: 100% also missed this one.

scripts/issue/71/consolidate_badger.py
sajanrajdev: Confirmed: Running the script yesterday, estimated a deficit of: `201661.80644761855 BADGER`. Today, after a few withdrawals already, the deficit is still the same: `201661.80644761855 BADGER`.

btcookies: @sajanrajdev this is to turn off earns for bbadger?

sajanrajdev: @btcookies, yes, we are deprecating the strat so there's no point on keep earning it

petrovska-petro (approved):
Suggestions were introduced ✅

Issue #136, opened 2022-02-16
week 7: tree BADGER top-up 


petrovska-petro:
BADGER emissions comes as 34565.28490125317 + 1w of remBADGER emissions 7692.307692

Coverage for 17th-24th Feb = `42257.592593253175 BADGER`

Ideally to be included in an updated version of #123

jorijnsmit:
https://etherscan.io/tx/0xc600c1a297733019c8da7ce8c73521f15cd74b417015d1dec6edce0e3516dd18

Issue #112, opened 2022-02-08
week 7: bi-weekly direct transfer into remBADGER sett


petrovska-petro:
- Send 11,538.461538 BADGER to [remBADGER sett](https://etherscan.io/address/0x6af7377b5009d7d154f36fe9e235ae1da27aea22)

jorijnsmit:
https://etherscan.io/tx/0xc600c1a297733019c8da7ce8c73521f15cd74b417015d1dec6edce0e3516dd18

PR #138, opened 2022-02-16
Include w7 badger topup and biweekly transfer to remBADGER sett

petrovska-petro:
Tackles #112 and #136

Command:

```
brownie run badger/topups/week_6_and_7
```


jorijnsmit (approved):
adds topups for week 7 as well, for both badger tree and rembadger vault.

Issue #131, opened 2022-02-14
do one last migration of all tokens in dev msig once all fees have been redirected


jorijnsmit:
do not forget fusdc-22 and fwbtc-22 tokens.

depends on #57.

jorijnsmit:
https://etherscan.io/tx/0xb3148dac4d7e177012f155a635e3aec27a0bc32343f11ba875fc18860274a85b

PR #140, opened 2022-02-16
get rid of decimal column, add some more tokens to scan for

jorijnsmit:
this is to solve #131:
```
brownie run gnosis/scrape_tokens scrape 0xb65cef03b9b89f99517643226d76e286ee999e77 0x042b32ac6b453485e357938bdc38e0340d4b9276
```


petrovska-petro (approved):
Includes sett_tokens and rari_tokens, csv file includes the balances from `fWBTC-22` & `fUSDC-22` to be sent to treasury ops for processing and removes decimal column ✅

petrovska-petro:
@jorijnsmit maybe it is great also to include in this PR the csv output, right?

jalbrekt85 (approved):
looks good

jorijnsmit:
good call @petrovska-petro, adding the csv that actually solves #131.

petrovska-petro (approved):
Got same csv output till row **24**, 25 & 26 got same **value**, but as they are not going to be process they are going to the vault, content looks good !

scripts/gnosis/scrape_tokens.py
```diff
@@ -15,46 +15,47 @@ def main():
...
             table.add_column(key, justify="right", style="bright_yellow")
-            
-        for token_name in registry.eth.treasury_tokens:
-            token = interface.ERC20(registry.eth.treasury_tokens[token_name])
-            decimals = token.decimals()
+
+        for token_addr in token_list:
+            if token_addr in [registry.eth.rari.dai_manager, registry.eth.rari.unitroller]:
+                continue
+            token = Contract(token_addr)
```
jalbrekt85, line 43: running this for the first time takes about ~5 minutes. Any reason to use `Contract` over `interface.ERC20` here?

jorijnsmit: the problem with interface.erc20 in this case is that there are also lp tokens included in that for loop. this means that if you call the lp token at a later time again with Contract(), it will revert to that interface.erc20 it was locally saved as, and miss attributes.

jalbrekt85 (approved):

PR #30, opened 2022-01-24
send all dust to treasury ops, excl $digg/$badger/$across-badger

jorijnsmit:
- leaving badger and digg on the wallet for emissions/quick corrections that might be needed
- across-badger position of 20k badger is moved to treasury vault
- all other 'dust' goes to treasury ops for processing

built gnosis airdrop csv using #23 

issue #28

Issue #28, opened 2022-01-24
sweep techops dust into treasury ops


jorijnsmit:
to get it all ready to process in one go.

petrovska-petro:
tx: https://etherscan.io/tx/0x7a030e8add06993a57f6559d45ecd2b722297b758b39880d77f8a2205c9c53ac

PR #144, opened 2022-02-17
changes in balances on techops for sweep

jorijnsmit:
solves #28, replaces #30.

basically drain techops into trops, except for badger/eth. also abadger position is sent straight to treasury vault.


petrovska-petro (approved):
Update csv output, badger stays, rest is going to treasury ops for processing, except *aBADGER* going to vault

Tritium-VLK (approved):

petrovska-petro:
@jorijnsmit checking again as i got same output, there is duplicates in this csv for two tokens: 

0x6dEf55d2e18486B9dDfaA075bc4e4EE0B28c1545 and 0xfd05D3C7fe2924020620A8bE4961bBaA747e6305

probably since it is looping thru them as some tokens are duplicate in treasury tokens and sett_vaults, it also happened to me as i used the version from this [PR](https://github.com/Badger-Finance/badger-multisig/blob/issue/131/scripts/gnosis/scrape_tokens.py#L29)

PR #149, opened 2022-02-18
fix: remove duplicates from `token_list`

jorijnsmit:
turn list into set to remove duplicate lp tokens present in both treasury_tokens and sett_vaults.


petrovska-petro (approved):
Great usage of `set` avoid duplicates for tokens such as `bveCVX`

Issue #156, opened 2022-02-22
dogfood all of treasury's bvecvxcvxf


jorijnsmit:
since yield redirection will be implemented soon (https://github.com/Badger-Finance/badger-rewards/issues/420), dogfooding the treasury's bvecvxcvx curve lp token makes sense now since it will have a claim on the sett's non-badger rewards.

https://etherscan.io/token/0x04c90c198b2eff55716079bc06d7ccc4aa4d7512?a=0xd0a7a8b98957b9cd3cfb9c0425abe44551158e9e

jorijnsmit:
posted nonce 8: 
```
snapshot result for 0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e:
                          balance_before              balance_after               balance_delta
symbol                                                                                         
bveCVX-CVX-f   32,086.270561381178524219       0.000000000000000000  -32,086.270561381178524219
bbveCVX-CVX-f       0.000000000000000000  32,086.270561381178524219   32,086.270561381178524219 
```
```
Events In This Transaction
--------------------------
├── Curve.fi Factory Plain Pool: Badger Locked CVX (0x04c90C198b2eFF55716079bc06d7CCc4aa4d7512)
│   ├── (unknown)
│   │   ├── topic1: 0x8c5be1e5ebec7d5bd14f71427d1e84f3dd0314c0f7b2291e5b200ac8c7c3b925
│   │   ├── topic2: 0x000000000000000000000000d0a7a8b98957b9cd3cfb9c0425abe44551158e9e
│   │   ├── topic3: 0x000000000000000000000000937b8e917d0f36edebba8e459c5fb16f3b315551
│   │   └── data: 0x0000000000000000000000000000000000000000000006cb6674533a96adc23b
│   └── (unknown)
│       ├── topic1: 0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef
│       ├── topic2: 0x000000000000000000000000d0a7a8b98957b9cd3cfb9c0425abe44551158e9e
│       ├── topic3: 0x000000000000000000000000937b8e917d0f36edebba8e459c5fb16f3b315551
│       └── data: 0x0000000000000000000000000000000000000000000006cb6674533a96adc23b
│   
├── Badger Sett Curve.fi Factory Plain Pool: Badger Locked CVX (0x937B8E917d0F36eDEBBA8E459C5FB16F3b315551)
│   └── Transfer
[...]

PR #160, opened 2022-02-22
deposit all bveCVX lp into bbveCVX

jalbrekt85:
**#156**

### DIFF
--- a/scripts/issue/156/issue156.py
+++ b/scripts/issue/156/issue156.py
@@ -19,7 +19,7 @@ def main():
         owner=safe.account
     )
 
-    safe.take_snapshot(tokens=[bvecvx_cvx_lp.address, bbveCVX.address])
+    safe.take_snapshot(tokens=[bvecvx_cvx_lp.address, bbveCVX_sett.address])
 
     bvecvx_cvx_lp.approve(bbveCVX_sett, bvecvx_cvx_lp.balanceOf(safe))
     bbveCVX_sett.depositAll()


jorijnsmit (commented):
can you add snapshots so we can verify it visually at the end?

jorijnsmit (approved):
script does what it needs to do, except for fails in the call trace:
```
    │           └── SettV4h.depositAll  [DELEGATECALL]  2104:2504  [84990 / 8274 gas]
    │               │   ├── address: 0x0B7Cb84bc7ad4aF3E1C5312987B6E9A4612068AD
    │               │   ├── input arguments: None
    │               │   └── revert reason: Access denied for caller
```

 we will have to whitelist treasury vault msig first on bbvecvxcvx before running this. i still suggest to merge it though, and have it ready to post.

scripts/issue/156/issue156.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+def main():
+    """
+    deposit all bveCVX lp to bbveCVX sett
+    """
+
+    safe = GreatApeSafe(registry.eth.badger_wallets.treasury_vault_multisig)
+
+    bveCVX = interface.ICurveLP(
```
petrovska-petro, line 13: The variable naming was confusing for me, mixing it up with the real `bvecvx` sett token, could this be renamed to reflect the bvecvx_cvx_lp instead?

jalbrekt85: agreed, these token names can get confusing

jorijnsmit (approved):

petrovska-petro (approved):

PR #165, opened 2022-02-23
solve name collision on curve v1 and v2 registry

jorijnsmit:
had to do some hot fixes yesterday in order to run issue34.py


jalbrekt85 (approved):
name collision problem is fixed. added badger with `MintableForkToken` and curve deposit script works perfectly

petrovska-petro (approved):
Solved registry naming collision ✅

PR #168, opened 2022-02-23
withdraw 9 digg from tcl, add dest feat to univ2 class

jorijnsmit:
solve #159 

withdraws exactly 9 digg from the treasury vault slp (and corresponding wbtc) to the treasury ops wallet.

added destination parameter to univ2 class methods.

PR #171, opened 2022-02-23
add method to uni_v2 to calc lp tokens to withdraw given underlying

jalbrekt85:
helper function to calculate the amount of lp tokens that should be withdrawn in order to get `x` amount of underlying tokens out

solution to do this is taken from @jorijnsmit in #168 and abstracted to a helper function as it will probably be useful in the future


great_ape_safe/ape_api/uni_v2.py
```diff
@@ -13,6 +14,17 @@ def __init__(self, safe):
         self.max_weth_unwrap = 0.01
         self.deadline = 60 * 60 * 12
 
+
+    def get_lp_to_withdraw_given_token(self, lp_token, underlying_token, mantissa_underlying):
+        # calc amount of `lp_token` to withdraw from pool to get `mantissa_underlying` of `underlying_token`
+        # credit: https://github.com/Badger-Finance/badger-multisig/blob/a0eab1de153d99fd00bb696ba93ba1fab60a1266/scripts/issue/159/withdraw_9_digg_from_tcl.py
+        x = Symbol('x')
+        reserve_index = 0 if lp_token.token0() == underlying_token else 1
```
jorijnsmit, line 22: maybe assert token1 == underlying_token or something?

just to make sure the underlying_token is actually present in the lp_token. otherwise i am guessing func will still return some amount and script might continue without us catching the mistake. better to raise so that we are aware.

jorijnsmit (commented):
good idea to add this to the library.

saw there is a merge conflict though, can you resolve?

great_ape_safe/ape_api/uni_v2.py
jalbrekt85: good catch

jorijnsmit: nice use of .index

jorijnsmit (approved):

petrovska-petro (approved):
Nice addition!

Issue #164, opened 2022-02-22
whitelist treasury vault on bbvecvxcvx


jorijnsmit:
see #160

also consider to whitelist other setts for which that is not the case yet

jorijnsmit:
posted nonce 1008: https://gnosis-safe.io/app/eth:0xB65cef03b9B89f99517643226d76e286ee999e77/transactions/0xa5a0509a6c0c3e9abf04c00a7cbe8add0a2f2f2a2c3bfdd7e039f90db55b0278
```
whitelisted/queued:
[
    'bBADGER',
    'bDIGG',
    'bharvestcrvRenBTC',
    'buniWbtcBadger',
    'bslpWbtcBadger',
    'bslpWbtcibBTC',
    'buniWbtcDigg',
    'bslpWbtcDigg',
    'bslpWbtcEth',
    'bcrvIbBTC',
    'bcrvTricrypto',
    'bcrvTricrypto2',
    'bcvxCRV',
    'bCVX',
    'bveCVX',
    'bimBTC',
    'bFpMbtcHbtc',
    'bbveCVX-CVX-f',
    'remBADGER',
    'remDIGG'
]
```
```
Events In This Transaction
--------------------------
└── Gnosis Safe (0xB65cef03b9B89f99517643226d76e286ee999e77)
    └── ExecutionSuccess
        ├── txHash: 0x2b6655496c96d49a8b66543df3545681c414e9c31d46f3c3b47b615c7f966017
        └── payment: 0

{
│   'ethereum_client': <gnosis.eth.ethereum_client.EthereumClient object at 0x10d4b7d60>,
│   'safe_address': '0xB65cef03b9B89f99517643226d76e286ee999e77',
│   'to': '0x40A2aCCbd92BCA938b02010E17A5b8929b49130D',
│   'value': 0,
[...]

PR #172, opened 2022-02-23
whitelist treasury vault multisig to all setts

jalbrekt85:
solves #164, small tweaks to `Badger` and `whitelist.py`. ~~script approves msig to all setts, not sure if just specifying certain setts would be better instead.~~

whitelist treasury vault as so:
`brownie run badger/whitelist all_setts 0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e`


jorijnsmit (commented):
are these __init__ files needed to import like you did in issue164.py:L1?

because if so, it might be cleaner to just run it directly from console:
`brownie run badger/whitelist all_setts 0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e`

and i think forlooping all setts is the way to go here, im sure there will be other moments when we will need to interact with them. might as well whitelist ourselves on all of them.

jalbrekt85:
> are these **init** files needed to import like you did in issue164.py:L1?
> 
> because if so, it might be cleaner to just run it directly from console: `brownie run badger/whitelist all_setts 0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e`
> 
> and i think forlooping all setts is the way to go here, im sure there will be other moments when we will need to interact with them. might as well whitelist ourselves on all of them.

I wasn't able to get brownie to recognize the imports without the init files. don't know why I didn't just think of running from cmd, I agree that is cleaner

petrovska-petro (approved):
Smooth, good idea including output summary of setts approved

Issue #32, opened 2022-01-24
deploy next lower range for univ3 tcl


jorijnsmit:
the `.000063 - .000125` range has to be deployed according to bip-74: https://forum.badger.finance/t/bip-74-treasury-controlled-liquidity-for-badger/5082/

requires 57.74 wbtc, which needs to be funded from vault, but nft position itself could be created on treasury ops i guess.

lowest range deployed is currently in range: https://app.uniswap.org/#/pool/167046

pool: https://info.uniswap.org/#/pools/0xe15e6583425700993bd08f51bf6e7b73cd5da91b

petrovska-petro:
Waiting for funds to reach on treasury ops once [nonce 6](https://gnosis-safe.io/app/eth:0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e/transactions/queue) executes, awaiting signers

jorijnsmit:
think this can be closed for now, no? good to have #100 merged and ready, but this ticket is not needed for now..?

petrovska-petro:
> think this can be closed for now, no? good to have #100 merged and ready, but this ticket is not needed for now..?

we could close it, but it will be just to open again in the future the same, right?

depending on market conditions at the end till it becomes relevant again, but not see damage on leaving it open

petrovska-petro:
snap:

```
snapshot result for 0x042B32Ac6b453485e357938bdC38e0340d4b9276:
                         balance_before         balance_after           balance_delta
symbol                                                                               
bibbtc/sbtcCRV-f  21.040302707297603800  0.000000000000000000  -21.040302707297603800
WBTC  
```

```
{
│   'ethereum_client': <gnosis.eth.ethereum_client.EthereumClient object at 0x7fd2a0d72340>,
│   'safe_address': '0x042B32Ac6b453485e357938bdC38e0340d4b9276',
│   'to': '0x40A2aCCbd92BCA938b02010E17A5b8929b49130D',
│   'value': 0,
[...]

PR #192, opened 2022-03-01
Chore: Updating script as was run yesterday

petrovska-petro:
ref: https://github.com/Badger-Finance/badger-multisig/issues/32#issuecomment-1054628718

Fix the range, wbtc amount deposited and wd all from ibbtc sett


jorijnsmit (approved):

jalbrekt85 (approved):
new range set and accounted for slippage ✅

Issue #187, opened 2022-02-28
make cf on rari fuse 22 uniform


jorijnsmit:
current situation:
- 50%: ibbtc
- 65%: badger, digg
- 75%: wbtc, usdc, fei, eth, dai, frax
- 85%: dola

Tritium-VLK:
Suggestion:  75% for USD stables, BTC and ETH (large-cap crypto)
65% for small cap tokens: Badger, Digg
50% for large cap vault or interest baring tokens: ibBTC/ibBTClp
35% for low AUM  vault tokens:  Tricrypto, bveCVX

In the end @petrovska-petro is taking ownership over this and should set the final guidelines.

jorijnsmit:
https://etherscan.io/tx/0x4bbb97eb6e34ead3b92d90b19f3731fbbfa682564b296d9d84819da48149482a

PR #206, opened 2022-03-03
set cf of remaining stables in Rari Fuse to 85%

jalbrekt85:
**for #187**


jorijnsmit (approved):

petrovska-petro (approved):
CF change for stables looks good ✅

Issue #56, opened 2022-01-28
general dogfood script that can be run monthly/quarterly


jorijnsmit:
wait until all applicable msig tree rewards are redirected to treasury ops (https://github.com/Badger-Finance/badger-rewards/issues/539)

should do the following, indifferent of actual balances:
- extract wbtc/renbtc/*btc from every possible lp (uni, sushi, crv)
- swap to wbtc or mint ibbtc (depending on arb oppo in the ibbtcbtccrv-f pool?)
- deposit into ibbtcsbtccrv-f
- dogfood it into the bibbtc vault using depositfor(vault)

bonus:
- unwrap bdigg and bbadger into naked tokens so they are ready for emissions
- unwrap any weth
- dogfood cvxcrv
- dogfood cvx (either swap for bvecvx or lock into bvecvx)
- deposit bvecvx into curve pool if balance cvx>bvecvx to bring back to 1:1

TODO
- think about threshold per position before processing

jorijnsmit:
https://gnosis-safe.io/app/eth:0x042B32Ac6b453485e357938bdC38e0340d4b9276/transactions/0x61489dc0900b650021b6e836864a0c8add1c871a801b149191a3dad6c255e326
```
{
│   'ethereum_client': <gnosis.eth.ethereum_client.EthereumClient object at 0x10881f8b0>,
│   'safe_address': '0x042B32Ac6b453485e357938bdC38e0340d4b9276',
│   'to': '0x40A2aCCbd92BCA938b02010E17A5b8929b49130D',
│   'value': 0,
[...]

PR #214, opened 2022-03-04
amendmends needed to successfully run thresher

jorijnsmit:
see #56


petrovska-petro (approved):
Added more conditionals to only run certain scenarios when it is > 0

jalbrekt85 (approved):
nice changes that can reduce multicall complexity and allow thresher script to be execd more consistently

Issue #183, opened 2022-02-28
month 3: claim and airdrop EPS for month of February


petrovska-petro:
- [x] Claim February month

- [x] Update script distributor to cut the recipients if less than a dollar is going to be given as seems the consensus

- [ ] Distribute EPS

petrovska-petro:
*TOTAL EPS CLAIMED 59,356.615218518221809063*

```
{
│   'ethereum_client': <gnosis.eth.ethereum_client.EthereumClient object at 0x7f9ffd5a7520>,
│   'safe_address': '0x329543f0F4BB134A3f7a826DC32532398B38a3fA',
│   'to': '0x40A2aCCbd92BCA938b02010E17A5b8929b49130D',
│   'value': 0,
[...]

PR #218, opened 2022-03-07
Prep for discussions of adding threshold $1 in distribution

petrovska-petro:
Tackles 2nd point in #183


jorijnsmit (approved):
makes sense, saves lots of rows.

only cavity is to watch for $eps volatility and have csv generation and actual airdrop as close to each other as possible.

jalbrekt85 (approved):
accurately filters out entries that are less than threshold of $1

Issue #216, opened 2022-03-07
Include signing & exec methods using hardware wallet in greatape


petrovska-petro:
- [x] Test in rinkeby safe flow

- [x] Include sign with frame in great ape and posting signature

- [x] Exec method thru frame

petrovska-petro:
Comments:

- Signing from brownie/frame was smooth. Only caveat is that if signing occurs after the tx posting, we will need to specify in `multisend_from_receipts` method the **safe nonce**, so it produces the same safe tx hash and points to the right one [here](https://github.com/banteg/ape-safe/blob/master/ape_safe.py#L253)

- Exec was getting stuck with `GS020`. Ref: https://github.com/gnosis/safe-contracts/blob/main/docs/error_codes.md#general-signature-validation-related . Despite of being able to exec with MM and two owners signatures being correct. Likely related with the gas estimation for the thru this [method](https://github.com/banteg/ape-safe/blob/master/ape_safe.py#L350). Workaround for the payload...?

Txs rinkeby, 2/2 policy:

- https://rinkeby.etherscan.io/tx/0xe6543e4111293add22062b96c3cf6b3f0eac1be0d3af765e4d37521a01b493d5

jorijnsmit:
solved in #220

PR #220, opened 2022-03-08
Testing frame hardware wallet flow 

petrovska-petro:
Tackles the addition of the signing and exec method as #216 

Signing smooths for 1.3 safes, exec still need others to replicate and confirm issue with gas estimation

Posting:

```
brownie run --network rinkeby-fork issue/216/frame_hardware_testing main 0x... 0x... 0x...
```

For signing will be needed to pass the nonce of your safe tx hash posted using default action_trigger flag:

```
brownie run --network rinkeby-fork issue/216/frame_hardware_testing main 0x... 0x... 0x... 1 0
```

Exec once policy of signer is meet:

```
brownie run --network rinkeby-fork issue/216/frame_hardware_testing main 0x... 0x... 0x... 2 0
```


petrovska-petro:
For testing latest changes ref to this [comment](https://github.com/Badger-Finance/badger-multisig/pull/231#issuecomment-1063337653)

jorijnsmit (approved):
so my conclusion in regards to both v1.1 and v1.3 safes, is that in both cases frame underestimates the gas.

even though i introduced the `gas_coef` param, setting this higher will not solve the issue.

**in all cases, when execing with frame, the gas limit on the actual exec has to be set higher (2x?) then what it itself estimates.**

jalbrekt85 (approved):
tested signing through frame, worked smoothly

PR #224, opened 2022-03-08
fix arg data type bug and lock for vault instead of ops

jorijnsmit:


petrovska-petro (approved):
- Change to default 98% of quoted and sending directly dao fee to vault as discussed in tg

- T now is supported and included speel in the swaps💪

jalbrekt85 (approved):

Issue #226, opened 2022-03-08
Add bveCVX into rari pool


petrovska-petro:
- [x] Deploy fbveCVX-22 with reserve factor 0%, admin fee 0%, admin as dev_msig and interest model jumpratemodel_fei_dai

- [x] Include the following check-up/changes on the methods in rari class to [enable](https://github.com/Badger-Finance/badger-multisig/blob/main/great_ape_safe/ape_api/rari.py#L19)/[disable](https://github.com/Badger-Finance/badger-multisig/blob/main/great_ape_safe/ape_api/rari.py#L24) borrowing of an asset

Add an extra check-up that if an asset whenever is going to be switch from disable to enable that the reserve factor is set to default to 15%  (but allow to pass this as arg), but on the case of going from enable to disable to set the reserve factor to 0%. Since, it does not make sense to have a reserve factor > 0, while an asset is not borrowable as nothing will accrue in the reserve.

- [x] Support the market and set CF to 50%, in this occasion atomically disable borrowing thru [controller](https://etherscan.io/address/0x94b2200d28932679def4a7d08596a229553a994e#code#F8#L1207). `fbveCVX-22`should **not** be borrowable

petrovska-petro:
safe tx: https://gnosis-safe.io/app/eth:0xB65cef03b9B89f99517643226d76e286ee999e77/transactions/0x407d4c2babf671323d177d76b153af973c37f2e29224f2eca2d0389b161e6ef1

jorijnsmit:
fyi this got rejected because dev msig had to execute on setting cf for $dola to 0.

jorijnsmit:
```
Events In This Transaction
--------------------------
├── IRariComptroller (0xe3952d770FB26CC61877CD34Fbc3A3750881e9A1)
│   └── ActionPaused
│       ├── cToken: 0x26F6F27fDBc3B9cDe4b1943b1c07606CAF2c4C6C
│       ├── action: Borrow
│       └── pauseState: True
│   
├── Badger Pool Badger Vested Escrow Convex Token (0x26F6F27fDBc3B9cDe4b1943b1c07606CAF2c4C6C)
│   ├── AccrueInterest
│   │   ├── cashPrior: 0
│   │   ├── interestAccumulated: 0
│   │   ├── borrowIndex: 1000000000000000000
│   │   └── totalBorrows: 0
│   └── Failure
│       ├── error: 10
│       ├── info: 84
│       └── detail: 0
│   
[...]

PR #232, opened 2022-03-09
disable borrowing of fbveCVX

jalbrekt85:
~~**attempt to solve part of #226 where we add `fbveCVX` to our fuse pool.** getting a revert when calling `rari.add_ftoken_to_pool`, despite same method working fine when deploying fDOLA/fRAX~~

script now disables borrowing of new fbveCVX (address pending) and checks that relevant params are set, input fbveCVX address as input to run:

`brownie run issue/226/disable_borrow disable <fbveCVX address>`


petrovska-petro:
```
sri yantra — 03/11/2022
oh I see what the issue is. Etherscan is showing the old impl on its interface but your actual current comptroller implementation (see on the contract) is 0xe16db319d9da7ce40b666dd2e365a4b8b3c18217. _supportMarket() is internal there so you're going to have to use comptroller._deployMarket()
```

petrovska-petro:
New implementation for fbvecvx was exec: https://etherscan.io/tx/0x55bbdb8e08e9f22b7ec17cf24557599ca80dc3c1a7c7f09846d98a2b5155348f

great_ape_safe/ape_api/rari.py
```diff
@@ -17,13 +17,21 @@ def ftoken_is_paused(self, ftoken_addr):
...
+        ftoken = interface.IFToken(ftoken_addr, owner=self.safe.account)
         self.unitroller._setBorrowPaused(ftoken_addr, True)
+        if ftoken.reserveFactorMantissa() != 0:
+            ftoken._setReserveFactor(0)
         assert self.ftoken_is_paused(ftoken_addr)
+        assert ftoken.reserveFactor() == 0
 
 
-    def ftoken_unpause(self, ftoken_addr):
+    def ftoken_unpause(self, ftoken_addr, rf=0.15):
```
jorijnsmit, line 28: isnt a default `rf` of `.15` very low? looks to me like `.65` or even `.8` would make more sense. what is the default in that vercel ui app?

great_ape_safe/ape_api/rari.py
```diff
@@ -17,13 +17,21 @@ def ftoken_is_paused(self, ftoken_addr):
 
 
     def ftoken_pause(self, ftoken_addr):
+        ftoken = interface.IFToken(ftoken_addr, owner=self.safe.account)
         self.unitroller._setBorrowPaused(ftoken_addr, True)
+        if ftoken.reserveFactorMantissa() != 0:
+            ftoken._setReserveFactor(0)
         assert self.ftoken_is_paused(ftoken_addr)
+        assert ftoken.reserveFactor() == 0
```
jorijnsmit, line 25: what is the rationale behind resetting the rf to `0`? in my mind this just creates problems when unpausing, because you have to pass the right rf again for every asset.

scripts/issue/226/disable_borrow.py
```diff
@@ -0,0 +1,18 @@
+
+from great_ape_safe import GreatApeSafe
+from helpers.addresses import registry
+from brownie import interface
+from eth_abi import encode_abi
+
+
+def disable(ftoken_addr):
```
jorijnsmit, line 8: suggesting to keep using `main` as function name if there is only one function in the file; prevents from having to specify the function manually.

jorijnsmit (changes requested):
not sure about this setting the rf to 0. maybe i am missing something. perhaps another solution would be to have separate getters/setters for the rf?

just wondering about a use case where we want to say pause all ftokens for a week or whatever. including the rf resetter in the pause wrapper makes this more complicated than needed imo, but maybe i am missing something.

great_ape_safe/ape_api/rari.py
petrovska-petro: .15 rf is already on the higher end, would it makes sense to have something around ~0.05 as default instead

most pools which are more attractive has much lower than 15%, even some 0%, but i do not think so we need to go that down, as still is a protective measure

great_ape_safe/ape_api/rari.py
```diff
@@ -16,13 +16,19 @@ def ftoken_is_paused(self, ftoken_addr):
         return self.unitroller.borrowGuardianPaused(ftoken_addr)
 
 
-    def ftoken_pause(self, ftoken_addr):
+    def ftoken_pause(self, ftoken_addr, rf=None):
         self.unitroller._setBorrowPaused(ftoken_addr, True)
+        if rf:
```
petrovska-petro, line 21: perhaps here is better to check if it is diff than 'None' as passing zero, will not exec this conditional block

jalbrekt85: very nice catch

great_ape_safe/ape_api/rari.py
jorijnsmit: ok i see we use .15 as a default for all ftokens so makes sense

jorijnsmit (approved):
nice solution by setting the param `rf=None` by default.

petrovska-petro (approved):

PR #240, opened 2022-03-10
Update csv file with latest cut-off treshold using spot eps price

petrovska-petro:


jorijnsmit (approved):
some more wallets eligible due to $eps price going up since last push

jalbrekt85 (approved):
changes reflect latest eps price

Issue #245, opened 2022-03-11
Include in rari class methods to upgrade into a new implementation


petrovska-petro:
- [x] Method to upgrade into a new implementation the fTokens

- [x] Method to upgrade into a new implementation the comptroller

- [x] Script to upgrade`fbveCVX-22` into latest implementation `0x67Db14E73C2Dce786B5bbBfa4D010dEab4BBFCF9`

Rari's dev feedback: 

<img width="871" alt="Screenshot 2022-03-11 at 21 07 33" src="https://user-images.githubusercontent.com/84875062/157951932-9eaf6eb0-d668-479a-8d7f-4b0b2bfd4874.png">

PR #246, opened 2022-03-11
introduce methods to upgrade ftokens/comptroller, script to upgrade fbveCVX implementation

jalbrekt85:
solves #245 in response to #232 , upgrades fbveCVX implementation to = `0x67Db14E73C2Dce786B5bbBfa4D010dEab4BBFCF9` w/ new methods to do so, which will solve some compatibility issues and bring us one step closer to getting fbveCVX added to our rari pool

`brownie run issue/245/issue245.py`


great_ape_safe/ape_api/rari.py
```diff
@@ -59,3 +59,16 @@ def add_ftoken_to_pool(self, ftoken_addr, cf=None):
         else:
             self.unitroller._supportMarket(ftoken_addr)
         assert ftoken_addr in self.unitroller.getAllMarkets()
+
+
+    def upgrade_comptroller(self, implementation):
+        self.unitroller._setPendingImplementation(implementation)
+        assert self.unitroller.pendingComptrollerImplementation() == implementation
+        self.unitroller._acceptImplementation()
```
petrovska-petro, line 67: I believe this needs to be called instead from the new implementation, [ref](https://etherscan.io/address/0xe3952d770FB26CC61877CD34Fbc3A3750881e9A1#code#F23#L63)

method: https://etherscan.io/address/0xe16db319d9da7ce40b666dd2e365a4b8b3c18217#code#F9#L1406

jalbrekt85: yup, I believe this is right, that would make sense as to why there are separate methods to set pending implementation and accepting it

petrovska-petro:
Code looks good, just will be great to add into the `IRariComptroller.sol` the events related with `NewPendingImplementation` and `NewImplementation` so in the console when it shows the events in the transactions is able to show them nicely otherwise will be unknown and topic1 and data, which are hard to depict

jalbrekt85:
> Code looks good, just will be great to add into the `IRariComptroller.sol` the events related with `NewPendingImplementation` and `NewImplementation` so in the console when it shows the events in the transactions is able to show them nicely otherwise will be unknown and topic1 and data, which are hard to depict

fixed in latest commit, this is just setting the same implementation as lack of a better way to test, but looks better now:
```{
    'NewImplementation': OrderedDict([('oldImplementation', '0xE16DB319d9dA7Ce40b666DD2E365a4b8B3C18217'), ('newImplementation', 
'0xE16DB319d9dA7Ce40b666DD2E365a4b8B3C18217')]),
    'NewPendingImplementation': OrderedDict([('oldPendingImplementation', '0xE16DB319d9dA7Ce40b666DD2E365a4b8B3C18217'), 
('newPendingImplementation', '0x0000000000000000000000000000000000000000')])
}```

**Review by petrovska-petro — APPROVED** [review]
LGTM!

File: scripts/issue/245/issue245.py [hunk 2]
```diff
@@ -0,0 +1,26 @@
+from great_ape_safe import GreatApeSafe
+from helpers.addresses import registry
+from brownie import interface
+from rich.console import Console
+
+
+console = Console()
+
+NEW_IMPLEMENTATION = "0x67Db14E73C2Dce786B5bbBfa4D010dEab4BBFCF9"
```
jorijnsmit, line 9: confirmed that all other f22tokens have this address as their implementation.

also seen that it has been deployed months ago, i wonder how we ended up with the old version of the implementation for fbvecvx...?

great_ape_safe/ape_api/rari.py
```diff
@@ -59,3 +59,19 @@ def add_ftoken_to_pool(self, ftoken_addr, cf=None):
...
+
+
+    def upgrade_comptroller(self, implementation):
+        self.unitroller._setPendingImplementation(implementation)
+        assert self.unitroller.pendingComptrollerImplementation() == implementation
+        comptroller = interface.IRariComptroller(
+            implementation, owner=self.safe.account
+        )
+        comptroller._become(self.unitroller.address)
+        assert self.unitroller.comptrollerImplementation() == implementation
```
jorijnsmit, line 71: not sure why we would need this...?

Issue #21, opened 2022-01-20
bridge .007 eth to fantom msig


jorijnsmit:
for gas and testing bridge.

i guess it can be sourced from techops.

see #20 for fantom msig details.

jorijnsmit:
https://docs.multichain.org/developer-guide/how-to-integrate-front-end-bridges

jorijnsmit:
still waiting on response from support for whitelisting our multisig; bridge can only be used by contracts if whitelisted.

jorijnsmit:
update:
<img width="618" alt="Screenshot 2022-02-25 at 11 47 39" src="https://user-images.githubusercontent.com/2835259/155710252-b8338110-dfa3-4282-b258-84af754c32ee.png">

jorijnsmit:
we have to do this again, minimum crosschain amount is 0.006 $eth.

jorijnsmit:
success!
https://etherscan.io/tx/0x0512a0d767b23b166f646c8d598db09f1f4584183ce4550e79a6c486fca0e24a
https://ftmscan.com/tx/0xb6a6a316840aa49479c18bc3a58cfb6932719ef52782c1b8492221049182b12e

PR #251, opened 2022-03-15
bridge test amount of eth to fantom

jorijnsmit:
bit difficult to wrap this in general functions that we could use for any token, since anyswap is using a lot of different versions for their router, abis, etc depending on the origin, destination and token itself.

for now this should work from mainnet $eth to fantom, to other chains shouldnt be much different.

guess in the future it would help to also figure out how to bridge badger for example (and do all bridging from msig<>msig in the future).

note: contracts that interact with both the bridge and the router have to be whitelisted; currently only mainnet techops msig is whitelisted.

#21


jalbrekt85 (approved):
very helpful addition of anyswap. functionality of bridging tx looks good. approved so that this can fully tested on mainnet.

petrovska-petro (approved):
Bridging code and addresses added to book looks good!

PR #291, opened 2022-03-23
chore: add bvecvx voter msig to addresses

mitche50:


mitche50:
For reference, this is a new accounting multisig that will house all of the treasury controlled bveCVX and bveCVX LP.  This is necessary to efficiently blacklist the treasury from receiving emissions from voting distributions that the rest of bveCVX votes on so we do not short change users.

Discussion starts here: https://discord.com/channels/785315893960900629/799006521702613023/954420159773294702

Issue #281, opened 2022-03-21
Update emission script according to BIP-87


petrovska-petro:
https://forum.badger.finance/t/bip-87-bvecvx-restructure-voting-strategy-and-emissions-revised-with-community-feedback/5521

```
...
bcvxCRV and Badger emissions from the treasury towards bveCVX would stop, but there would be an option to buy up to 21.25% of bveCVX voting power by replacing the bribes with Badger emissions if deemed cost-efficient.
```

petrovska-petro:
pendant of pushing the json update*

PR #296, opened 2022-03-24
Update csv & json file removing bveCVX/CVX emissions

petrovska-petro:
Tackles #281 

Pushed updates files


jorijnsmit (approved):
set $badger emissions to bbvecvxcvx sett to 0

jalbrekt85 (approved):
bveCVX emissions for `24-03-22` updated to `0`

Issue #277, opened 2022-03-21
month 3: claim and airdrop EPS for month of March


petrovska-petro:
- [x] Claim March month

- [ ] Distribute EPS

Note: will be last distribution. [Announcement](https://twitter.com/Ellipsisfi/status/1504416316945735685)

petrovska-petro:
`TOTAL EPS CLAIMED 41,557.060288641325314529`

https://gnosis-safe.io/app/bnb:0x329543f0F4BB134A3f7a826DC32532398B38a3fA/transactions/0x40b4d0ff903e679b15c7f553127871468a3782879f8db0811ec13a9c4a2681dc

petrovska-petro:
claim: https://bscscan.com/tx/0xe40a601a4b7d62d7c89249eafcf738b31f7f354d33093f7aa4fa9ea6cef06c18

petrovska-petro:
`Threshold is 5.263102493657962 EPSs and current EPS price is $0.190002`

petrovska-petro:
update: `Threshold is 4.298930426109984 EPSs and current EPS price is $0.232616`

petrovska-petro:
distribution:

https://bscscan.com/tx/0xe6585c7aabf5b3a838beb5ffb6c0d261eea005a93188074311095cf2b6619e4b 
https://bscscan.com/tx/0xb3edfe9c00d6b0f56b867868447db5dd7d89818c2ac7df70b72c0b92107f436b

PR #299, opened 2022-03-24
Distribution for March - last EPS airdrop

petrovska-petro:
Tackles #277 second point, final distribution 🥳


jorijnsmit (approved):
getting slight diff locally due to change in price, other than that lgtm!

jalbrekt85 (approved):

Issue #294, opened 2022-03-24
week 12: tree DIGG top-up


petrovska-petro:
Weekly emission: 1.302461219 DIGG

petrovska-petro:
https://etherscan.io/tx/0x81706d7a15201cc1c15e49443b45e5644c69361472cfefa7af9e86aa1f5dc9f2

Issue #293, opened 2022-03-24
week 12: tree BADGER top-up


petrovska-petro:
This week of emissions 23994.02 and 1w of remBADGER emissions 7692.307692

For 24th-31th March top-up is: 31686.327692 BADGER

Deficit for ref: 177229.61869451255

petrovska-petro:
https://etherscan.io/tx/0x81706d7a15201cc1c15e49443b45e5644c69361472cfefa7af9e86aa1f5dc9f2

PR #300, opened 2022-03-24
weekly topup from trops to tree

jorijnsmit:
for #293 and #294


petrovska-petro (approved):
Amounts & destination looks correct, handles: 

- this week of badger & digg emissions 

- sends all balance of bvecvx to `bvecvx_voting_multisig`

sajanrajdev (approved):
- Tree top up amounts match those from the issues. 
- Will transfer all balance of bveCVX on TreasuryOps to the bveCVX voting multi

LGTM

PR #317, opened 2022-03-29
Emissions schedule for BIP-88 bribe payout (Q2-2022)

Tritium-VLK:
For #316

PR #337, opened 2022-04-06
Upload csv, json and change default file in scripts for Q2 emmission - BIP88

petrovska-petro:
Duplicate of #317 (had some issue with unsigned commit)


jorijnsmit (approved):
confirming there is no diff with #317

jalbrekt85 (approved):

PR #334, opened 2022-04-04
first couple of steps in oxd trade

jorijnsmit:
steps and details in #333 

last func still has to be written (the actual oxd swap on fantom)

Issue #333, opened 2022-04-04
sell $dfd and $comp for $usdc, bridge to fantom and buy $oxd


gosuto-inzasheru:
> We got the go ahead from the council to sell our DSD and COMP to USDC and move the funds to Fantom to TWAP OXD(v2). Let me know what support you need. We can check with Tritum to make sure we are swapping from the lowest slippage pool OXD is in
- [x] move dsd to treasury ops
- [x] swap dsd for usdc
- [x] swap comp for usdc
- [x] bridge said usdc to fantom
- [ ] swap for oxd

jorijnsmit:
move dsd to treasury ops: https://etherscan.io/tx/0x55c67a5dbfe1225073ab96dbabaa8bc51aa16b9a4cb76bec25f435a95e013e1f

jorijnsmit:
both swaps to usdc in one tx: https://etherscan.io/tx/0xfa4a834293a484b84b651eb80321bfd40bd14e0f5e74b5e7d30cf5f3f5bf7143

jorijnsmit:
```
Events In This Transaction
--------------------------
├── USDC (0x7EA2be2df7BA6E54B1A9C70676f668455E329d29)
│   ├── Transfer
│   │   ├── from: 0x0000000000000000000000000000000000000000
│   │   ├── to: 0x042B32Ac6b453485e357938bdC38e0340d4b9276
│   │   └── value: 13107469094
│   └── Transfer
│       ├── from: 0x042B32Ac6b453485e357938bdC38e0340d4b9276
│       ├── to: 0x0000000000000000000000000000000000000000
│       └── value: 13107469094
│   
├── IAnyswapV6Router (0x6b7a87899490EcE95443e979cA9485CBE7E71522)
│   └── LogAnySwapOut
│       ├── token: 0x7EA2be2df7BA6E54B1A9C70676f668455E329d29
│       ├── from: 0x042B32Ac6b453485e357938bdC38e0340d4b9276
│       ├── to: 0xf109c50684EFa12d4dfBF501eD4858F25A4300B3
│       ├── amount: 13107469094
│       ├── fromChainID: 1
│       └── toChainID: 250
│   
└── Gnosis Safe (0x042B32Ac6b453485e357938bdC38e0340d4b9276)
    └── ExecutionSuccess
        ├── txHash: 0x3ca227696d7ce7740d8556571c22f11118518dacc709dc0d0c8b63ea5e74ba50
        └── payment: 0
[...]

PR #346, opened 2022-04-06
solidly class addition, swap usdc for oxd

jalbrekt85:
- add `Solidly` class to solve last step  in #333
- changes in `uniV2` to dynamically call swap functions based on the chain (`swapETH`, `swapFTM`, etc). will make using uni style routers on different chains easier in the future
- implemented last function in #334 to perform swap
- usdc/oxd liquidity on solidly seemed to better compared to others, however there might be improvement by routing usdc -> wftm -> oxd through another dex


great_ape_safe/ape_api/uni_v2.py
```diff
@@ -145,7 +146,7 @@ def swap_exact_eth_for_tokens(self, amountIn, path, destination=None):
         # https://docs.uniswap.org/protocol/V2/reference/smart-contracts/router-02#swapethforexacttokens
         destination = self.safe.address if not destination else destination
 
-        tokenOut = path[-1]
+        tokenOut = path[0][-2] if type(path[0][-1]) == bool else path[-1]
```
jorijnsmit, line 149: suggesting to test for `len(path)` instead of `type(path[0][-1])` (or `type(path[0] == tuple`?); bit weird to subscript `path[0]` in the traditional case where it is an actual address.

great_ape_safe/ape_api/uni_v2.py
```diff
@@ -155,7 +156,8 @@ def swap_exact_eth_for_tokens(self, amountIn, path, destination=None):
         address to,
         uint256 deadline
         """
-        self.router.swapExactETHForTokens(
+        signature = getattr(self.router, f'swapExact{self.native_symbol}ForTokens')
```
jorijnsmit, line 159: this is a nice solution. at least they actually changed the func names, i remember seeing pancakeswap just using the func names with ETH even though it really is BNB...

jorijnsmit (changes requested):
left some minor comments, but most important is that the actual swap doesnt result in the expected amount.

here is the whale transfer i add before `.take_snapshot()`:
```
USDC_FTM.transfer(TROPS_FTM, 15_000e6, {'from': accounts.at(
    '0x1a8a4dc716e9379e84e907b0c740d2c622f7cfb7', force=True
)})
```
Screenshot 2022-04-07 at 08 32 41

scripts/issue/333/swap_dfd_comp_for_oxd.py
```diff
@@ -53,17 +59,27 @@ def bridge_usdc_to_ftm(mantissa):
...
+    TROPS_FTM.init_solidly()
+    TROPS_FTM.take_snapshot([USDC_FTM, OXD])
+
+    TROPS_FTM.solidly.swap_tokens_for_tokens(
+        USDC_FTM,
+        USDC_FTM.balanceOf(TROPS_FTM) if swap_all else amount_mantissa,
+        [(USDC_FTM, OXD, True)]
+        )
+
+    TROPS_FTM.print_snapshot()
```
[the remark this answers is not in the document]
jorijnsmit: ok i got it, so the actual path we want to pass here is: `[(USDC_FTM, WFTM, False), (WFTM, OXD, False)]`. the pair that belongs to the direct path from $usdc to $oxd does exist, but isnt deep enough. note you then have to use `tokenOut = path[-1][-2] if type(path[0]) == tuple else path[-1]` in `uni_v2.py`.

however, even better, by using `TROPS_FTM.solidly.router.getAmountOut` we can get the stable bool programmatically; ie it will return the amount and whether the stable or non stable route is the most efficient.

if we want to do it properly, i actually think we should pass the path in the traditional sense (`[USDC_FTM, WFTM, OXD]`, also keeping the signature the same) and let `Solidly.swap_tokens_for_tokens` build the list of tuples to pass to `swapExactTokensForTokens` based on `getAmountOut`.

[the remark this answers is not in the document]
jalbrekt85: you are right, using `getAmountOut` returns the pool type, pushing fixes so that `univ2.swap` is more intelligent with that and the path

great_ape_safe/ape_api/uni_v2.py
jalbrekt85: that can be reverted back to `tokenOut = path[-1]` now that correct path is built inside the function

great_ape_safe/ape_api/uni_v2.py
jalbrekt85: ya some routers use the right native symbol and some dont, this is the best solution I could think of to address this inconsistency without duping a lot of code

jorijnsmit (approved):
looks good to me sir

petrovska-petro (approved):
Gave a try with the sim of sending 15k to trops, v neat the inheritance from UniV2

Address added into book matching with [solidly repo](https://github.com/solidlyexchange/solidly#mainnet) ✅

jorijnsmit:
merging!

Issue #347, opened 2022-04-07
deploy vesting contract for topping up rembadger via a keeper


jorijnsmit:
idea now is to use the open zeppelin contract, much like the one deployed on arbitrum for topping up emissions.

jorijnsmit:
https://etherscan.io/address/0xd87f434fe6d5b349f4376d2daba762b213e403c7#code

PR #348, opened 2022-04-07
rembadger dripper contract deployment

jorijnsmit:
issue #347 

wip

### DIFF
--- a/contracts/RemBadgerDripper.sol
+++ b/contracts/RemBadgerDripper.sol
@@ -26,19 +26,6 @@ contract RemBadgerDripper is VestingWallet {
         _keeper = keeperAddress;
     }
 
-    /**
-     * @dev Getter for the controller address.
-     */
-    function controller() public view virtual returns (address) {
-        return _controller;
-    }
-
-    /**
-     * @dev Getter for the governance address.
-     */
-    function governance() public view virtual returns (address) {
-        return _governance;
-    }
 
     /**
      * @dev Getter for the keeper address.
--- a/contracts/RemBadgerDripper.sol
+++ b/contracts/RemBadgerDripper.sol
@@ -58,11 +58,11 @@ contract RemBadgerDripper is VestingWallet {
         _keeper = _newKeeper;
     }
 
-    function released(address token) override public view virtual returns (uint256) {
+    function released(address token) public view virtual override returns (uint256) {
         return _erc20Released[token];
     }
 
-    function release(address token) override public virtual {
+    function release(address token) public virtual override {
         require(
             _msgSender() == keeper() || _msgSender() == controller() || _msgSender() == governance(),
             "RemBadgerDripper: onlyAuthorized"
--- a/contracts/RemBadgerDripper.sol
+++ b/contracts/RemBadgerDripper.sol
@@ -5,13 +5,71 @@ pragma solidity ^0.8.0;
 import "@openzeppelin/contracts/finance/VestingWallet.sol";
 
 contract RemBadgerDripper is VestingWallet {
+    mapping(address => uint256) private _erc20Released;
+    address private immutable _controller;
+    address private immutable _governance;
+    address private _keeper;
+
     constructor(
         address beneficiaryAddress,
         uint64 startTimestamp,
-        uint64 durationSeconds
+        uint64 durationSeconds,
+        address keeperAddress
     ) VestingWallet(
         beneficiaryAddress,
         startTimestamp,
         durationSeconds
-    ) {}
+    ) {
+        _controller = 0x86cbD0ce0c087b482782c181dA8d191De18C8275; //techops
+        _governance = 0xB65cef03b9B89f99517643226d76e286ee999e77; //dev
+        require(keeperAddress != address(0), "RemBadgerDripper: keeper is zero address");
+        _keeper = keeperAddress;
+    }
+
+    /**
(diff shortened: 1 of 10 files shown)


jorijnsmit:
```
brownie run deployment/deploy_rembadger_dripper
```
or
```
brownie test tests/deployment/test_rembadger_dripper.py
```

contracts/RemBadgerDripper.sol
```diff
@@ -0,0 +1,75 @@
...
+     */
+    function setKeeper(address _newKeeper) public {
+        require(
+            _msgSender() == controller() || _msgSender() == governance(),
+            "RemBadgerDripper: onlyAuthorized"
+        );
+        _keeper = _newKeeper;
+    }
+
+    function released(address token) public view virtual override returns (uint256) {
```
GalloDaSballo, line 61: No need to override this

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/5c4b75a9cdda99af901b8c153d0d1e8a90b0139a/contracts/finance/VestingWallet.sol#L80

GalloDaSballo (commented):
LGTM code wise

Basically a VestinWallet with extra role checks

Would like to see a vesting test with more formal math as you overridden the release function (although you just overrode roles to be fair)

contracts/RemBadgerDripper.sol
```diff
@@ -0,0 +1,75 @@
+// contracts/RemBadgerDripper.sol
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "@openzeppelin/contracts/finance/VestingWallet.sol";
+
+contract RemBadgerDripper is VestingWallet {
+    mapping(address => uint256) private _erc20Released;
+    address private immutable _controller;
+    address private immutable _governance;
```
shuklaayush, line 10:
```suggestion
    address public constant controller = 0x86cbD0ce0c087b482782c181dA8d191De18C8275;
    address public constant governance = 0xB65cef03b9B89f99517643226d76e286ee999e77;
```
You can keep the variables as public constants if they're going to be hardcoded. Constants are inlined during compilation and don't take a storage slot. Will save some `SLOAD` gas when the variables are accessed. The `public` automatically adds a getter.

contracts/RemBadgerDripper.sol
```diff
@@ -0,0 +1,75 @@
...
+    function controller() public view virtual returns (address) {
+        return _controller;
+    }
+
+    /**
+     * @dev Getter for the governance address.
+     */
+    function governance() public view virtual returns (address) {
+        return _governance;
+    }
```
shuklaayush, line 41:
```suggestion

```

shuklaayush (changes requested):
Some nits

contracts/RemBadgerDripper.sol
```diff
@@ -0,0 +1,75 @@
...
+
+    function release(address token) public virtual override {
+        require(
+            _msgSender() == keeper() || _msgSender() == controller() || _msgSender() == governance(),
+            "RemBadgerDripper: onlyAuthorized"
+        );
+        uint256 releasable = vestedAmount(token, uint64(block.timestamp)) - released(token);
+        _erc20Released[token] += releasable;
+        emit ERC20Released(token, releasable);
+        SafeERC20.safeTransfer(IERC20(token), beneficiary(), releasable);
```
shuklaayush, line 73:
```suggestion
        VestingWallet.release(token);
```
This should also work

GalloDaSballo: I agree with Aysuh, basically just add the check + use the super function

contracts/RemBadgerDripper.sol
jorijnsmit: i assume i can then also remove these lines from the constructor?
```
        _controller = 0x86cbD0ce0c087b482782c181dA8d191De18C8275; //techops
        _governance = 0xB65cef03b9B89f99517643226d76e286ee999e77; //dev
```

contracts/RemBadgerDripper.sol
jorijnsmit: thanks, this is what i was looking for, also solves having to overwrite/redefine `_erc20Released`!

contracts/RemBadgerDripper.sol
shuklaayush: Yep. Would also have to change references to `controller()`/`governance()` to `controller/governance`

jorijnsmit:
deployed to rinkeby: https://rinkeby.etherscan.io/tx/0xed8b5157cdc49d603c58c761b9f36115a9e7c6c4633c3eed13526f8949ee7830

would like to do run some tests onchain:
- [ ] send funds
- [ ] call release before start
- [ ] setKeeper from gov
- [ ] setKeeper from random
- [ ] call release after start

GalloDaSballo:
Code LG it's basicaly a VestedWallet with extra permission check

I'd like @dapp-whisperer to give review as well as this contract is not ugradeable

shuklaayush (approved):
LGTM. Can add upgradeability if desired

shuklaayush (approved):
LGTM

GalloDaSballo:
I'd say after changes the code LGTM

OZ VestingWallet with extra sweep functionality

Saves from needing upgradeability

dapp-whisperer (approved):

PR #350, opened 2022-04-08
hotfix: `ftoken_set_admin_fee` requires ftoken address

jorijnsmit:


jalbrekt85 (approved):
dont know how that slipped through, looks good now

petrovska-petro (approved):
good catch!

PR #351, opened 2022-04-11
spookyswap class, better routing for swapping to oxd

jalbrekt85:
- adds `SpookySwap` class to api.
- `swap_usdc_for_oxd` in #346 now first swaps usdc -> wftm through spooky then swaps wtfm -> oxd through solidly. results in **~22% more oxd**
- revised the ever expanding `GreatApeSafe.init_all`

### DIFF
--- a/great_ape_safe/great_ape_safe.py
+++ b/great_ape_safe/great_ape_safe.py
@@ -7,7 +7,7 @@ from io import StringIO
 
 import pandas as pd
 from ape_safe import ApeSafe
-from brownie import Contract, network, ETH_ADDRESS
+from brownie import Contract, network, ETH_ADDRESS, exceptions
 from rich.console import Console
 from rich.pretty import pprint
 from tqdm import tqdm
@@ -25,6 +25,7 @@ from great_ape_safe.ape_api.opolis import Opolis
 from great_ape_safe.ape_api.pancakeswap_v2 import PancakeswapV2
 from great_ape_safe.ape_api.rari import Rari
 from great_ape_safe.ape_api.solidly import Solidly
+from great_ape_safe.ape_api.spookyswap import SpookySwap
 from great_ape_safe.ape_api.sushi import Sushi
 from great_ape_safe.ape_api.uni_v2 import UniV2
 from great_ape_safe.ape_api.uni_v3 import UniV3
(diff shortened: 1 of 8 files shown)


jorijnsmit:
linking #333

scripts/issue/333/swap_dfd_comp_for_oxd.py
```diff
@@ -74,12 +74,19 @@ def bridge_usdc_to_ftm(mantissa):
...
         USDC_FTM,
-        USDC_FTM.balanceOf(TROPS_FTM) if not amount_mantissa else amount_mantissa,
-        [USDC_FTM, WFTM, OXD]
+        USDC_FTM.balanceOf(TROPS_FTM) if amount_mantissa is None else amount_mantissa,
+        [USDC_FTM, WFTM]
+        )
+
+    TROPS_FTM.solidly.swap_tokens_for_tokens(
+        WFTM,
+        WFTM.balanceOf(TROPS_FTM),
```
jorijnsmit, line 88: this will need to be dusty as it is market dependent and will change between posting and execing.

great_ape_safe/great_ape_safe.py
```diff
@@ -51,21 +52,13 @@ def __init__(self, address, base_url=None, multisend=None):
...
-        self.init_sushi()
-        self.init_uni_v2()
-        self.init_uni_v3()
+        for method in self.__dir__():
+            if method.startswith('init_') and method != 'init_all':
+                try:
+                    getattr(self, method)()
+                except exceptions.ContractNotFound:
+                    # different chain
+                    pass
```
jorijnsmit, line 61: very nice!

there is actually another optimisation i have been thinking about for a while: all the constants in the protocol classes such as `self.max_slippage` and `self.deadline` should actually be declared outside the `__init__` block. this will make them inheritable, saving having to add them manually to every time ;)

maybe you can push that through in this pr as well, or write a new ticket for it?

jorijnsmit (changes requested):

scripts/issue/333/swap_dfd_comp_for_oxd.py
jalbrekt85: good point

great_ape_safe/great_ape_safe.py
jalbrekt85: good idea, will add in this pr

jorijnsmit (approved):
22% increase in resulting tokens is def worth it :)

Tritium-VLK (approved):

Issue #358, opened 2022-04-12
Set up pause guardian in pool-22


petrovska-petro:
Roles:

- [ ] pauseGuardian

The role of pause guardian will be given to the [guardian](https://github.com/Badger-Finance/badger-multisig/blob/main/helpers/addresses.py#L15)

petrovska-petro:
https://docs.google.com/document/d/1LVoeSdT5lJkFBrXMMh-jgXS4vF8pZAO7X-If_uYofE8/edit

jorijnsmit:
what i understand from your doc is:
- dev msig will have to call `RariComptroller._setPauseGuardian(address newPauseGuardian)` on our pool
- with that in place, support for calling `RariComptroller._setBorrowPaused(CToken cToken, bool state)` will then have to be added to our warroom proxy.

i suggest to leave it at this conclusion for now, and not do any more work on this until rari things have cleared up.

PR #359, opened 2022-04-12
Set guardian in rari pool-22

petrovska-petro:
Tackles #358 

Still draft, pendant of defining who will be the guardian as scope of `pauseGuardian` is wide


jorijnsmit (approved):
sure lets merge in case we need this in the future

PR #371, opened 2022-04-13
Include quarterly emissions scripts for future usage + fix minor issue on target_file

petrovska-petro:
Since emissions had get constant, pushing this scripts to post whole quarters if possible

```
brownie run --network {arbitrum-main-fork || mainnet-fork} emissions/quarterly_emissions
```


jorijnsmit:
looks like these two scripts have a lot of overlap, is it maybe an idea to have one `quarterly_emissions.py` script and add a separate function for arb1?

feels a bit weird having all these helper functions duplicated, especially making maintenance a lot harder.

now that i think of it, even the weekly scripts are also duplicated, so for example we will end up with four the same `find_thursday` funcs..?

petrovska-petro:
> looks like these two scripts have a lot of overlap, is it maybe an idea to have one `quarterly_emissions.py` script and add a separate function for arb1?
> 
> feels a bit weird having all these helper functions duplicated, especially making maintenance a lot harder.
> 
> now that i think of it, even the weekly scripts are also duplicated, so for example we will end up with four the same `find_thursday` funcs..?

good suggestion, will convert this pr into draft till i push the unification

scripts/emissions/quarterly_emissions.py
```diff
@@ -0,0 +1,134 @@
...
+
+        for sett in emissions_data["setts"]:
+            if sett["network"] == "ETHEREUM":
+                beneficiary = sett["address"]
+
+                if sett["badger_allocation"] != 0:
+                    #  make difference between native setts as per WeeklyEmissions.md
+                    if beneficiary in AUTOCOMPOUND_100_SETTS:
+                        continue  # nothing to emit
+                    else:
```
jorijnsmit, line 60: strictly taken not needed after the `continue`

scripts/emissions/quarterly_emissions.py
```diff
@@ -0,0 +1,134 @@
...
+
+    emissions_data = data[target_week.strftime("%d-%m-%y")]
+    time_range = emissions_data["timerange"]
+
+    last_quarter_time_rage = list(data.keys())[-1]
+    end_quarter_timestamp = data[last_quarter_time_rage]["timerange"]["endtime"]
+
+    duration = end_quarter_timestamp - time_range["starttime"]
+
+    weeks_no = duration / WEEK_DURATION
```
jorijnsmit, line 35:
```suggestion
    weeks_no = duration / WEEK_DURATION
    assert weeks_no == duration // WEEK_DURATION
```
i think we need a round number here, which it looks like works out in this case, but this could be enforced with `weeks_no = duration // WEEK_DURATION`.

or maybe even better would be adding an extra `assert weeks_no == duration // WEEK_DURATION` to catch possible mistakes in the `duration`.

jorijnsmit (commented):
tiny comments but good to go imo.

scripts/emissions/quarterly_emissions.py
petrovska-petro: sounds cool enforcing more defensive coding here 👍

jorijnsmit (approved):

jalbrekt85 (approved):

PR #373, opened 2022-04-14
sign and exec tx without gnosis api

gosuto-inzasheru:
instead of posting the tx to the gnosis api, this helper func can post an execTransaction directly to the gnosis safe smart contract. calling the helper func multiple times, passing it a partial signature every time, will eventually build the final signature when enough signatures have been reached. calldata can then be posted manually to the safe.

no gnosis api or ui needed in any of these steps, this is from local straight to onchain.


jorijnsmit:
https://rinkeby.etherscan.io/tx/0xda040e1de76d88dbe57ab89725f087281a212213550f7cb0fb251dd8750b12d6

petrovska-petro (approved):
Beautiful, tested the flow on rinkeby together. removing dependency from endpoint & ui 💪 

Good leverage of `bytes.fromhex(...)` 👌

great_ape_safe/great_ape_safe.py
```diff
@@ -252,6 +255,71 @@ def contract(self, address, Interface=None, from_explorer=False):
         else:
             return Contract(address, owner=self.account)
 
+    def post_safe_tx_manually(self):
+        safe_tx = self.post_safe_tx(events=False, silent=True, post=False)
```
jalbrekt85, line 259: in the case that there is already a partial signature, wouldn't it better to get the `safe_tx` from `_get_safe_tx_by_nonce`?. what happens if signers run the same script but produce different calldata?

gosuto-inzasheru: that func relies on the gnosis api which the tx is never posted to---the whole idea here is to avoid that web2 api

your point is valid though; if the locally run tx differs slightly (ie diff min out calc based on market or something) the calldata will be diff.

maybe final calldata should be passed from signers as well, just like the partial sig

gosuto-inzasheru:
suggesting we review and merge this. we have proven it works multiple times.

just cant get the checkSignatures to work locally, so had to revert to a naive way to check completeness of signature instead (`len(sig) == n * 65`)

jalbrekt85 (approved):

petrovska-petro (approved):
confirmed logic and test run in mainnet: https://etherscan.io/tx/0x758c46f0035c1473872f87fc9053286a917c08d182f5ca886910bdcb6aaa1d59 ✅

petrovska-petro (approved):

jalbrekt85 (approved):

Issue #375, opened 2022-04-14
reorg stable farms for next quarter


jorijnsmit:
the treasury pod estimated that fin ops needs $1.2m for next quarter (2022-q2). treasury pod is also working on reorganising some of the stable farms. first steps that can be made here are as follows:
- [x] withdraw ausdc position to usdc from aave
- [ ] ~~withdraw cvx3pool to usdc from convex~~
- [x] send a total of $1.2m to fin ops

refarming the remainder will follow later in a separate issue, once it is known into what farms.

jorijnsmit:
```
snapshot result for 0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e:
                      balance_before         balance_after                  balance_delta
symbol                                                                                   
aUSDC   1,310,154.257035000000000000  0.000000000000000000  -1,310,154.257035000000000000 

snapshot result for 0x30a9c1D258F6c2D23005e6450E72bDD42C541105:
                  balance_before                 balance_after                 balance_delta
symbol                                                                                      
USDC    4,116.708985000000000000  1,314,270.988386000000000000  1,310,154.279401000000000000 
```
```
Events In This Transaction
--------------------------
├── InitializableImmutableAdminUpgradeabilityProxy (0xd784927Ff2f95ba542BfC824c8a8a98F3495f6b5)
│   ├── AssetIndexUpdated
│   │   ├── asset: 0xBcca60bB61934080951369a648Fb03DF4F96263C
│   │   └── index: 33246369940192167987150214
│   ├── UserIndexUpdated
│   │   ├── user: 0x464C71f6c2F760DdA6093dCB91C24c39e5d6e18c
│   │   ├── asset: 0xBcca60bB61934080951369a648Fb03DF4F96263C
│   │   └── index: 33246369940192167987150214
│   └── RewardsAccrued
│       ├── user: 0x464C71f6c2F760DdA6093dCB91C24c39e5d6e18c
│       └── amount: 17520889104361858
│   
├── Aave interest bearing USDC (0xBcca60bB61934080951369a648Fb03DF4F96263C)
│   └── (unknown)
│       ├── topic1: 0x4c209b5fc8ad50758f13e2e1088ba56a560dff690a1c6fef26394f4c03821c4f
│       ├── topic2: 0x000000000000000000000000464c71f6c2f760dda6093dcb91c24c39e5d6e18c
[...]

PR #376, opened 2022-04-14
withdraw ausdc from aave to payments msig

jorijnsmit:
this to cover q2 budget, see #375


jalbrekt85 (approved):
having to use `skip_preview=True` to get to posting part but rest looks good

Issue #352, opened 2022-04-11
swap last bit of $eps on bsc msig to $bnb


jorijnsmit:
we wont be airdropping any more $eps, converting it to $bnb makes it available for gas.

sajanrajdev:
Rejected Nonce 47 due to expired deadline. Resubmitted at nonce 49: 
```
{
│   'ethereum_client': <gnosis.eth.ethereum_client.EthereumClient object at 0x7f16be544f70>,
│   'safe_address': '0x329543f0F4BB134A3f7a826DC32532398B38a3fA',
│   'to': '0x40A2aCCbd92BCA938b02010E17A5b8929b49130D',
│   'value': 0,
│   'data': HexBytes('0x8d80ff0a000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000001f200a7f552078dcc247c2684336020c03648500c6d9f00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b300000000000000000000000010ed43c718714eb63d5aa57b78b54704e256024e00000000000000000000000000000000000000000000004b5b8e2cfbc327d17f0010ed43c718714eb63d5aa57b78b54704e256024e0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000010418cbafe500000000000000000000000000000000000000000000004b5b8e2cfbc327d17f00000000000000000000000000000000000000000000000008a44b81710ce55000000000000000000000000000000000000000000000000000000000000000a0000000000000000000000000329543f0f4bb134a3f7a826dc32532398b38a3fa00000000000000000000000000000000000000000000000000000000626c9aa40000000000000000000000000000000000000000000000000000000000000002000000000000000000000000a7f552078dcc247c2684336020c03648500c6d9f000000000000000000000000bb4cdb9cbd36b01bd1cbaebf2de08d9173bc095c0000000000000000000000000000'),
│   'operation': 1,
│   'safe_tx_gas': 0,
│   'base_gas': 0,
│   'gas_price': 0,
│   'gas_token': '0x0000000000000000000000000000000000000000',
│   'refund_receiver': '0x0000000000000000000000000000000000000000',
│   'signatures': b'',
│   '_safe_nonce': 49,
│   '_safe_version': '1.3.0',
│   '_chain_id': None
}
[...]

PR #381, opened 2022-04-19
swap eps for bnb script

jalbrekt85:
handles #352


helpers/addresses.py
```diff
@@ -475,6 +475,7 @@
         "BTCb": "0x7130d2A12B9BCbFAe4f2634d864A1Ee1Ce3Ead9c",
         "WBNB": "0xbb4cdb9cbd36b01bd1cbaebf2de08d9173bc095c",
         "CAKE": "0x0E09FaBB73Bd3Ade0a17ECC321fD13a19e81cE82",
+        "EPS": "0xA7f552078dcC247C2684336020c03648500C6d9F",
```
petrovska-petro, line 478: for processing this token, already it is on the addr book. [ref](https://github.com/Badger-Finance/badger-multisig/blob/main/helpers/addresses.py#L510). i think after converting this to gas we likely will not touch eps again

jorijnsmit (approved):
looking good

i will make sure you are added as a delegate so you can post it too :)

petrovska-petro (approved):

Issue #380, opened 2022-04-19
increase gas refill threshold for ops_botsquad on fantom


jorijnsmit:

jorijnsmit:
we need to figure out how much gas botsquad rougly needs on a monthly basis and update that amount in the fantom gas refill script accordingly.

botsquad: https://ftmscan.com/address/0xF8dbb94608E72A3C4cEeAB4ad495ac51210a341e

jorijnsmit:
last gas refill, 13 days ago: https://ftmscan.com/tx/0xfe89a490535cbaaa13d9571f7e5b6812367b6ade2098ca840fe84a699106fcbe

jalbrekt85:
> we need to figure out how much gas botsquad rougly needs on a monthly basis and update that amount in the fantom gas refill script accordingly.
> 
> botsquad: https://ftmscan.com/address/0xF8dbb94608E72A3C4cEeAB4ad495ac51210a341e

botsquad uses an avg ~330 ftm/month. avg gas price on ftm is[ much lower now than previous month](https://ftmscan.com/chart/gasprice) however. 350 ftm for monthly refill seems to be a safe amount that can account for gas spikes

jorijnsmit:
i would suggest to make that margin a but broader, not much to lose here. like 500 maybe.

PR #394, opened 2022-04-21
increase ops_botsquad refill amount to 500 ftm

jalbrekt85:
handles #380


jorijnsmit (approved):
nice, lets see if this will be enough :)

petrovska-petro (approved):
bump up gas refill ✅

Issue #412, opened 2022-04-28
add balancer to great_ape_safe


jorijnsmit:
we need our standard deposit and withdraw functions, not sure of the technicalities but preferably something as close to the convex functions as possible.

most important ones for now:
- `deposit_and_stake(self, underlying, mantissa)`
- `claim_all(self)`
- `unstake_all_and_withdraw_all(self, underlying, claim=1)`

petrovska-petro:
for dev ref: https://dev.balancer.fi/resources/joins-and-exits

PR #417, opened 2022-05-02
balancer addition to great ape safe

jalbrekt85:
first draft for #412 
- balancer has no on-chain registry, pools are cached from subgraph
- no on-chain functions for calcs such as `getAmountsOut`, relevant functions ported from their [sdk](https://github.com/balancer-labs/balpy)

- [x] `deposit_and_stake`
- [x] `unstake_and_withdraw`
- [x] `claim_all`
- [x] support single asset deposit/withdrawal
- [ ] support stable pools
- [ ] support native eth


tests/conftest.py
```diff
@@ -190,3 +190,45 @@ def cvxCRV(safe):
...
+
+
+@pytest.fixture
+def bpt(dev):
+    return dev.contract(registry_addr.eth.balancer['50-BTC-50-WETH'])
+
+
+@pytest.fixture
+def staked_bpt(dev):
+    return dev.contract(registry_addr.eth.balancer['50-BTC-50-WETH-G'])
```
jorijnsmit, line 209: i think it is good to start with moving protocol specific fixtures to `tests/<protocol>/conftest.py`.

i saw this being done here: https://github.com/Citadel-DAO/citadel-multisig/pull/6/files

tests/balancer/test_balancer_withdraw.py
```diff
@@ -0,0 +1,31 @@
+from brownie import chain
+import pytest
+
+@pytest.fixture(scope="function", autouse=True)
```
jorijnsmit, line 4: nit: i think `scope='function'` is the default

helpers/addresses.py
```diff
@@ -430,6 +430,12 @@
     "arbitrum": {
         "outbox": "0x760723CD2e632826c38Fef8CD438A4CC7E7E1A40",
     },
+    "balancer": {
+        "vault": "0xBA12222222228d8Ba445958a75a0704d566BF2C8",
+        "gauge_factory": "0x4E7bBd911cf1EFa442BC1b2e9Ea01ffE785412EC",
+        "50-BTC-50-WETH": "0xA6F548DF93de924d73be7D25dC02554c6bD66dB5",
+        "50-BTC-50-WETH-G": "0x4E3c048BE671852277Ad6ce29Fd5207aA12fabff"
+    }
```
jorijnsmit, line 438: confirmed through https://dev.balancer.fi/resources/vebal-and-gauges/gauges

also might be easier to grab the gauge from the gauge factory programmatically, instead of hardcoding @jalbrekt85?

last minor comment is that `DotMap` doesnt like dashes, so i would suggest to use underscores instead. this prevents having to fall back to traditional dict subscripting, eg `registry.a.b['c-d']`.

tests/conftest.py
jalbrekt85: good idea

great_ape_safe/ape_api/balancer.py
```diff
@@ -152,47 +210,140 @@ def unstake_all_and_withdraw_all(
...
+
+
+    def unstake_and_withdraw_all_single_asset(
+        self, asset, underlyings=None, pool=None, unstake=True, claim=True, is_eth=False, destination=None
+    ):
+        if not underlyings and not pool:
+            raise TypeError('must provide either underlyings or pool')
+
+        if underlyings:
+            underlyings = sorted(underlyings)
```
jorijnsmit, line 266:
```suggestion
            underlyings = sorted([x.address for x in underlyings])
```
also in L206 but i cant comment there for some reason

great_ape_safe/ape_api/balancer.py
```diff
@@ -152,47 +210,140 @@ def unstake_all_and_withdraw_all(
...
 
-        balances_after = [Contract(x).balanceOf(destination) for x in underlyings]
-        assert all([x > y for x, y in zip(balances_after, balances_before)])
+        balances_after = [Contract(x).balanceOf(destination) for x in request[0]]
+        assert any([x > y for x, y in zip(balances_after, balances_before)])
+
+
+    def claim_all(self, pool):
+        # claim reward token from pool's gauge
+        gauge = self.safe.contract(self.gauge_factory.getPoolGauge(pool))
```
jorijnsmit, line 336:
```suggestion
    def claim_all(self, underlyings=None, pool=None):
        # claim reward token from pool's gauge
        if underlyings:
            underlyings = sorted([x.address for x in underlyings])
            pool_id = self.find_pool_for_underlyings(underlyings)
            pool = self.safe.contract(self.vault.getPool(pool_id)[0])        
        gauge = self.safe.contract(self.gauge_factory.getPoolGauge(pool))
```
add a lazy `underlyings` in case pool addr is not at hand

great_ape_safe/ape_api/helpers/balancer/stable_math.py
```diff
@@ -0,0 +1,337 @@
+# https://github.com/balancer-labs/balpy/blob/a907f7b984f4e3ba3460a1ef064003d95da5e884/balpy/balancerv2cad/src/balancerv2cad/StableMath.py
```
jorijnsmit, line 1: is there a reason you removed the `getcontext().prec = 28` lines? i believe this is setting the precision in the background for the decimal module, which is definitely being used no?

great_ape_safe/ape_api/helpers/balancer/weighted_math.py
```diff
@@ -1,53 +1,244 @@
...
+        return result
+
+    @staticmethod
+    def calc_bpt_out_given_exact_tokens_in(pool, balances: list, amounts_in: list):
+
+        balances = [Decimal(x) for x in balances]
+        amounts_in = [Decimal(x) for x in amounts_in]
+        normalized_weights = [Decimal(x / 1e18) for x in pool.getNormalizedWeights()]
+        bptTotalSupply = Decimal(pool.totalSupply())
+        swap_fee = Decimal(pool.getSwapFeePercentage() / 1e18)
```
jorijnsmit, line 79: im wondering if it wouldnt be better to do these type of conversions using a helper func in the balancer class? this way weighted_math.py can remain untouched and used completely raw from source. dunno.

jorijnsmit (commented):
awesome job on the peripheries, balancer is even worse in that regards than univ3!

great_ape_safe/ape_api/helpers/balancer/stable_math.py
jalbrekt85: including `getcontext().prec = 28` gives this error: 

`vyper.exceptions.DecimalOverrideException: Overriding decimal precision disabled`

 will have to look more into how much of a difference this is actually making

helpers/addresses.py
```diff
@@ -440,6 +440,10 @@
         "gauge_factory": "0x4E7bBd911cf1EFa442BC1b2e9Ea01ffE785412EC",
         "B_50_BTC_50_WETH": "0xA6F548DF93de924d73be7D25dC02554c6bD66dB5"
     }
+    "hidden_hand": {
```
jorijnsmit, line 443: comma missing in L442

jorijnsmit (commented):
not all tests are passing

if we want to merge i think it is good to mark the non functioning methods and tests as such so that we know it still needs work

```
==================================================== short test summary info =====================================================
FAILED tests/balancer/test_balancer_deposit.py::test_deposit_and_stake_single_asset - assert 20631664 == 0
FAILED tests/balancer/test_balancer_withdraw.py::test_unstake_and_withdraw_all_stable - OverflowError: exitPool '(('0x6B175474E...
ERROR tests/balancer/test_balancer_claim.py::test_claim_all - brownie.exceptions.VirtualMachineError: revert: BAL#506
============================ 2 failed, 5 passed, 2 xfailed, 6 warnings, 1 error in 118.11s (0:01:58) =============================
```

will have a closer look tomo morning; in ~9h.

jorijnsmit (dismissed):
pretty crazy protocol, lots to do still and lots implemented already. gets through tests without errors/fails.

think merging now, to at least be able to use the basic deposit and withdraw methods, makes a lot of sense. otherwise operations will fall behind.

tests/balancer/test_balancer_deposit.py
```diff
@@ -0,0 +1,104 @@
...
+    bal_before_staked_bpt = weighted_staked_bpt.balanceOf(dev)
+    bal_before_dai = dai.balanceOf(dev)
+
+    balancer.deposit_and_stake_single_asset(
+        dai,
+        dev.account.balance(),
+        weighted_bpt
+    )
+
+    assert dev.account.balance() < bal_before_dai
```
petrovska-petro, line 83:
```suggestion
    assert dai.balanceOf(dev) < bal_before_dai
```

great_ape_safe/ape_api/balancer.py
```diff
@@ -0,0 +1,366 @@
...
+            underlyings[weth_index] = ZERO_ADDRESS
+
+        request = (
+            underlyings,
+            [min_out if x == asset.address else 0 for x in underlyings],
+            data_encoded,
+            False
+        )
+
+        pool.approve(self.vault, amount_in)
```
petrovska-petro, line 317: i think this approve is not needed, neither in L258

helpers/addresses.py
```diff
@@ -434,6 +435,11 @@
     "arbitrum": {
         "outbox": "0x760723CD2e632826c38Fef8CD438A4CC7E7E1A40",
     },
+    "balancer": {
```
petrovska-petro, line 438: addrs verified against [deployments source](https://dev.balancer.fi/references/contracts/deployment-addresses) ✅

great_ape_safe/ape_api/balancer.py
```diff
@@ -0,0 +1,366 @@
...
+            assert pool.balanceOf(self.safe) >= balance_before_gauge
+
+        balances_before = [Contract(x).balanceOf(destination) for x in request[0]]
+
+        self.vault.exitPool(
+            pool_id, self.safe, destination, request
+        )
+
+        balances_after = [Contract(x).balanceOf(destination) for x in request[0]]
+        assert any([x > y for x, y in zip(balances_after, balances_before)])
```
petrovska-petro, line 343: should not here check *all* for the `unstake_all_and_withdraw_all` and *any* for `unstake_and_withdraw_all_single_asset`?

great_ape_safe/ape_api/balancer.py
```diff
@@ -0,0 +1,366 @@
...
+
+    def claim_all(self, underlyings=None, pool=None):
+        # claim reward token from pool's gauge
+        if underlyings:
+            underlyings = sorted([x.address for x in underlyings])
+            pool_id = self.find_pool_for_underlyings(underlyings)
+            pool = self.safe.contract(self.vault.getPool(pool_id)[0])
+
+        gauge = self.safe.contract(self.gauge_factory.getPoolGauge(pool))
+        reward_address = gauge.reward_tokens(0)
```
petrovska-petro, line 354: minor comment here, not sure how often that may happen that a gauge has multiple rewards tokens for the grabs, but perhaps worthy to leave it more dynamic this method based on the `gauge.reward_count()`, instead of hardcoding the checkup on only one token

great_ape_safe/ape_api/balancer.py
jalbrekt85: assertion will always fail on single asset withdrawal using `all`, so `any` is used to be agnostic towards both types of withdrawals

great_ape_safe/ape_api/balancer.py
jalbrekt85: good idea

great_ape_safe/ape_api/balancer.py
petrovska-petro: nice workaround sir!

jorijnsmit (approved):
L G T M

Issue #404, opened 2022-04-26
deploy dripper for the tree


jorijnsmit:

jorijnsmit:
#421

jorijnsmit:
deployment: https://etherscan.io/tx/0xa0aa5a906ae953846c14db98507fc03644e7e8f9ae9347ff34f7b7dab69ad04c

still needs to be registered with chainlink, techops will needs some $link in order to do so

jorijnsmit:
deployer swapped some (W)ETH for LINK, sent directly to techops: https://etherscan.io/tx/0x38e683e21cf6122a44813d70cce78bbe9eff7e99931387dad3715f3201a3ead5

jorijnsmit:
```
Events In This Transaction
--------------------------
├── ChainLink Token (0x514910771AF9Ca656af840dff83E8264EcF986CA)
│   ├── Transfer
│   │   ├── from: 0x86cbD0ce0c087b482782c181dA8d191De18C8275
│   │   ├── to: 0x09B1de9Ecf6600FD44e1cB1b46DF6B520f33ec6d
│   │   └── value: 125000000000000000000
│   └── Transfer
│       ├── from: 0x86cbD0ce0c087b482782c181dA8d191De18C8275
│       ├── to: 0x09B1de9Ecf6600FD44e1cB1b46DF6B520f33ec6d
│       ├── value: 125000000000000000000
│       └── data: 0xc4110e5c00000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000000140000000000000000000000000fa55d407f48849ae98a6e4f11dc5c18e035f46ec0000000000000000000000000000000000000000000000000000000000061a8000000000000000000000000086cbd0ce0c087b482782c181da8d191de18c82750000000000000000000000000000000000000000000000000000000000000180000000000000000000000000000000000000000000000006c6b935b8bbd400000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001154726565447269707065723230323251320000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
│   
├── IUpkeepRegistrationRequests (0x09B1de9Ecf6600FD44e1cB1b46DF6B520f33ec6d)
│   └── RegistrationRequested
[...]

PR #421, opened 2022-05-04
feat: add AutonomousDripper contract, deploy and tests

jorijnsmit:
some changes in regards to previous `EmissionsDripper`:
- no more `keeper` role; the upkeep function needs to be permissionless in order for chainlink keepers to interact with it
- because of this i also removed the `controller` and `governance` roles and replaced them with one single `owner`
- `sweep` remains unchanged (callable by `owner` only)

still need to implement a setter function for the `assetsWatchlist`, which is now only set on deployment

### DIFF
--- a/contracts/AutonomousDripper.sol
+++ b/contracts/AutonomousDripper.sol
@@ -1,0 +1,137 @@
+// contracts/EmissionsDripper.sol
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import "@openzeppelin/contracts/finance/VestingWallet.sol";
+import "@chainlink/contracts/src/v0.8/KeeperCompatible.sol";
+
+/**
+ * @title The AutonomousDripper Contract
+ * @author gosuto.eth
+ * @notice A Chainlink Keeper-compatible version of OpenZeppelin's
+ * VestingWallet; removing the need to monitor the interval and/or call release
+ * manually. Also adds an owner that can sweep all ether and ERC-20 tokens.
+ */
+contract AutonomousDripper is VestingWallet, KeeperCompatibleInterface {
+    event EtherSwept(uint256 amount);
+    event ERC20Swept(address indexed token, uint256 amount);
+
+    address private _owner;
+    uint public lastTimestamp;
+    uint public immutable interval;
+    // TODO: write setter methods that can add or remove entries from this list
+    address[] public assetsWatchlist;
+
+    constructor(
+        address beneficiaryAddress,
+        uint64 startTimestamp,
+        uint64 durationSeconds,
+        address ownerAddress,
+        uint intervalSeconds,
+        address[] memory watchlistAddresses
+    ) VestingWallet(
+        beneficiaryAddress,
+        startTimestamp,
+        durationSeconds
+    ) {
+        _owner = ownerAddress;
+        lastTimestamp = startTimestamp;
+        interval = intervalSeconds;
+        assetsWatchlist = watchlistAddresses;
+    }
+
+    /**
+     * @dev Loop over the assetsWatchlist and check their local balance.
+     * Returns a filtered version of the assetsWatchlist for which the local
+     * balance is greater than zero.
+     * Strongly inspired by https://github.com/smartcontractkit/chainlink/blob/0adf6c0fb256d09e4fed2f2f86a2c027ae82535d/contracts/src/v0.8/upkeeps/EthBalanceMonitor.sol#L87-L116
+     */
+    function _getAssetsHeld() internal view returns (address[] memory) {
+        address[] memory _assetsHeld = new address[](assetsWatchlist.length);
+        uint256 count = 0;
+        for (uint idx = 0; idx < assetsWatchlist.length; idx++) {
+            uint256 balance = IERC20(assetsWatchlist[idx]).balanceOf(address(this));
+            if (balance > 0) {
+                _assetsHeld[count] = assetsWatchlist[idx];
+                count++;
+            }
+        }
+        if (count != assetsWatchlist.length) {
+            assembly {
+                mstore(_assetsHeld, count)
+            }
+        }
+        return _assetsHeld;
+    }
+
+    /**
+     * @dev Runs off-chain at every block to determine if the `performUpkeep`
+     * function should be called on-chain.
+     */
+    function checkUpkeep(bytes calldata) external view override returns (
+        bool upkeepNeeded, bytes memory
+    ) {
+        address[] memory assetsHeld = new address[](assetsWatchlist.length);
+        uint256 count = 0;
+
+        if ((block.timestamp - lastTimestamp) > interval) {
+            assetsHeld = _getAssetsHeld();
+            if (assetsHeld.length > 0) {
+                return (true, abi.encode(assetsHeld));
+            }
+        }
+    }
+
+    /**
+     * @dev Contains the logic that should be executed on-chain when
+     * `checkUpkeep` returns true.
+     */
+    function performUpkeep(bytes calldata performData) external override {
+        if ((block.timestamp - lastTimestamp) > interval) {
+            lastTimestamp = block.timestamp;
+            address[] memory assetsHeld = abi.decode(performData, (address[]));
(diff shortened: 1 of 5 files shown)


jorijnsmit:
hourly dripper live here on rinkeby: https://rinkeby.etherscan.io/address/0xfA55d407F48849aE98a6e4f11Dc5C18E035F46Ec#tokentxns
keeper dashboard: https://keepers.chain.link/rinkeby/302

contracts/AutonomousDripper.sol
```diff
@@ -0,0 +1,145 @@
...
+    /**
+     * @dev Contains the logic that should be executed on-chain when
+     * `checkUpkeep` returns true.
+     */
+    function performUpkeep(bytes calldata performData) external override {
+        if ((block.timestamp - lastTimestamp) > interval) {
+            lastTimestamp = block.timestamp;
+            address[] memory assetsHeld = abi.decode(performData, (address[]));
+            for (uint idx = 0; idx < assetsHeld.length; idx++) {
+                if (IERC20(assetsHeld[idx]).balanceOf(address(this)) > 0) {
```
petrovska-petro, line 102: thinking that this will be called immediately after `checkUpkeep` splits out `True` and already there has being check if the balance of the token address was greater than zero. thinking that above this L in two ways:

- seems redundant check after the reads on `checkUpkeep`

- will add a bit of extra cost of reading that again on the chainlink keeper side that we should pay for

what are your thoughts?

jorijnsmit: https://docs.chain.link/docs/chainlink-keepers/compatible-contracts/#revalidate-performupkeep

petrovska-petro: gotcha!  🧠

shuklaayush:
One issue that I see is that someone can call `performUpkeep` with an empty array to grief the system

jorijnsmit:
> One issue that I see is that someone can call `performUpkeep` with an empty array to grief the system

mind having another look @shuklaayush? the timestamp is now only updated if an actual release is performed (0c701c1).

also for additional security i introduced the modifiers `onlyOwner`, `onlyKeeperRegistry` and `whenNotPaused`.

contracts/AutonomousDripper.sol
```diff
@@ -0,0 +1,187 @@
...
+        uint256 count = 0;
+        for (uint idx = 0; idx < assetsWatchlist.length; idx++) {
+            uint256 balance = IERC20(assetsWatchlist[idx]).balanceOf(address(this));
+            if (balance > 0) {
+                _assetsHeld[count] = assetsWatchlist[idx];
+                count++;
+            }
+        }
+        if (count != assetsWatchlist.length) {
+            assembly {
```
GalloDaSballo, line 75: What's the deal with random uncommented assembly?

contracts/AutonomousDripper.sol
```diff
@@ -0,0 +1,187 @@
...
+    */
+    function getKeeperRegistryAddress() external view returns (address keeperRegistryAddress) {
+        return _keeperRegistryAddress;
+    }
+
+    /**
+    * @dev Setter for the keeper registry address.
+    */
+    function setKeeperRegistryAddress(address keeperRegistryAddress) public onlyOwner {
+        require(keeperRegistryAddress != address(0));
```
shuklaayush, line 144: Can change this to a custom error for consistency

shuklaayush (dismissed):
LGTM. The `onlyKeeperRegistry` modifier implies that the owner has to modify the registry address if it changes

tests/deployment/test_autonomous_dripper.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+@pytest.fixture(scope='module')
+def deployer():
+    return accounts[0]
+
+
+@pytest.fixture(scope='module')
+def dripper():
+    from scripts.deployment.deploy_autonomous_dripper import main
```
sajanrajdev, line 18: Love that you use the real deployment script within the test. Good stuff. I am going to have to loose my fear of deploying/verifying from brownie, if you say you have been successful from Brownie's latest it will make the deploying process a lot easier in the future. 

I wonder how you handle gas strategies and setting the priority fee. I couldn't find anywhere in the script where you do this programmatically.

contracts/AutonomousDripper.sol
shuklaayush: I think it comes from here:
https://github.com/smartcontractkit/chainlink/blob/develop/contracts/src/v0.8/upkeeps/EthBalanceMonitor.sol#L110-L113
Looks like it truncates the return data array size

tests/deployment/test_autonomous_dripper.py
jorijnsmit: https://eth-brownie.readthedocs.io/en/stable/core-contracts.html#transaction-parameters

jorijnsmit: https://eth-brownie.readthedocs.io/en/stable/core-gas.html

sajanrajdev:
@jorijnsmit, all tests for the autonomous dripper LGTM and are passing. One test for the emissions dripper is failing though:
image

It seems to be due to TechOps not being properly simulated. Adding a fixture for it, did the trick on my side:
```
@pytest.fixture(scope='module')
def techops():
    return accounts.at(registry.eth.badger_wallets.techops_multisig, force=True)
```
image

tests/deployment/test_autonomous_dripper.py
sajanrajdev: You should look into Andrii's work on the `bader-utils` for the customizable gas-strategies: https://github.com/Badger-Finance/badger-utils#using-gas_utils-module

contracts/AutonomousDripper.sol
GalloDaSballo: Would just add comment

contracts/AutonomousDripper.sol
```diff
@@ -0,0 +1,187 @@
...
+     * @dev Contains the logic that should be executed on-chain when
+     * `checkUpkeep` returns true.
+     */
+    function performUpkeep(bytes calldata performData) external override onlyKeeperRegistry whenNotPaused {
+        if ((block.timestamp - lastTimestamp) > interval) {
+            address[] memory assetsHeld = abi.decode(performData, (address[]));
+            for (uint idx = 0; idx < assetsHeld.length; idx++) {
+                if (IERC20(assetsHeld[idx]).balanceOf(address(this)) > 0) {
+                    VestingWallet.release(assetsHeld[idx]);
+                    lastTimestamp = block.timestamp;
```
GalloDaSballo, line 107: Can save gas by putting this after the loop (something like if(nonZero) -> lastTimestamp = block.timestamp

GalloDaSballo (approved):
See my notes, beside some gas LGTM

contracts/AutonomousDripper.sol
jorijnsmit: thanks, implemented in 276fe10

jorijnsmit:
confident enough to deploy to mainnet, looking for last two greenlights.

jorijnsmit:
@sajanrajdev to be fair that test that failed is for the old `EmissionsDripper`, not `AutonomousDripper`. i see now what alex is talking about with creating a too big repo with outdated stuff 😝 

anyway i think you are right, making a module wide fixture for it is better. however, here they all pass for me. sometimes it is just a weird thing in the ganache instance i think, and rerunning it will simply already do the trick.

i will specify the relevant tests next time; `brownie test tests/deployment/test_autonomous_dripper.py --gas`

contracts/AutonomousDripper.sol
jorijnsmit: yes you are right.

however i opted for removing it all together (7b117b7), since there is no security risk to setting it to 0 and can always be reset.

contracts/AutonomousDripper.sol
jorijnsmit: 276fe10b45fa3573c55ded5b12fc33f71bee86d9

shuklaayush (approved):
LGTM

jorijnsmit:
linking #404

PR #422, opened 2022-05-04
feat: perform transfers onfork instead of to generate a csv

jorijnsmit:


petrovska-petro (approved):
Amounts looks good, test release in dripper has being done successfully ✅

jalbrekt85 (approved):

PR #425, opened 2022-05-06
fix: typos

jorijnsmit:
dont break your head too much over that BAL#506, i think it is a shortcoming of the balpy module.

maybe write one more unit test for the claim? after i think at least all reqs of the ticket are met: #412


jalbrekt85 (approved):
nice changes, thanks @jorijnsmit

Issue #413, opened 2022-04-28
stable farm rebalance q2


gosuto-inzasheru:
treasury pod came up with the following new allocations:
- AAVE - USDC: $614,746
- AAVE - USDT: $614,746
- AAVE - FEI: $307,373
- Balancer - 3pool: $307,373 [depends on Badger-Finance/badger-multisig/issues/412]
- Compound - DAI: $614,746
- Convex - ust-(wormhole): $307,373
- GMX - GLP: $307,373 [this will take some r&d, can be left for a later script]

possible required preparation:
- [ ] send all 3pool from trops to vault

im thinking the following steps in one script:
- [x] withdraw all stables from convex to 3pool or eurs
- [x] swap 3pool to 614_746 usdc
- [x] swap 3pool to 614_746 usdt
- [x] swap 3pool to 307_373 fei
- [x] swap 3pool to 307_373 ust
- [x] swap all eurs to >614_746 dai
- [x] deposit (and stake) into aave
- [x] deposit (and stake) into balancer
- [x] deposit (and stake) into compound
- [x] deposit (and stake) into convex

this can then be left for another time:
- [ ] deposit (and stake?) into gmx

jorijnsmit:
suggesting to wait with the 3pool deposit on balancer; there will be a new contract deployed for it in the next couple of weeks due to a bug: https://forum.balancer.fi/t/vulnerability-disclosure/3179

petrovska-petro:
https://etherscan.io/tx/0x780b8ec118eb57bdef2cfa446f65e08bc83807f7b44d1d4544a90d71a4a2ac92

gosuto-inzasheru:
bpt_3pool deposit and aura stake: https://etherscan.io/tx/0xebe14af59a85197b8bd6ae9ec9f0e52cbc42ff4a5ccb6b085d74cc5a679e233b

Issue #435, opened 2022-05-10
Deploy Balancer BADGER/WBTC Badger 80/20 TCL


Tritium-VLK:
Treasury Request: https://docs.google.com/document/d/1EHo8sNuyB0P64zeBjo0y_dmCNGlsByj8ZJmW3jrKzKI/edit?usp=sharing

Approval: https://discord.com/channels/785315893960900629/971131359261241385/971540395790053406

The BIP calls for upto 15 WBTC and required BADGRE expected to number around 300,000 to be deposited into BADGER/WBTC 80/20 balancer pool and staked for farming.

Note that the pool currently has low liquidity and may not be trading at the market price.  It would be good to be mindful of this, and make the investment slowly in a number of chunks in order to ensure that we don't open unnecessarily large arb ops. 

As this is a large investment @Hal69k(aka 1500Badger) would like to be involved in the decisions about specific investment sizes and timing.  Please consult him before making investments.

petrovska-petro:
Test round:

- [ ] deposit & stake 0.05 wbtc and its equivalent of badger

- [ ] `unstake_all_and_withdraw_all` from prev action

petrovska-petro:
deposit & stake: https://etherscan.io/tx/0xdcc1dd7b3f8c8a63a82a1a1fa1b45b495fe6dcd2cdec2db9651685de46e6a33d

petrovska-petro:
```
snapshot result for 0x042B32Ac6b453485e357938bdC38e0340d4b9276:
                                   balance_before               balance_after             balance_delta
symbol                                                                                                 
BADGER                 117,667.483808517678617726  118,768.709776643171477881  1,101.225968125492860155
WBTC                        21.368888020000000000       21.408154930000000000      0.039266910000000000
20WBTC-80BADGER              4.592033090249046259        0.000000000000000000     -4.592033090249046259
20WBTC-80BADGER-gauge      279.336916265198500000        0.000000000000000000   -279.336916265198500000 
```

```
Events In This Transaction
--------------------------
├── Balancer 20WBTC-80BADGER Gauge Deposit (0xAF50825B010Ae4839Ac444f6c12D44b96819739B)
│   └── UpdateLiquidityLimit
│       ├── user: 0x042B32Ac6b453485e357938bdC38e0340d4b9276
│       ├── original_balance: 0
│       ├── original_supply: 12063750852839724806382
│       ├── working_balance: 0
│       └── working_supply: 4825500341135889922551
[...]

PR #444, opened 2022-05-12
stable farm rebalance script + balancer badger/wbtc pool tests

jalbrekt85:
- handles #413
- added test for deposits/withdrawals from our balancer pool 
- the balancer stablemath for deposits has been removed for now, so no `minAmountOut` type calc is present for those. the bpt out from balancer 3 pool should be checked carefully


petrovska-petro (approved):
- reviewed the script portion thru #447 which got merged

- test looks good, same functions were tested in #435 successfully 

- address added on the book are correct

jorijnsmit (approved):

PR #447, opened 2022-05-13
slight different way of shuffling. adds sweep func for when tx is execd.

jorijnsmit:


petrovska-petro (commented):
nothing against this way of handling the stable farming reorg, ends meeting goal targets by treasury pod. will leave final decision to @jalbrekt85 

re the claim of the univ3 fees is cool, but thought would be quarterly cadence

jalbrekt85 (approved):
lots of ways to do the ordering here. this way seems to better, esp for deposting equal amounts into balancer and reducing number of txs overall

Issue #442, opened 2022-05-12
add all new owners of the treasury vault atomically


jorijnsmit:
following bip89: https://forum.badger.finance/t/bip-89-badgerdao-treasury-management-policy/5544/2

dapp and signers are current owners, so only need to add the 8 new ones:
> Initial Council nominees
    Jonto [T]
    ~~Dapp [S]~~
    Lipp [T]
    Po [T]
    1500 [T]
    Gosuto [M]
    Saj [S]
    Ayush [M]
    Petro [S]

jorijnsmit:
```
Events In This Transaction
--------------------------
└── Gnosis Safe (0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e)
    ├── AddedOwner
    │   └── owner: 0xFb055FecDaDf080B10c5a5bFaf7B3b70dd660547
    ├── ChangedThreshold
    │   └── threshold: 5
    ├── AddedOwner
    │   └── owner: 0xaC7B5f4E631b7b5638B9b41d07f1eBED30753f16
    ├── AddedOwner
    │   └── owner: 0x9c8C8bcD625Ed2903823b0b60DeaB2D70B92aFd9
    ├── AddedOwner
    │   └── owner: 0x66496eBB9d848C6A8F19612a6Dd10E09954532D0
    ├── AddedOwner
    │   └── owner: 0x6C6238309f4f36DFF9942e655A678bbd4EA3BC5d
    ├── AddedOwner
    │   └── owner: 0xD10617AE4Da733d79eF0371aa44cd7fa74C41f6B
    ├── AddedOwner
    │   └── owner: 0x96AC69183216074dd8CFA7A380e873380445EaDc
    ├── AddedOwner
    │   └── owner: 0x0a9af7FAba0d5DF7A8C881e1B9cd679ee07Af8A2
    └── ExecutionSuccess
        ├── txHash: 0x04fc415ded86b6c63b2b3c7e1e28dbf6c968a333872b12a09db02c81ddee85b3
        └── payment: 0
```
```
{
│   'ethereum_client': <gnosis.eth.ethereum_client.EthereumClient object at 0x105688a60>,
│   'safe_address': '0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e',
│   'to': '0x40A2aCCbd92BCA938b02010E17A5b8929b49130D',
│   'value': 0,
[...]

PR #450, opened 2022-05-16
script: for loop treasury council members and add them as owner

jorijnsmit:
script to solve #442


petrovska-petro (approved):
Script looks good and changes threshold to 5

Confirming at my end is the right address for my signer

would be great to get same confirmation for all the addr additions on this PR as there is no other way to confirm address veracity

dapp-whisperer:
confirmed dapp address

sajanrajdev (approved):
Iterates through list of signers' addresses and adds them to the multisig if not already added. Also, sets the threshold to 5. 

LGTM

Also, confimed Saj's address.

Hal69k (commented):
Confirming 1500 address is correct

shuklaayush:
Confirming ayush's address

MrrPo:
Confirming Po's address is correct

Issue #464, opened 2022-05-18
Voting snapshot with gnosis safe


petrovska-petro:
Currently only gnosis safes can sign as a contract in the snapshot spaces, since only subgraphs and specific servers are support it. 

Ideally, we'd like to test the flow of getting a vote signed programmatically and post into the safe for signing

PR #465, opened 2022-05-18
Initial R&D for voting on snapshot space

petrovska-petro:
Starts tackling #464. still draft, will wait for another proposal being posted

- [x] Sign thru UI

- [x] Mirror in the script and perhaps change voting decision

**Note:** in this pr i am including a new argument to `post_safe_tx(...)`method, **safe_tx_arg** which will allow us if we build the safe_tx ourselves and we're confident thru local testing or tenderly simulation of the targets, raw data call and operation type to allow us posting txs which locally are getting some nasty reverts (e.g: rari's tx...etc)


petrovska-petro:
vote: https://etherscan.io/tx/0x965223bd42683bea22c03cd722f2a1fe3faf63b650603d0469c3ff71c8bfe333

proposal: https://vote.convexfinance.com/#/proposal/QmRkUTjJFezEL4mc8JKaN2BRpx7qQq4mTZhJytch2AKTsB

Screenshot 2022-05-19 at 18 56 08

petrovska-petro:
tx sent thru the script: https://etherscan.io/tx/0x2e08286986f16587555fa5bbc214fe6d7c4fab1924a492495ca9a19dceb9e4bc

notify the relayer thru the [endpoint]("https://snapshot-relayer.herokuapp.com/api/message") and caught the voting preference at later type and [processed](https://github.com/snapshot-labs/snapshot-relayer/blob/master/src/check.ts#L47) and casted into the hub

Screenshot 2022-05-30 at 16 43 35


proposal: https://vote.convexfinance.com/#/proposal/QmWuQY5XTyWmHLQrnxves9CEQePKUTrx52WVMC3ayrPz1m

great_ape_safe/great_ape_safe.py
```diff
@@ -232,13 +232,16 @@ def remove_ansi_escapes(text):
             f.write(remove_ansi_escapes(buffer.getvalue()))
 
 
-    def post_safe_tx(self, skip_preview=False, events=True, call_trace=False, reset=True, silent=False, post=True, replace_nonce=None, log_name=None, csv_destination=None, gas_coef=1.5):
+    def post_safe_tx(self, skip_preview=False, events=True, call_trace=False, reset=True, silent=False, post=True, replace_nonce=None, log_name=None, csv_destination=None, gas_coef=1.5, safe_tx_arg=None):
```
jorijnsmit, line 235:
```suggestion
    def post_safe_tx(self, skip_preview=False, events=True, call_trace=False, reset=True, silent=False, post=True, replace_nonce=None, log_name=None, csv_destination=None, gas_coef=1.5, safe_tx=None):
```

great_ape_safe/great_ape_safe.py
```diff
@@ -232,13 +232,16 @@ def remove_ansi_escapes(text):
...
         # build a gnosis-py SafeTx object which can then be posted
         # skip_preview=True: skip preview **and with that also setting the gas**
         # events, call_trace and reset are params passed to .preview
         # silent=True: prevent printing of safe_tx attributes at end of run
         # post=True: make the actual live posting of the tx to the gnosis api
-        safe_tx = self.multisend_from_receipts()
+        if safe_tx_arg:
+            safe_tx = safe_tx_arg
+        else:
+            safe_tx = self.multisend_from_receipts()
```
jorijnsmit, line 244:
```suggestion
        if not safe_tx:
            safe_tx = self.multisend_from_receipts()
```

helpers/addresses.py
```diff
@@ -457,6 +457,9 @@
         "feed_registry": "0x47Fb2585D2C56Fe188D0E6ec628a38b74fCeeeDf",
         "keeper_registry": "0x7b3EC232b08BD7b4b3305BE0C044D907B2DF960B",
     },
+    "gnosis": {
+        "sign_message_lib": "0xA65387F16B013cf2Af4605Ad8aA5ec25a2cbA3a2"
+    }
```
jorijnsmit, line 462:
```suggestion
    "gnosis": {
        "sign_message_lib": "0xA65387F16B013cf2Af4605Ad8aA5ec25a2cbA3a2",
    },
```

interfaces/gnosis/ISignMessageLib.sol
```diff
@@ -0,0 +1,12 @@
...
+pragma solidity ^0.7.0;
+
+interface ISignMessageLib {
+    function getMessageHash(bytes memory message)
+        external
+        view
+        returns (bytes32);
+
+    function signMessage(bytes memory _data) external;
+}
```
jorijnsmit, line 12:
```suggestion
// SPDX-License-Identifier: LGPL-3.0-only
// !! THIS FILE WAS AUTOGENERATED BY abi-to-sol v0.5.3. SEE SOURCE BELOW. !!
pragma solidity ^0.7.0;

interface ISignMessageLib {
    event SignMsg(bytes32 indexed msgHash);

    function getMessageHash(bytes memory message)
        external
        view
        returns (bytes32);

    function signMessage(bytes memory _data) external;
}
```

scripts/snapshot/CVX/vote_proposal.py
```diff
@@ -0,0 +1,70 @@
...
+
+    payload = {
+        "version": "0.1.3",
+        "timestamp": str(int(time.time())),
+        "space": "cvx.eth",
+        "type": "vote",
+        "payload": {
+            "proposal": proposal,
+            "choice": choice,  # json.dumps(choice_json, separators=(",", ":")),
+            "metadata": json.dumps({}),
```
jorijnsmit, line 35: what does the `json.dumps` here do?

scripts/snapshot/CVX/vote_proposal.py
```diff
@@ -0,0 +1,70 @@
...
+
+    # notify relayer to watch for this tx to be mined so it is included in the snapshot proposal
+    # https://github.com/snapshot-labs/snapshot-relayer/blob/7b656d204ad78fc9c06cedafcb4288aa47775adc/src/api.ts#L36
+    response = requests.post(
+        SNAPSHOT_VOTE_RELAYER,
+        headers=SNAPSHOT_DEFAULT_HEADERS,
+        data=json.dumps(
+            {
+                "address": test_msig.address,
+                "msg": json.dumps(payload, separators=(",", ":")),
```
jorijnsmit, line 57:
```suggestion
                "msg": payload_stringify,
```

jorijnsmit (commented):
ok so i understand this posts a vote to the snapshot api, but because it does so on behalf of a multisig it cannot provide a sig yet.

it will then wait until it sees a tx from that multisig with the exact same payload, hash and timestamp. only then does the vote get registered.

do you know how much time there can be between posting to the api and the tx needing to be mined?

petrovska-petro:
> ok so i understand this posts a vote to the snapshot api, but because it does so on behalf of a multisig it cannot provide a sig yet.
> 
> it will then wait until it sees a tx from that multisig with the exact same payload, hash and timestamp. only then does the vote get registered.
> 
> do you know how much time there can be between posting to the api and the tx needing to be mined?

yes, understanding is correct

it will look for it for 2 days max, so needs to be mined within that time interval

scripts/snapshot/CVX/vote_proposal.py
petrovska-petro: it simply stringify the braces, but will be same if we leave it as **'{}'**. wanted to leave it like this as reminder that can accept later on for other occasion a dictionary with other info within it

jorijnsmit (approved):
looking good, important step in decentralising our voting process.

tested multiple times on our live 1.3.0 msig ([nonce 5](https://etherscan.io/tx/0xbd2ed7cf22d904812aa6d6a3259dea7e329ab2ae6a1ded7a90cd1f84bbf9e0db), [nonce 6](https://etherscan.io/tx/0x2e08286986f16587555fa5bbc214fe6d7c4fab1924a492495ca9a19dceb9e4bc)).

PR #471, opened 2022-05-19
dripper deloyed; script to register it as an upkeep

jorijnsmit:
i confirmed with the chainlink team that the `encryptedEmail` and `source` parameters can remain null.


petrovska-petro (approved):
looks good and v excited to get this rolling 🔥 

wondering if the `encryptedEmail` could be leveraged for the notifications for the upkeeps, perhaps worthy to have a general email for these logs purposes if understand properly from their docs seems

jorijnsmit:
yes maybe in the future, would need some better understanding. i am not sure when exactly it sends notifications to that email address; i have deployed on rinkeby and have not received one email, not on successful upkeeps nor failed upkeeps...

anyway out of scope for deployment for this quarter i would say.

jorijnsmit:
linking #404

Issue #460, opened 2022-05-17
week 20: native topups to tree


jorijnsmit:
see #431 for last week's

hopefully last manual one, looking to deploy dripper tomo.

jorijnsmit:
https://etherscan.io/tx/0x0a90ee1ad5cea035304f2b226b7eca061f3071e24689d34c08d6ab727bd43c39

Issue #449, opened 2022-05-13
Send trops BADGER for Balancer TCL completion


petrovska-petro:
Requires 349364.20573405287 BADGER

```
15*30637 / .2 * .8 / 4.71 = 390280.2547770701

390280.2547770701 + 40916.049043017229664256 = 431196.3038200873

349364.20573405287 * 1.1 (buffer) ~= 475k
```

already *40916.049043017229664256* badger had being deposited. [ref](https://etherscan.io/tx/0x6f81605c66b6c65fbd073e576222cfc25b753c342a7caf408bf61f6a961c358b). but they're meant for q2 emissions.

related to #435

petrovska-petro:
```
snapshot result for 0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e:
                      balance_before                 balance_after                balance_delta
symbol                                                                                         
BADGER  7,069,626.388705043165002833  6,594,626.388705043146128465  -475,000.000000000018874368 
```

```
{
│   'ethereum_client': <gnosis.eth.ethereum_client.EthereumClient object at 0x7fc7101c1c70>,
│   'safe_address': '0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e',
│   'to': '0x40A2aCCbd92BCA938b02010E17A5b8929b49130D',
│   'value': 0,
│   'data': HexBytes('0x8d80ff0a00000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000099003472a5a71965499acd81997a54bba8d852c6e53d00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044a9059cbb000000000000000000000000042b32ac6b453485e357938bdc38e0340d4b9276000000000000000000000000000000000000000000006495cd356e241400000000000000000000'),
│   'operation': 1,
│   'safe_tx_gas': 0,
│   'base_gas': 0,
│   'gas_price': 0,
│   'gas_token': '0x0000000000000000000000000000000000000000',
│   'refund_receiver': '0x0000000000000000000000000000000000000000',
│   'signatures': b'',
│   '_safe_nonce': 25,
│   '_safe_version': '1.3.0',
│   '_chain_id': None
}
```

petrovska-petro:
https://etherscan.io/tx/0xc0dd377dfbb1f64b5505a9151bd923f7b7917333b5cea602cb9470037dcbf022

PR #472, opened 2022-05-19
week 20 topups

jalbrekt85:
#460


petrovska-petro (approved):
LGTM!

we will be able to run it once we get vault_msig badger funds

jorijnsmit (approved):
indeed, depends on #449

PR #477, opened 2022-05-23
feat: delay bal3pool deposit, add separate stake method

jorijnsmit:
delay bal3pool deposit in light of https://forum.balancer.fi/t/vulnerability-disclosure/3179


petrovska-petro (approved):
great security measure and clean refactor on stake - balancer class. 👌

Issue #83, opened 2022-02-02
airdrop remdigg to victims that had (underlying) digg stolen


jorijnsmit:

jorijnsmit:
https://etherscan.io/tx/0xd93299c8cd177037968fb056e817363aab22e2398d794a7e8d4cb3297787dfde

PR #485, opened 2022-05-23
script: generate digg airdrop csv

jorijnsmit:
this is basically the same script as the one we used to airdrop the remdigg itself: https://github.com/Badger-Finance/badger-multisig/issues/83, just replacing the token address with digg's.


petrovska-petro (approved):
confirm unique diff with [checker](https://www.diffchecker.com/cLJmdGz8). got identical output ✅

btcookies (approved):
LGTM

Issue #480, opened 2022-05-23
Prevent DIGG rebasing by removing providers to the oracle


jorijnsmit:
(original writeup by @mitche50; i hate how original author gets lost when converting draft to issue...)

To update the DIGG Median Oracle ([0x058ec2Bf15011095a25670b618A129c043e2162E](https://etherscan.io/address/0x058ec2bf15011095a25670b618a129c043e2162e#readContract)) providers need to push reports.  Removing the providers from the oracle prevents the price from being updating and prevents rebasing.

You can pull providers by the below command (there are 3 currently in indexes 0-2)
```
oracle.providers(0)
```

You can remove providers by the below command
```
oracle.removeProvider(address)
```

Provider list:
- [0x72dc16CFa95beB42aeebD2B10F22E55bD17Ce976](https://etherscan.io/address/0x72dc16CFa95beB42aeebD2B10F22E55bD17Ce976)
- [0xB572f69edbfC946af11a1b3ef8D5c2f41D38a642](https://etherscan.io/address/0xB572f69edbfC946af11a1b3ef8D5c2f41D38a642)
- [0x73083058e0f61D3fc7814eEEDc39F9608B4546d7](https://etherscan.io/address/0x73083058e0f61D3fc7814eEEDc39F9608B4546d7)

jorijnsmit:
```
Events In This Transaction
--------------------------
├── IMedianOracle (0x058ec2Bf15011095a25670b618A129c043e2162E)
│   ├── ProviderRemoved
│   │   └── provider: 0x72dc16CFa95beB42aeebD2B10F22E55bD17Ce976
│   ├── ProviderRemoved
│   │   └── provider: 0x73083058e0f61D3fc7814eEEDc39F9608B4546d7
│   └── ProviderRemoved
│       └── provider: 0xB572f69edbfC946af11a1b3ef8D5c2f41D38a642
│   
└── Gnosis Safe (0xB65cef03b9B89f99517643226d76e286ee999e77)
    └── ExecutionSuccess
        ├── txHash: 0xa20f48dc116b7e2b6ae731591a7b7fdfbc7d449fd82d02281f0dc2d5c2320b50
        └── payment: 0
```
```
{
│   'ethereum_client': <gnosis.eth.ethereum_client.EthereumClient object at 0x11bc30f70>,
│   'safe_address': '0xB65cef03b9B89f99517643226d76e286ee999e77',
[...]

PR #487, opened 2022-05-24
script: remove all providers from digg oracle

jorijnsmit:
would solve #480


scripts/issue/480/disable_digg_oracle.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def main():
+    dev = GreatApeSafe(registry.eth.badger_wallets.dev_multisig)
+    oracle = interface.IMedianOracle(
+        registry.eth.oracles.oracle, owner=dev.account
+    )
+
+    for _ in range(oracle.providersSize()):
+        oracle.removeProvider(oracle.providers(0))
```
mitche50, line 20: This should be:
```
for i in range(oracle.providersSize()):
    oracle.removeProvider(oracle.providers(i))
```

jorijnsmit: no, because removing the first entry shortens the length of the array

jorijnsmit: https://etherscan.io/address/0x058ec2Bf15011095a25670b618A129c043e2162E#code#L424

mitche50: Interesting - good catch.  They could have learned something from balancer with linked lists to make it more gas efficient :D

This LGTM then!

mitche50 (approved):

btcookies (approved):
LGTM

PR #493, opened 2022-05-25
chore: initial migration progress

axejintao:


axejintao:
Partial for #365

sajanrajdev (approved):
Adapts the GreatApeSafe wrapper for the badger.registry to work with the newly deployed RegistryV2 (only on Fantom so far). Adds the new functions included in this new version as well as a script to migrate all keys from the Registry V1 to V2. 

Implementations to allow fo chain resiliency of the wrapper.

Missing keys from the Fantom addresses registry were also added and LG. The RegistryV2 address was also added, confirmed that its logic matches the developed and it was deployed by an address funded by Ops Executor 4.

Results from running the migrate_registry_keys for Fantom LGTM:
image

helpers/addresses.py
```diff
@@ -789,7 +792,9 @@
     "spookyswap": {
         "router": "0xF491e7B69E4244ad4002BC14e878a34207E38c29",
         "factory": "0x152eE697f2E276fA89E96742e9bB9aB1F2E61bE3"
-    }
+    },
+    "registry": "0xFda7eB6f8b7a9e9fCFd348042ae675d1d652454f",
+    "registryV2": "0xdc602965F3e5f1e7BAf2446d5564b407d5113A06",
```
jorijnsmit, line 797: getting a diff here with the registry v2 contract as it can be found here: https://github.com/jack-the-pug/badger-registry/blob/2.0/contracts/BadgerRegistry.sol

https://www.diffchecker.com/eIaJs6W3

all other addresses check out.

helpers/addresses.py
```diff
@@ -11,6 +11,7 @@
     "rewardsLogger" : "0x0A4F4e92C3334821EbB523324D09E321a6B0d8ec",
     "EmissionControl": "0x31825c0A6278b89338970e3eB979b05B27FAa263",
     "registry": "0xFda7eB6f8b7a9e9fCFd348042ae675d1d652454f",
+    "registryV2": "0xdc602965F3e5f1e7BAf2446d5564b407d5113A06",
```
jorijnsmit, line 14: this is not deployed

jorijnsmit, line 14: this is not deployed

helpers/addresses.py
```diff
@@ -823,6 +828,18 @@ def checksum_address_dict(addresses):
...
+    if chain_id == 1:
+        return registry.eth
+    elif chain_id == 137:
+        return registry.poly
+    elif chain_id == 56:
+        return registry.bsc
+    elif chain_id == 42161:
+        return registry.arbitrum
+    elif chain_id == 250:
+        return registry.ftm
```
jorijnsmit, line 841: nice, i like this a lot, @petrovska-petro and i were talking about doing this as well.

i dont see the harm in instead of passing `chain_id` to just test it against `chain.id` directly. this then also enables `from helpers.addresses import get_registry as contract_registry`

jorijnsmit, line 841: nice, i like this a lot, @petrovska-petro and i were talking about doing this as well.

scripts/badger/registry.py
```diff
@@ -1,19 +1,37 @@
 from brownie import web3
 from rich.console import Console
 from great_ape_safe import GreatApeSafe
-from helpers.addresses import registry
-
+from helpers.addresses import registry, get_registry
```
jorijnsmit, line 4:
```suggestion
from helpers.addresses import get_registry as registry
```

scripts/badger/registry.py
```diff
@@ -1,19 +1,37 @@
 from brownie import web3
```
jorijnsmit, line 1:
```suggestion
from brownie import chain, web3
```

jorijnsmit, line 1:
```suggestion
from brownie import chain, web3
```

scripts/badger/registry.py
```diff
@@ -1,19 +1,37 @@
...
 from rich.console import Console
 from great_ape_safe import GreatApeSafe
-from helpers.addresses import registry
-
+from helpers.addresses import registry, get_registry
+from brownie import chain
```
jorijnsmit, line 5:
```suggestion
```

jorijnsmit, line 5:
```suggestion
```

helpers/addresses.py
axejintao: i had this from before i added the get by chain, will remove for now

scripts/badger/registry.py
axejintao: wat

sajanrajdev (approved):
Changes LGTM and script continues to work as expected.

petrovska-petro (commented):
script looks good for migrating keys to new registry

- verified address added on book ✅

- [contract](https://ftmscan.com/address/0x00000b7665850f6b1e99447a68db1e83d8deafe3#code) logic deployed matches content in the [repo](https://github.com/Badger-Finance/badger-registry/blob/v0.2.1/contracts/BadgerRegistry.sol) ✅ 

- i like the addition of `get_registry()` leveraging `chain.id` 🔥 

only comment will be that if the current badger class will be initiated in other chain different than fantom will break as will not be able to find `contract_registry.registryV2` maybe worth adding some extra check-up while this registry format gets deployed in all chains to avoid breaking other scripts

great_ape_safe/ape_api/badger.py
```diff
@@ -25,32 +25,39 @@ class Badger():
...
+
         self.registry = interface.IBadgerRegistry(
-            registry.eth.registry, owner=self.safe.account
+            contract_registry.registry, owner=self.safe.account
         )
-        self.bribes_processor = interface.IBribesProcessor(
-            registry.eth.bribes_processor, owner=self.safe.account
+        self.registryV2 = interface.IBadgerRegistryV2(
+            contract_registry.registryV2, owner=self.safe.account
         )
```
jorijnsmit, line 58: yea i agree with @petrovska-petro here, this now only works on chains for which `registryV2` is present in the address book (currently only fantom).

temp solution:
```suggestion
        if chain.id == 250:
            self.registryV2 = interface.IBadgerRegistryV2(
                contract_registry.registryV2, owner=self.safe.account
            )
```

jorijnsmit (commented):
it makes sense to makes registry chain aware; it enables great_ape_safe to be used on every chain.

however, we will need to find a solution for querying contract addresses from it that are not known on the chain in question.

great_ape_safe/ape_api/badger.py
axejintao: i've gone for a bit more of a completed approach, please have a look

axejintao:
```
{'badger': <IBadger Contract '0x753fbc5800a8C8e3Fb6DC6415810d627A387Dfc9'>}
{'tree': <IBadgerTreeV2 Contract '0x89122c767A5F543e663DB536b603123225bc3823'>}
strat_bvecvx not defined for chain 250
timelock not defined for chain 250
bribes_processor not defined for chain 250
{'registry': <IBadgerRegistry Contract '0xFda7eB6f8b7a9e9fCFd348042ae675d1d652454f'>}
{'registryV2': <IBadgerRegistryV2 Contract '0xdc602965F3e5f1e7BAf2446d5564b407d5113A06'>}
```

jorijnsmit (approved):

petrovska-petro (approved):

PR #498, opened 2022-05-26
script: config gas station

jorijnsmit:
@btcookies can you have a look here, mostly at the printed dataframe when runnning `brownie run issue/497/config_gas_station`, and see if that makes sense? i think now is a good time to rethink the way we refill.

ie: sweep and deprecate these (no activity for almost a year now):
- guardian; `0x29F7F8896Fb913CF7f9949C623F896a154727919`
- keeper; `0x872213E29C85d7e30F1C8202FC47eD1Ec124BB1D`
- cycle bot; `0x68de9E2b015904530593426d934CE608e117Fa7A`
- earner; `0x46099Ffa86aAeC689D11F5D5130044Ff7082C2AD`
- harvester; `0x73433896620E71f7b1C72405b8D2898e951Ca4d5`
- external_harvester; `0x64E2286148Fbeba8BEb4613Ede74bAc7646B2A2B`

also the minimal top up of .1 eth might not make sense for all the bots? currently the gas station is configured to refill once every 24 hours. maybe for the cycle0 bot for example, it might make more sense to set the min top up higher or something? just to prevent it contantly being topped up?


scripts/issue/497/config_gas_station.py
```diff
@@ -0,0 +1,50 @@
...
+    for label, addr in registry.eth.badger_wallets.items():
+        min_bal = 2e18
+        min_top_up = .1e18
+        if not label.startswith('ops_') or 'multisig' in label:
+            continue
+        if 'executor' in label:
+            min_bal = 1e18
+        if label == 'ops_botsquad':
+            min_bal = 5e18
+        if label == 'ops_botsquad_cycle0':
```
btcookies, line 23: do you think we should store the custom configs in a dict somewhere?

btcookies, line 23: do you think we should store the custom configs in a dict somewhere?

jorijnsmit: yea we had something like that in the original refill script, but being able to already cut out so many of the old addresses, it looks like the exceptions are getting less and less. also there is dealing with both `min_bal` and `min_top_up`, basically making it more work to intro a dict then to write these three ifs.

tbh im hoping we can move towards deprecating the botsquad accounts as well and replace them by chainlink, making this gas station only for non bot eoas (deployers and executors).

petrovska-petro (approved):
- confirmed addresses in book ✅ 

- ownership transfer triggered ✅ 

- contract content matches the [repo](https://github.com/Badger-Finance/badger-multisig/blob/main/contracts/GasStationExact.sol) content and interface updated ✅ 

- neat link class integration, surely will be use quite a bit!

btcookies (approved):
LGTM

Issue #446, opened 2022-05-13
uni v3 class' `collect_fees` method relies on `calc_all_accum_fees`


jorijnsmit:
the uni v3 class' `collect_fees` method relies on `calc_all_accum_fees` to see if there are fees to be collected, but `calc_all_accum_fees` returns all fees accumulated *ever*, not only those that are still pending.

i had to remove asserts here:
https://github.com/Badger-Finance/badger-multisig/blob/6879a310ef0ea323942b4916898e377bd0be162d/great_ape_safe/ape_api/uni_v3.py#L116-L147

either we have to write an alternative to `calc_all_accum_fees` or leave as is and for loop all token_ids.

i guess this also affects `positions_info`:
```
owner: 0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e
{
│   'nonce': 0,
│   'operator': '0x0000000000000000000000000000000000000000',
│   'token0': '0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599',
│   'token1': '0x3472A5A71965499acd81997a54BBA8D852C6E53d',
│   'fee': 3000,
│   'tickLower': 299340,
│   'tickUpper': 306240,
│   'liquidity': 217886475739066675,
│   'feeGrowthInside0LastX128': 41513451917362791121577959067,
│   'feeGrowthInside1LastX128': 784591595948463347023150117924402115620696,
│   'tokensOwed0': 0,
│   'tokensOwed1': 0
}
accumulated fees:
6.2205931 WBTC
26520.860934187214 BADGER
owner: 0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e
{
│   'nonce': 0,
│   'operator': '0x0000000000000000000000000000000000000000',
│   'token0': '0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599',
│   'token1': '0x3472A5A71965499acd81997a54BBA8D852C6E53d',
│   'fee': 3000,
│   'tickLower': 313200,
│   'tickUpper': 320100,
│   'liquidity': 176860535297231819,
│   'feeGrowthInside0LastX128': 134763002747325240849479048594,
│   'feeGrowthInside1LastX128': 6549213728763063372637121633754183731383027,
│   'tokensOwed0': 0,
│   'tokensOwed1': 0
}
accumulated fees:
1.55107467 WBTC
8024.681829727605 BADGER
[...]

PR #510, opened 2022-05-27
update: only exec fee collection if there is any fee to be collected

petrovska-petro:
Tackles #446

Since we [claimed](https://etherscan.io/tx/0x780b8ec118eb57bdef2cfa446f65e08bc83807f7b44d1d4544a90d71a4a2ac92) recently the fees now there may not be any claims available to test it properly. For testing, please leverage the forking at specific block height or locally to explore the values returned in the `.call(params, block_identifier)`. Use block below **14856066**


jalbrekt85 (approved):
nice use of  `.call()` to check fee amounts, lgtm

jorijnsmit (approved):
yeeees

Issue #520, opened 2022-06-01
Upgrade DIGG Proxy


mitche50:
Upgrade DIGG Proxy using new logic deployed here: https://etherscan.io/address/0x020eb84309243ed4b8e6c197af145125dde4afda#code#L727

Reference PR: https://github.com/Badger-Finance/badger-system/pull/267

sajanrajdev:
Upgrade queue to Timelock data:
```
{
    "data": "000000000000000000000000798d1be841a82a273720ce31c822c61a67a601c3000000000000000000000000020eb84309243ed4b8e6c197af145125dde4afda",
    "eta": 1654526146,
    "eth": 0,
    "signature": "upgrade(address,address)",
    "target": "0x20Dce41Acca85E8222D6861Aa6D23B6C941777bF"
}
```

sajanrajdev:
Executed: https://etherscan.io/tx/0xd99b539b452504119d34b1693a952ba6435dcbdc582dd97c59d9dbbb22d8abd4

PR #523, opened 2022-06-01
Digg upgrade script

sajanrajdev:
Closes Issue #520 

Upgrades Digg to new logic with oneTimeMint, Sweep and Toggle Rebase

Run queue with:
```
brownie run scripts/issue/520/upgrade_digg.py
```
Run execution with:
```
brownie run scripts/issue/520/upgrade_digg.py main false
```
Run simulation with:
```
brownie run scripts/issue/520/upgrade_digg.py main false true
```

### DIFF
--- a/helpers/addresses.py
+++ b/helpers/addresses.py
@@ -22,6 +22,7 @@ ADDRESSES_ETH = {
         "tree_2022_q2": "0xfA55d407F48849aE98a6e4f11Dc5C18E035F46Ec",
     },
     "bribes_processor": "0xb2Bf1d48F2C2132913278672e6924efda3385de2",
+    "digg_monetary_policy": "0x327a78D13eA74145cc0C63E6133D516ad3E974c3",
     # the wallets listed here are looped over by scout and checked for all treasury tokens
     "badger_wallets": {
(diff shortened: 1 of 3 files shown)


mitche50 (approved):

shuklaayush:
I think we should also disable rebasing in the same script using `toggleRebase`

jorijnsmit:
> The error comes from it trying to call `oneTimeMint` from Digg with the old logic. Not sure why Brownie does that but the simulation checks confirms that the function will be callable after upgrade.

i think this occurs because in running the preview, it can only gather the receipts for the msig txs. the `proxyAdmin.upgrade(digg.address, NEW_LOGIC, {'from': timelock})` thus doesnt get taken into account after preview resets the chain.

locally i isolated the upgrade logic into a different function, and then the preview for the one time mint does run successfully.

jorijnsmit (commented):
thorough upgrade script saj!

good to go imo except for ayush's comment on explicitly disabling rebasing as well.

mitche50 (approved):
LGTM - Saj showed tx trace from simulation which shows `Rebase paused` revert as expected as you cannot use `with brownie.reverts()` on scripts

jorijnsmit (approved):
lessgooo

mitche50 (approved):

PR #527, opened 2022-06-02
feat: consolidate rari redeems into one multicall

jorijnsmit:
we all missed this bug: `to=registry.fusdc.address` so had to hack it anyway.


jalbrekt85 (approved):
lgtm!

petrovska-petro (approved):
good catch re the `receiver` field

Issue #533, opened 2022-06-07
open 1.5x btc long


gosuto-inzasheru:
https://forum.badger.finance/t/1-5x-btc-long-using-5-treasury-btc-and-oasis/5693

gosuto-inzasheru:
testing script with 1btc. plan is to leave open for a week or two, monitor and then test closing again.
```
Events In This Transaction
--------------------------
├── Wrapped BTC (0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599)
│   └── Approval
│       ├── owner: 0x042B32Ac6b453485e357938bdC38e0340d4b9276
│       ├── spender: 0x7d2768dE32b0b80b7a3454c06BdAc94A69DDc7A9
│       └── value: 100000000
│   
├── Aave interest bearing WBTC (0x9ff58f4fFB29fA2266Ab25e75e2A8b3503311656)
│   ├── Transfer
│   │   ├── from: 0x0000000000000000000000000000000000000000
│   │   ├── to: 0x464C71f6c2F760DdA6093dCB91C24c39e5d6e18c
│   │   └── value: 35142
│   └── Mint
│       ├── from: 0x464C71f6c2F760DdA6093dCB91C24c39e5d6e18c
│       ├── value: 35142
│       └── index: 1002449659105953542280035001
│   
├── InitializableImmutableAdminUpgradeabilityProxy (0x7d2768dE32b0b80b7a3454c06BdAc94A69DDc7A9)
│   └── ReserveDataUpdated
│       ├── reserve: 0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599
│       ├── liquidityRate: 247419749277288490345269
│       ├── stableBorrowRate: 37625895729359523968843211
│       ├── variableBorrowRate: 6100716583487619175074569
│       ├── liquidityIndex: 1002449659105953542280035001
│       └── variableBorrowIndex: 1010971538682397617093127628
│   
├── Wrapped BTC (0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599)
│   └── Transfer
│       ├── from: 0x042B32Ac6b453485e357938bdC38e0340d4b9276
│       ├── to: 0x9ff58f4fFB29fA2266Ab25e75e2A8b3503311656
│       └── value: 100000000
│   
├── InitializableImmutableAdminUpgradeabilityProxy (0xd784927Ff2f95ba542BfC824c8a8a98F3495f6b5)
[...]

PR #534, opened 2022-06-07
script: deploy ds proxy via maker for oasis.app use

jorijnsmit:
some considerations in the context of #533:
- understanding the inner workings of the dai system in such a way to be able to comfortably open, increase, decrease and close leveraged positions is not currently worth our time
- i suggest to use the oasis.app ui instead, available as a gnosis app
- this makes the tx considerably less trustless and less verifiable; for this reason this script uses treasury ops to deploy the proxy from, opposed to the vault
- we could even consider a separate wallet for this (although imo this is exactly what treasury ops is for)
- another reason to not use the vault is that we might need to adjust the leverage quickly during unexpected market turmoil, in which case a lower signing threshold is favourable.

will discuss these points with treasury council as well.


petrovska-petro (dismissed):
- Confirmed addr of proxy registry on [dev-guide](https://github.com/makerdao/developerguides/blob/master/gnosis-multisig/dsr-gnosis-multisig-guide/dsr-gnosis-multisig-guide.md#step-1-building-a-ds-proxy-contract-for-the-multisig) ✅ 

- Creation of proxy looks correct and events emitted in tx in fork env aligns ✅

helpers/addresses.py
```diff
@@ -498,6 +498,8 @@
     "across_bridge": {
         "hub_pool": "0x6Bb9910c5529Cb3b32c4f0e13E8bC38F903b2534",
     },
+    "maker": {
```
petrovska-petro, line 501: forgot to mention that locally needed this little modification, was missing a curly bracket

```suggestion
    "maker": {
        "proxy_registry": "0x4678f0a6958e4D2Bc4F1BAF7Bc52E8F3564f3fE4",
    },
```

jorijnsmit: pushed

petrovska-petro (approved):

Issue #318, opened 2022-03-30
Sweep stuck bSLP from ibBTC/wBTC SLP Vault Contract


sajanrajdev:
A user transferred around ~30k worth of bSLP to the ibBTC/wBTC SLP contract accidentally: https://etherscan.io/tx/0x5ecf417d17afefc7664636157e714327c5630cd2abb56f7a88a59cffd8e3bd68

The vault contract must be upgraded to sweep these tokens and transfer them back to the user.

Tritium-VLK:
User agreed to pay a fee of the greater of

-  20% of the value recovered
- gas costs + $1000 USD in labor

PR #535, opened 2022-06-07
feat: recover user stuck assets

sajanrajdev:
Tackles #318 

Sweeps stuck bSLP tokens stuck on itself, charges a 20% service fee, and transfers reminder to affected user.


jorijnsmit (approved):
runs good here on local.

petrovska-petro (approved):
scripts looks good, events emitted for the transfer makes sense of the split given to the user and treasury amounts ✅

Issue #537, opened 2022-06-08
claim rewards from balancer


jorijnsmit:
deadline: june 9

jorijnsmit:
```
Events In This Transaction
--------------------------
├── Balancer 20WBTC-80BADGER Gauge Deposit (0xAF50825B010Ae4839Ac444f6c12D44b96819739B)
│   └── UpdateLiquidityLimit
│       ├── user: 0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e
│       ├── original_balance: 173042973718142955604702
│       ├── original_supply: 751468426440959664328050
│       ├── working_balance: 69217189487257182241880
│       └── working_supply: 507755212293505605332749
│   
├── IBalancerMinter (0x239e55F427D44C3cc793f49bFB507ebe76638a2b)
│   └── Minted
│       ├── recipient: 0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e
│       ├── gauge: 0xAF50825B010Ae4839Ac444f6c12D44b96819739B
│       └── minted: 12579255839514842233977
│   
├── Balancer (0xba100000625a3754423978a60c9317c58a424e3D)
│   └── Transfer
│       ├── from: 0x0000000000000000000000000000000000000000
│       ├── to: 0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e
│       └── value: 12579255839514842233977
│   
└── Gnosis Safe (0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e)
    └── ExecutionSuccess
        ├── txHash: 0x78d1b8881a9579fe2baa0e3b7adac7fee0a1d3454b389b78abc0a53b60d81421
        └── payment: 0
```
```
{
│   'ethereum_client': <gnosis.eth.ethereum_client.EthereumClient object at 0x1259b2f70>,
│   'safe_address': '0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e',
│   'to': '0x40A2aCCbd92BCA938b02010E17A5b8929b49130D',
│   'value': 0,
[...]

PR #538, opened 2022-06-08
script: claim balancer rewards for a given gauge

jorijnsmit:
issue #537 

@jalbrekt85 had to add a quick custom func to claim from a particular gauge; the rewards functions you are leveraging in `balancer.claim_all` (`reward_count`, `claimable_tokens`) somehow dont align with what is truly claimable.


petrovska-petro (commented):
scripts looks good

```
              balance_before              balance_after              balance_delta
symbol                                                                            
BAL     0.000000000000000000  12,280.837775300882862492  12,280.837775300882862492
```

impersonator_bal

Just 2 BAL of difference compare to UI. minor, but would be great to investigate. Script and SS were taken almost at the same timestamp. Using impersonatooooor do you see the same little dif?

Confirmed address [here](https://github.com/balancer-labs/balancer-v2-monorepo/blob/master/pkg/deployments/tasks/20220325-gauge-controller/output/mainnet.json#L5)

Minor comment to change on the interface added to move it to the right folder, i.e: `interfaces/balancer/IBalancerMinter.sol`, instead of `autogenerated/...`

jorijnsmit:
thanks, interface moved. that was still an artifact from a script i wrote which automatically generated the interface based on an address :)

in regards to the delta, i think that is just a timing issue; ie what block did you fork on and how close was that to the one balancer is using to calc in the ui?

petrovska-petro (approved):

petrovska-petro:
> 0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e



> thanks, interface moved. that was still an artifact from a script i wrote which automatically generated the interface based on an address :)
> 
> in regards to the delta, i think that is just a timing issue; ie what block did you fork on and how close was that to the one balancer is using to calc in the ui?

you're right was timing issue at my end. checking now again in parallel claimable bal in ui and snap is same 👍

Issue #536, opened 2022-06-08
sweep old digg system contracts


gosuto-inzasheru:
script as much as possible to sweep these two contracts, then delegate to dapp:

-  [x] `SmartVestingDigg.sol`; holding 18 digg: https://etherscan.io/address/0x124fd4a9bd4914b32c77c9ae51819b1181dbb3d4#tokentxns
- [ ] `StabilizeStrategyDiggV1.sol`; holding 8 digg: https://etherscan.io/address/0xa6af1b913e205b8e9b95d3b30768c0989e942316#tokentxns

jorijnsmit:
turns out `SmartVestingDigg.release()` is actually permissionless :): https://etherscan.io/tx/0x8a123136ba2b3cb677355def5c6b5ce0b16f7ff3637489562326bd361fc988ad

petrovska-petro:
```
snapshot result for 0x042B32Ac6b453485e357938bdC38e0340d4b9276:
                    balance_before          balance_after          balance_delta
symbol                                                                          
ETH          52.820980687743951977  54.820980687743951977   2.000000000000000000
DIGG          0.000000000000000000  26.049100188000000000  26.049100188000000000
WBTC         12.785857640000000000  13.283877320000000000   0.498019680000000000
renBTC        0.871819400000000000   1.055047200000000000   0.183227800000000000
imBTC         0.000000000000000000   0.830583609141352465   0.830583609141352465
fPmBTC/HBTC   0.000000000000000000   0.134068808178073103   0.134068808178073103


snapshot result for 0xB65cef03b9B89f99517643226d76e286ee999e77:
                    balance_before         balance_after           balance_delta
symbol                                                                          
ETH           2.000000000000000000  0.000000000000000000   -2.000000000000000000
DIGG         18.060031516000000000  0.000000000000000000  -18.060031516000000000
WBTC          0.353588430000000000  0.000000000000000000   -0.353588430000000000
renBTC        0.183227800000000000  0.000000000000000000   -0.183227800000000000
yvWBTC        0.119981100000000000  0.000000000000000000   -0.119981100000000000
imBTC         0.830583609141352465  0.000000000000000000   -0.830583609141352465
fPmBTC/HBTC   0.134068808178073103  0.000000000000000000   -0.134068808178073103 
```


```
Events In This Transaction
--------------------------
├── mBTC/HBTC Feeder Pool (0x48c59199Da51B7E30Ea200a74Ea07974e62C4bA7)
[...]

PR #539, opened 2022-06-08
draft: empty stabiliser digg strat

jorijnsmit:
@sajanrajdev can you let me know what you think of this solution?

problem is that the current controller (`0x9b4efA18c0c6b4822225b81D150f3518160f8609`) returns some weird test (?) bdigg vault (`0x608b6D82eb121F3e5C0baeeD32d81007B916E83C`) for digg...


jorijnsmit:
could be a partial solution for #536

sajanrajdev:
@jorijnsmit, I took a deep dive into this. First I wondered why didn't we just simply withdrew the amount with dev as it was dev who deposit it and holds the vault's shares and then I saw the issue....

The bDIGG vault in this controller has (unfortunately) the same name as the live DIGG vault because of the automatic name generator. But this vault was meant to have an stabilizing mechanism that buys/sells Digg for BTC according to its daily value to help stabilization. It was quite faulty and we never got around deploying it but it did got to production testing stage, hence this vault...

The reason we can't withdraw from it is that it, naturally, checks the want balance of the strategy when withdrawing and, in this case, `balanceOf` uses a chainlink feeds to get the DIGG equivalent of the wBTC in the strat and return this value. This Chainlink feed is broken and not returning any value, hence the balanceOf function reverts and breaks the withdrawal.

We could try to fix these feeds but I really don't see any point and seems like it would be a lot more time and resources consuming with no clear guarantee.

I think your approach is super clever! According to the snapshot it will recover all of the DIGG and wBTC sitting on the strat and from a Solidity side, I don't see any issues. The strat wire-up will be broken which will offcially deprecate these contracts.

Maybe we will want to deploy the dummy controller separately to keep things clean on the multi, not the biggest concern tho...

jorijnsmit:
yea deployment happens in one go in this draft, but i intend to deploy it seperately from an eoa indeed.

jorijnsmit:
todo: add a sweep of all tokens (wbtc, renbtc, eth, digg, yvwbtc, imbtc, mbtchbtc, remdigg?) to treasury ops.

would then depend on exec of #540

dapp-whisperer (dismissed):

petrovska-petro (commented):
Smart solution and nice trick on this [method](https://etherscan.io/address/0xEC4bc89F35A7bf9ffe0375Fd67940Ded1e2253BF#code#F1#L19) ✅

sajanrajdev (approved):
Changes LGTM, adds sweeping of assets from dev to trops

petrovska-petro (approved):

Issue #540, opened 2022-06-08
Redirect wd fee from byvWBTC to treasury


petrovska-petro:
Still wd fees are coming to dev_msig, [ref-tx](https://etherscan.io/tx/0x499df01f1c0466da10ac0776d612b58d70d7ec4c8cb0fa150bd69d97a795742c)

change with [method](https://etherscan.io/address/0xc0a972a3cf8e52167589694d135abb66409f9b30#code#L1449)

jorijnsmit:
i would suggest to redirect them to treasury ops; this is where all our other fees are currently going.

sajanrajdev:
Performance and management fees have already been redirected by changing the badger treasury address on Yearn's address book to the DAO's treasury: https://github.com/yearn/yearn-exporter/pull/339

As for withdrawal fees, the byvWBTC contract (the wrapper) is hardcoded to send these to the contract's `affiliate` (which is virtually the contract's governance. This address can't be changed to Trops as originally intended since it would also give this multi the governance rights which doesn't agree with our access control policy, see [comments](https://github.com/Badger-Finance/badger-multisig/pull/542#issuecomment-1152454236).

Instead, we need to upgrade the wrapper's logic to add a `badgerTreasury` function (settable by the affiliate) and modify the withdraw function so that it directs withdrawal fees in its direction. This address will be set to Treasury Ops.

The `byvWBTC`'s proxy admin is the devProxyAdmin so Timelock is required: 
![image](https://user-images.githubusercontent.com/25463788/173098778-af3c6214-724d-4086-91cc-40b867a9d07e.png)

Also, the wrapper's development happened on [badger-system](https://github.com/Badger-Finance/badger-system/blob/master/contracts/yearn/SimpleWrapperGatedUpgradeable.sol) and the logic was never extracted so this could be a good opportunity to do so.

@btcookies, this would also be a good opportunity to introduce GAC pausability to this vault.

CC @dapp-whisperer @petrovska-petro

btcookies:
Logic deployed: https://etherscan.io/address/0x021ea7548ee9b40d57f47706a605043b05c6c92c#code

btcookies:
Nonce 1055
Queued Trace

```
Transaction was Mined 
---------------------
Tx Hash: 0xfa074f65721b2e5f421c52f7be94bb211a6e4c90a35c776b7d386b3bb29a9723
From: 0x15b8Fe651C268cfb5b519cC7E98bd45C162313C2
[...]

PR #542, opened 2022-06-09
Migrate wd fee to trops in byvWBTC sett

petrovska-petro:
Tackles #540 

Commands to be run in order, setting from dev_msig and the acceptance from the new recipient for the wd fee:

```
brownie run issue/540/migrate_wd_fee set_affiliate

brownie run issue/540/migrate_wd_fee accept_affiliate
```

### DIFF
--- a/scripts/issue/540/migrate_wd_fee.py
+++ b/scripts/issue/540/migrate_wd_fee.py
@@ -1,0 +1,20 @@
+from great_ape_safe import GreatApeSafe
+from helpers.addresses import registry
+
+
+def set_affiliate():
+    dev_msig = GreatApeSafe(registry.eth.badger_wallets.dev_multisig)
+    byvWBTC = dev_msig.contract(registry.eth.yearn_vaults.byvWBTC)
+
+    byvWBTC.setAffiliate(registry.eth.badger_wallets.treasury_ops_multisig)
+
+    dev_msig.post_safe_tx()
+
+
+def accept_affiliate():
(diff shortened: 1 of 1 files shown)


sajanrajdev:
I would defer the last approval to @dapp-whisperer. For the Yearn wrapper, the `Affiliate` is basically the "governance" equivalent from our vaults. In this case, the affiliate has the power to change the guardian, manager, withdrawal fees. Imo, this last one is the only parameter that could be dangerous should the affiliate be compromised as WD are capped up to MAX_BPS (100%).

Other than this concern, I don't see a big issue.

Tritium-VLK (approved):

dapp-whisperer:
I agree with Saj take, we should create a new variable for the treasury, which receives all fees. Changing the affiliate is akin to setting governance to the treasury. We want a separation between contract governance and the treasury

sajanrajdev:
@petrovska-petro, `byvWBTC` proxy's is owned by the devProxyAdmin, upgrade should go through the Timelock:
image

Unfortunately, the dev of this wrapper was done on [badger-system](https://github.com/Badger-Finance/badger-system/blob/master/contracts/yearn/SimpleWrapperGatedUpgradeable.sol)... it would probably be an overkill to extract it to make this change so we should probably handle it there as we did with Digg.

dapp-whisperer:
So my thought on this is:
- modify the contract, adding a `treasury` var that receives fees. This variable can be set by `affiliate`
- run the upgrade to this new logic, and add treasury vault to receive fees

petrovska-petro:
> @petrovska-petro i guess this can be closed now..?

closing 🔐

Issue #389, opened 2022-04-21
refactor tests to use `RevertContextManager` opposed to `@pytest.mark.xfail`


jorijnsmit:
see https://eth-brownie.readthedocs.io/en/stable/api-test.html#revertcontextmanager

PR #553, opened 2022-06-14
refactor tests

jalbrekt85:
solves #389 and organizes protocol specific fixtures

### DIFF
--- a/tests/aave/conftest.py
+++ b/tests/aave/conftest.py
@@ -1,0 +1,23 @@
+import pytest
+from helpers.addresses import registry
+
+
+@pytest.fixture
+def aave(safe):
+    safe.init_aave()
+    return safe.aave
+
+
+@pytest.fixture
+def aUSDC(safe):
+    return safe.contract(registry.eth.treasury_tokens.aUSDC)
+
+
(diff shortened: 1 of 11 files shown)


jorijnsmit (commented):
had to move dai and cvxcrv fixtures to the general conftest since they are used across multiple platform tests.

also still getting quite a few fails/errors when running `brownie test`, but maybe we can fix that in a separate issue?

jalbrekt85:
> had to move dai and cvxcrv fixtures to the general conftest since they are used across multiple platform tests.
> 
> also still getting quite a few fails/errors when running `brownie test`, but maybe we can fix that in a separate issue?

fixtures have been moved. getting rest of tests to pass involving protocols like balancer or curve will require a decent amount of changes to great ape safe classes so i'd say it is better to handle in another ticket

jorijnsmit (approved):
i created #568 so we dont forget

petrovska-petro (approved):
conftest refactoring per protocol looks fine, just bunch of test are E or F. in light of #568 approving.

test ref from local run:

Screenshot 2022-06-22 at 15 05 37

Issue #554, opened 2022-06-14
Upgrade bveOXD/OXD LP Strat to fix harvests


sajanrajdev:
Upgrade the bveOXD/OXD strat to the logic with the changes introduced at: https://github.com/sajanrajdev/bveOXD-OXD-Staking-Optimizer/pull/2

New logic: https://ftmscan.com/address/0x5eb4e278d54ade50bdfb305c698a7640944bd7dc#code

Ensure that harvests are fixed upon harvest.

sajanrajdev:
Executed: https://ftmscan.com/tx/0x51ddf29aae8e2c59792cfad1803a43273a9dc10c71604a4f756dcde849ae257f

PR #555, opened 2022-06-14
Upgrade bveOXD/OXD script

sajanrajdev:
Closes issue #554 

Run with:
```
brownie run scripts/issue/554/upgrade_bveOXD_LP.py --netowork ftm-main-fork
```

Run simulation with:
```
brownie run scripts/issue/554/upgrade_bveOXD_LP.py main true --netowork ftm-main-fork
```

### DIFF
--- a/scripts/issue/554/upgrade_bveOXD_LP.py
+++ b/scripts/issue/554/upgrade_bveOXD_LP.py
@@ -13,7 +13,7 @@ NEW_LOGIC = registry.ftm.logic["StrategybveOxdOxdStakingOptimizer"]
 
 def main(simulation="false"):
     safe = GreatApeSafe(DEV_MULTI)
-    strat_contract = Contract(STRAT_ADDRESS)
+    strat_contract = Contract.from_explorer(STRAT_ADDRESS)
     strat = interface.IStrategy(STRAT_ADDRESS, owner=safe.account)
 
     # Record storage variables
@@ -37,11 +37,18 @@ def main(simulation="false"):
     # Confirm storage variables
     for attr in strat_contract.signatures:
         try:
-            assert attributes[attr] == getattr(strat_contract, attr).call()
-        # Should fail if assertion is wrong
+            # Balance of rewards is expected to increase as they are estimated based on time
+            if attr == "balanceOfRewards":
+                balance_of_rewards = getattr(strat_contract, attr).call()
+                assert balance_of_rewards[0][1] > attributes[attr][0][1]
+                assert balance_of_rewards[1][1] > attributes[attr][1][1]
+            else:
+                assert attributes[attr] == getattr(strat_contract, attr).call()
         except KeyError:
             C.print(f"[red]Error confirming {attr}[/red]")
             pass
+        except AssertionError:
+            C.print(f"[red]Assertion error with {attr}[/red]")
 
     # Test harvest
     if simulation == "true":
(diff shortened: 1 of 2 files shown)


scripts/issue/554/upgrade_bveOXD_LP.py
```diff
@@ -0,0 +1,48 @@
...
+            C.print(f"[red]Error storing {attr}[/red]")
+            pass
+
+    # Test failing harvest
+    if simulation == "true":
+        with reverts("BaseV1Router: INSUFFICIENT_B_AMOUNT"):
+            strat.harvest({"from": safe.account})
+
+    # Execute upgrade
+    proxyAdmin = interface.IProxyAdmin(DEV_PROXY, owner=safe.account)
```
btcookies, line 34: should snake_case for consistence

sajanrajdev: Good call, fixed

scripts/issue/554/upgrade_bveOXD_LP.py
```diff
@@ -0,0 +1,48 @@
...
+    proxyAdmin.upgrade(strat.address, NEW_LOGIC)
+
+    # Confirm storage variables
+    for attr in strat_contract.signatures:
```
btcookies, line 38: should we make this a function somewhere? like a generalized check_storage(contract)? and also think we should fail if one of the attributes isn't equal so maybe have something like following

```
is_storage_wrong = False
for attr in attrs:
   try:
      whatever
   except:
      is_storage_wrong = True

if is_storage_wrong:
    fail 
```

btcookies: Hmm should we at least log which ones pass and which ones fail so we can at least manually visually confirm everything looks good?

shuklaayush (approved):
LGTM

scripts/issue/554/upgrade_bveOXD_LP.py
shuklaayush: Btw, this isn't an exhaustive check since it only verifies public value-type storage variables and public view functions

sajanrajdev: That's right, it has limitations. Not sure if this approach could be generalized efficiently.

scripts/issue/554/upgrade_bveOXD_LP.py
```diff
@@ -38,7 +38,9 @@ def main(simulation="false"):
     for attr in strat_contract.signatures:
         try:
             assert attributes[attr] == getattr(strat_contract, attr).call()
-        except:
+        # Should fail if assertion is wrong
+        except KeyError:
```
petrovska-petro, line 42: catching only this exception for me locally seems to jump out in other type of exception `AssertionError`.

maybe should be instead `except (AssertionError, KeyError) as e:` or just check for *attr* which are in the *attributes* variable

sajanrajdev: Interesting, I didn't get any assertion errors. The idea was for assertion errors to jump out, they would mean that either a view variable's value changed and this change should be accounted or that a new variable was introduced. In this case, neither of these happened and I don't see why this would be the case. 

Do you know which attribute caused the assertion error?

petrovska-petro: for the record was **balanceOfRewards** discussed in dm

sajanrajdev:
Changed script to fetch contract from the explorer so that it retrives the latest instance (instead of the stored locally) with all the attributes. This showed the balanceOfRewards didn't match before and after running. This is due to rewards increasing slightly over the time difference. Added a check to ensure that these balances indeed only increase.

Additionally, catching for assertion errors was introduced.

CC @petrovska-petro

petrovska-petro (approved):
LGTM! ✅

shuklaayush (approved):

Issue #441, opened 2022-05-12
Fill up univ3 0.0000313-0.000063 range


petrovska-petro:
Deposit 40.82 wbtc in the range

![Screenshot 2022-05-12 at 11 02 04](https://user-images.githubusercontent.com/84875062/168033875-0510d765-ebab-4376-9265-58f19c8b0c75.png)

petrovska-petro:
- [ ] undog 35 ibbtc set, withdraw to wbtc and sent to trops

- [ ] trops mints the new univ3 nft

- [ ] sent nft to vault

petrovska-petro:
```
snapshot result for 0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e:
                         balance_before          balance_after           balance_delta
symbol                                                                                
bibbtc/sbtcCRV-f  74.303585245672336694  33.103585245672336694  -41.200000000000000000
WBTC               0.433451320000000000   0.660774960000000000    0.227323640000000000 
```

```
Events In This Transaction
--------------------------
├── Badger Sett Curve.fi Factory BTC Metapool: ibBTC (0xaE96fF08771a109dc6650a1BdCa62F2d558E40af)
│   └── Transfer
│       ├── from: 0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e
│       ├── to: 0x0000000000000000000000000000000000000000
│       └── value: 41200000000000000000
│   
├── VirtualBalanceRewardPool (0xAAf75a94394F6D06E01CcE62e2545CeFFBFA1E2D)
│   └── Withdrawn
│       ├── user: 0x6D4BA00Fd7BB73b5aa5b3D6180c6f1B0c89f70D1
│       └── amount: 41200000000000000000
│   
├── Curve.fi Factory BTC Metapool: ibBTC Convex Deposit (0xb43ed35B5A3a9544BBEd8039c67AB04AD428deEa)
│   └── Transfer
│       ├── from: 0x4F2b8a15d0Dd58c1eB60bd53e966872828519Cee
│       ├── to: 0x0000000000000000000000000000000000000000
│       └── value: 41200000000000000000
│   
├── Curve.fi ibbtc/sbtcCRV-f Gauge Deposit (0x346C7BB1A7a6A30c8e81c14e90FC2f0FBddc54d8)
│   └── UpdateLiquidityLimit
│       ├── user: 0x989AEb4d175e16225E39E87d0D97A3360524AD80
[...]

PR #556, opened 2022-06-14
Include script to deploy 0.0000313 - 0.000063 range univ3

petrovska-petro:
Tackles #441 

```
brownie run issue/441/mint_lowest_range_uni_v3
```

**Note:** somehow thought this one i pushed it. but searching for posting seems it is not the case

### DIFF
--- a/scripts/issue/441/mint_lowest_range_uni_v3.py
+++ b/scripts/issue/441/mint_lowest_range_uni_v3.py
@@ -1,0 +1,50 @@
+from great_ape_safe import GreatApeSafe
+from helpers.addresses import registry
+from brownie import Contract, interface
+
+wbtc = Contract(registry.eth.treasury_tokens.WBTC)
+
+RANGE_0 = 0.0000313
+RANGE_1 = 0.000063
+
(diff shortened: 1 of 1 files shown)


jalbrekt85 (approved):
lgtm

scripts/issue/441/mint_lowest_range_uni_v3.py
```diff
@@ -0,0 +1,50 @@
+from great_ape_safe import GreatApeSafe
+from helpers.addresses import registry
+from brownie import Contract, interface
+
+wbtc = Contract(registry.eth.treasury_tokens.WBTC)
+
+RANGE_0 = 0.0000313
+RANGE_1 = 0.000063
```
jorijnsmit, line 8: think this should be `.0000625`, but in the previous script we also deployed from `.000063` so better to go with that now.

https://github.com/Badger-Finance/badger-multisig/blob/a5e97213ac0746a7d32fa36fb848c8705ccd53db/scripts/issue/32/mint_lowest_range_uni_v3.py#L7

jorijnsmit (approved):

Issue #355, opened 2022-04-11
[Dev-multisig]: Wire-up new vaults to a Timelock governed Controller


sajanrajdev:
Some of the new vaults on Ethereum were released wired-up to individual Controllers governed by the DevMultisig. This was done in order to ease and accelerate their release while preserving a high level of security. It is now necessary to migrate these vaults to one of the three Timelock governed controllers, preferably the `native` controller:

```
        "native": "0x63cF44B2548e4493Fd099222A1eC79F3344D9682",
```

### Technical Details ###
The vaults to migrate are: bbveCVX-CVX-f, bcrvIbBTC, bcrvBadger. Their controllers are:
```
        "bbveCVX-CVX-f": "0x0c41A8613fbeFCC8d6e5dF1020DBb336F875247F",
        "ibBTCCrv": "0xe505F7C2FFcce7Ae4b076456BC02A70D8fe8d4d2",
        "badgerCrv_temp": "0xa54d8a596B4022CC4436b692C8ea0E342405eB6e"
```

Note: The vaults wired to the mStable and Restitution controllers shouldn't be migrated as they require a different and specific controller setup. 

Migration process:

Step 1:
Queue addition of vaults to native controller through the Timelock:
```
timelock.queueTx(native_controller.setVault(migrating_vaults))
timelock.queueTx(native_controller.setStrategy(migrating_strategies))
```

Step 2: 
Atomically, finish the wire-up (doesn't require Timelock since devMulti is the strategist on the Controller):
```
native_controller.setStrategy(migrating_stragies)

For all vaults/strats:
vault.setController(native_controller)
strategy.setController(native_controller)
```

Include simulations of a successful earn/harvest post migration.

sajanrajdev:
Take a look at the following script for reference: https://github.com/Badger-Finance/badger-ape/pull/512

sajanrajdev:
Tx queued events trace:
```
Transaction was Mined
---------------------
Tx Hash: 0x5336115b5f61ca2672b7944a80de9f124f8784374ef977b3d55bdc9f884a0073
From: 0x15b8Fe651C268cfb5b519cC7E98bd45C162313C2
To: 0xB65cef03b9B89f99517643226d76e286ee999e77
Value: 0
Function: Gnosis Safe.execTransaction
Block: 15013299
Gas Used: 230626 / 271807 (84.8%)

Events In This Transaction
--------------------------
[...]

PR #558, opened 2022-06-15
feat: set new strat and vaults controller to native

btcookies:
Addresses #355 

Command to test `brownie run scripts/issue/355/set_new_sett_controllers.py main true false true`
Command to queue `brownie run scripts/issue/355/set_new_sett_controllers.py main`
Command to execute `brownie run scripts/issue/355/set_new_sett_controllers.py main true false`


scripts/issue/355/set_new_sett_controllers.py
```diff
@@ -0,0 +1,109 @@
...
+
+# Set up safe
+safe = GreatApeSafe(DEV_MULTI)
+safe.init_badger()
+
+# Other constants
+TX_DIR = "data/badger/timelock/upgrade_new_sett_controllers/"
+
+
+def main(submitTx="true", queue="true", simulation="true"):
```
sajanrajdev, line 45: Imo, the simulation should be written in such a way that allows to review the behavior of the whole process sucessful execution. Aka, when running the simulation, the script should emulate the timelock txs execution (You can impersonate the timelock for this) and then ensure that, once executed, regular operations will work as expected - as you are doing here.

In this case, you can only simulate once the txs are queued and the timelock period executed. Also, it would be good is the safe doesn't post when running the simulation, it is just cleaner and safer.

You can refer to this upgrade script: https://github.com/Badger-Finance/badger-multisig/pull/523/files

Finally, adding the commands with the different combination of possible flags to the PR description would be very handy to guide reviewers around their intended behavior.

scripts/issue/355/set_new_sett_controllers.py
```diff
@@ -0,0 +1,109 @@
...
+            console.print(f"Approving strategy and setting vault for {key}")
+            safe.badger.queue_timelock(
+                target_addr=NATIVE_CONTROLLER,
+                signature="approveStrategy(address,address)",
+                data=encode_abi(
+                    ["address", "address"],
+                    [strategy_contract.want(), strategy_address],
+                ),
+                dump_dir=TX_DIR,
+                delay_in_days=6,
```
sajanrajdev, line 64: If posting on Thursday, an ETA of 4 days should be enough.

helpers/addresses.py
```diff
@@ -339,7 +339,7 @@
         "bbveCVX-CVX-f": "0x0c41A8613fbeFCC8d6e5dF1020DBb336F875247F",
         "ibBTCCrv": "0xe505F7C2FFcce7Ae4b076456BC02A70D8fe8d4d2",
         "restitution": "0x3F61344BA56df00dad9bBcA05d98CA2AeC43Ba0B",
-        "badgerCrv_temp": "0xa54d8a596B4022CC4436b692C8ea0E342405eB6e",
+        "bcrvBadger": "0xa54d8a596B4022CC4436b692C8ea0E342405eB6e"
```
sajanrajdev, line 342: missing a comma at the eol

sajanrajdev (approved):
Queues for each of the Setts in matter:
- approveStrategy
- setVault

Executes queues and calls:
- controller.setSrategy
- strategy.setController
- vault.setController

Includes a simulation for these actions and the successful execution of an earn and harvest post migraton.

Script LGTM

jorijnsmit (approved):
for the necessary strats:
- queue `approveStrategy` and `setVault` on native controller
- exec queued tx
- `setStrategy` on native controller
- `setController` on strat and respective vault

Issue #559, opened 2022-06-16
lock all of treasury's bal for aurabal


jorijnsmit:
- [x] claim pending bal and transfer to voter msig
- [ ] lock bal single sided for aurabal

jorijnsmit:
```
Events In This Transaction
--------------------------
├── Balancer 20WBTC-80BADGER Gauge Deposit (0xAF50825B010Ae4839Ac444f6c12D44b96819739B)
│   └── UpdateLiquidityLimit
│       ├── user: 0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e
│       ├── original_balance: 173042973718142955604702
│       ├── original_supply: 382924797665179334908908
│       ├── working_balance: 69217189487257182241880
│       └── working_supply: 199718872272774396986432
│   
├── IBalancerMinter (0x239e55F427D44C3cc793f49bFB507ebe76638a2b)
│   └── Minted
│       ├── recipient: 0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e
│       ├── gauge: 0xAF50825B010Ae4839Ac444f6c12D44b96819739B
│       └── minted: 17138491989473326148466
│   
├── Balancer (0xba100000625a3754423978a60c9317c58a424e3D)
│   ├── Transfer
│   │   ├── from: 0x0000000000000000000000000000000000000000
│   │   ├── to: 0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e
│   │   └── value: 3897699368326991160923
│   └── Transfer
│       ├── from: 0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e
│       ├── to: 0xA9ed98B5Fb8428d68664f3C5027c62A10d45826b
│       └── value: 17138307110849779627165
│   
└── Gnosis Safe (0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e)
    └── ExecutionSuccess
        ├── txHash: 0xb0899385b20e5467eb58e6305bb7cb50afdc99fcfb9e0424272659f202d7efd0
        └── payment: 0
```
```
{
│   'ethereum_client': <gnosis.eth.ethereum_client.EthereumClient object at 0x121d6ff70>,
│   'safe_address': '0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e',
│   'to': '0x40A2aCCbd92BCA938b02010E17A5b8929b49130D',
│   'value': 0,
[...]

PR #560, opened 2022-06-16
script: lock bal for aurabal

jorijnsmit:
for #559; one-sided lock bal for aurabal


petrovska-petro (approved):
- Addresses added into book compared against [ref](https://docs.aura.finance/developers/deployed-addresses) ✅ 

- locks bal smoothly thru wrapper and prints balance of the voting_msig ✅

Issue #563, opened 2022-06-16
week 24: guarded emissions schedule for graviAURA


mitche50:
Per council decision set the below emissions schedule for graviAURA - dependent on adding 50/50 emissions rate (https://github.com/Badger-Finance/badger-multisig/issues/562)

```
beneficiary: graviAura (0xBA485b556399123261a5F9c95d413B4f93107407)
token: BADGER (0x3472A5A71965499acd81997a54BBA8D852C6E53d)
amount: 25000000000000000000000 (25k)
start: 1655481600 (tomorrow noon ET)
end: 1658505600 (5 weeks = July 22)
duration: 3024000
```

petrovska-petro:
going down for 2k/week in guard state first. starting today 22:00 utc and lasting the schedule for a week

```
Events In This Transaction
--------------------------
├── AdminUpgradeabilityProxy (0x0A4F4e92C3334821EbB523324D09E321a6B0d8ec)
│   └── UnlockScheduleSet
│       ├── beneficiary: 0xBA485b556399123261a5F9c95d413B4f93107407
│       ├── token: 0x3472A5A71965499acd81997a54BBA8D852C6E53d
│       ├── totalAmount: 2000000000000000000000
│       ├── start: 1655503200
│       ├── end: 1656108000
│       ├── duration: 604800
│       ├── timestamp: 14979976
│       └── blockNumber: 1655480545
│   
└── Gnosis Safe (0x86cbD0ce0c087b482782c181dA8d191De18C8275)
    └── ExecutionSuccess
        ├── txHash: 0x5b04c14ddff9c80522056c2872de83fb2b8e1ca64ebd1c02e1d00f9ab567b6de
        └── payment: 0
```


```
{
│   'ethereum_client': <gnosis.eth.ethereum_client.EthereumClient object at 0x7ff76e606400>,
│   'safe_address': '0x86cbD0ce0c087b482782c181dA8d191De18C8275',
│   'to': '0x40A2aCCbd92BCA938b02010E17A5b8929b49130D',
│   'value': 0,
[...]

Issue #562, opened 2022-06-16
Add 50/50 Emissions Rate for graviAURA on Emissions Control contract


mitche50:
Emissions for graviAURA should be split 50/50 boosted and prorata.  

EmissionControl contract: 0x31825c0a6278b89338970e3eb979b05b27faa263

Target Vault: [Gravitationally Bound AURA (graviAURA)](https://etherscan.io/token/0xBA485b556399123261a5F9c95d413B4f93107407)

Contract call: `setBoostedEmission(graviAura, 5000)`

petrovska-petro:
```
Events In This Transaction
--------------------------
├── IEmissionControl (0x31825c0A6278b89338970e3eB979b05B27FAa263)
│   └── TokenBoostedEmissionChanged
│       ├── _vault: 0xBA485b556399123261a5F9c95d413B4f93107407
│       └── _weight: 5000
│   
└── Gnosis Safe (0x86cbD0ce0c087b482782c181dA8d191De18C8275)
    └── ExecutionSuccess
        ├── txHash: 0x6286870675f3335b79bb87bea48e559b0333d079d74d87790274b1809d786930
        └── payment: 0
```


```
{
│   'ethereum_client': <gnosis.eth.ethereum_client.EthereumClient object at 0x7ffacc3b2b80>,
│   'safe_address': '0x86cbD0ce0c087b482782c181dA8d191De18C8275',
│   'to': '0x40A2aCCbd92BCA938b02010E17A5b8929b49130D',
│   'value': 0,
│   'data': HexBytes('0x8d80ff0a000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000990031825c0a6278b89338970e3eb979b05b27faa263000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000445326be32000000000000000000000000ba485b556399123261a5f9c95d413b4f93107407000000000000000000000000000000000000000000000000000000000000138800000000000000'),
│   'operation': 1,
│   'safe_tx_gas': 119151,
│   'base_gas': 0,
│   'gas_price': 0,
│   'gas_token': '0x0000000000000000000000000000000000000000',
│   'refund_receiver': '0x0000000000000000000000000000000000000000',
│   'signatures': b'',
│   '_safe_nonce': 278,
│   '_safe_version': '1.1.1',
│   '_chain_id': None
}
```

petrovska-petro:
[...]

PR #564, opened 2022-06-17
Set boost rate and emission schedules for graviAURA

petrovska-petro:
Tackles #563 & #562

```
brownie run issue/562/rate_emissions_graviaura

brownie run issue/563/emissions_graviaura
```

**Note:** will like to post them individually for batch testing thru the ui (posting first 562 and then 563)


jorijnsmit (dismissed):
looks good, address confirmed and timestamps doublechecked.

would like @mitche50 to give final green light

petrovska-petro:
pushed another commit to allow setting args for the emissions, since likely it will not be 5k/week linear over 5 weeks, since there'd be guard state with different amounts of emissions than when it is on full open state

scripts/issue/563/emissions_graviaura.py
```diff
@@ -0,0 +1,21 @@
+from great_ape_safe import GreatApeSafe
+from helpers.addresses import registry
+
+
+def main(amount=25000000000000000000000, start_ts=1655481600, end_ts=1658505600):
```
mitche50, line 5: We'll need to adjust this to `amount=2000000000000000000000`, `end_ts=1656086400`

sajanrajdev: These parameters can be passed to the script when running it, although it won't hurt to have the default being the correct ones.

mitche50 (approved):
LGTM - params will be passed per Saj's comments

Issue #502, opened 2022-05-26
quarter 3: native topups from vault to drippers and ops


jorijnsmit:
- [x] tree needs last bit of deficit of digg after all digg emissions have been stopped (#479): `~3.7 digg`
- [ ] tree dripper q3 needs badger for emissions: `assuming 27k badger * 13 weeks = 260k badger`
- [ ] rembadger dripper q3 needs badger for deposits: `5769.230769 * 13 weeks ~= 75k badger`
- [ ] treasury ops needs badger for bribes plus maybe a random float?: `8k * 13 weeks = 104k badger`
- [ ] fin ops needs badger for contributors: `500k badger`
- [ ] fin ops needs usdc for contributors: `1.2m usdc`

jorijnsmit:
digg to tree to solve deficit: https://etherscan.io/tx/0xa80885ab4130addceab8448c4e700fffc25cf0c0d91ed269822f16f5e4ed9b36

jorijnsmit:
waiting for badgerwbtc bpt migration to aura before running this; the script also claims bal which would then be the final claim.

jorijnsmit:
queued in timelock: https://etherscan.io/tx/0xf8f7413c24ec0509753a6fd63707ff5dd4809fd72309fb515bd118726fedfe87

jorijnsmit:
```
snapshot result for 0x042B32Ac6b453485e357938bdC38e0340d4b9276:
                           balance_before               balance_after               balance_delta
symbol                                                                                           
BADGER           1,691.980497086993488895  105,691.980497086993488895  104,000.000000000000000000
USDC             5,021.732304000000000000   49,546.146392000000000000   44,524.414088000000000000
WBTC                 7.247350540000000000       17.761158160000000000       10.513807620000000000
USDT                 0.000000000000000000   18,613.600758000000000000   18,613.600758000000000000
DAI              1,566.385266087288080979   25,632.762288256664098274   24,066.377022169376017295
EURS             1,202.930000000000000000    6,971.830000000000000000    5,768.900000000000000000
CRV                  0.000000000000000000    5,967.790918573782540416    5,967.790918573782540416
CVX                  0.995121610584372088      570.211122420524023434      569.216000809939651346
FEI                  0.000000000000000000    2,396.212960120952025081    2,396.212960120952025081
ANGLE            1,021.808396951618359590   10,694.812495826891359778    9,673.004098875273000188
[...]

PR #567, opened 2022-06-20
quarterly top up, finalise stable farm rebalance and sweep vault dust 

jorijnsmit:
- quarterly top up (#502)
- deposit and stake bpt_3pool (last action in #413)
- also a general clean up of lots of dust in the vault, which is sent to trops

the preview doesnt like `safe.balancer.deposit_and_stake` hence `skip_preview=True`


scripts/issue/502/top_ups_for_q3.py
```diff
@@ -0,0 +1,86 @@
...
+    # stake ~$7k worth of dust bpt
+    safe.init_balancer()
+    safe.balancer.stake_all(bpt_badgerwbtc)
+
+    # balancer 3pool deposit & stake;
+    # https://github.com/Badger-Finance/badger-multisig/issues/413
+    underlyings = [usdt, usdc, dai]
+    amounts = [int(102_457e6), int(102_457e6), int(102_457e18)]
+    safe.balancer.deposit_and_stake(underlyings, amounts)
+    safe.balancer.stake_all(bpt_3pool) * .995
```
petrovska-petro, line 59:
```suggestion
    safe.balancer.stake_all(bpt_3pool, dusty=.995)
```

scripts/issue/502/top_ups_for_q3.py
```diff
@@ -0,0 +1,86 @@
...
+        badger, usdc, weth, bal, usdt, dai, eurs, crv, cvx, fei, angle,
+        threepool, ibbtc_lp, slp, bpt_badgerwbtc, bpt_3pool, bvecvx
+    ])
+    trops.take_snapshot([
+        badger, usdc, weth, bal, usdt, dai, eurs, crv, cvx, fei, angle,
+        threepool, ibbtc_lp, slp, bpt_badgerwbtc, bpt_3pool, bvecvx
+    ])
+
+    # badger for emissions, rembadger, bribes and contribs q3
+    badger.transfer(r.drippers.tree_2022_q2, 260_000e18)
```
petrovska-petro, line 37: maybe for the drippers already use dummy placers as `r.drippers.tree_2022_q3` & `r.drippers.rembadger_2022_q3` even if using same address as q2, so if it gets merged before deploying the q3 ones, then it just matter of updating the address book or worth deploying them already and include them in this pr

jorijnsmit: pushed in eb1249f

petrovska-petro (approved):
LGTM!

For other reviewers, figures deducted from [L40](https://github.com/Badger-Finance/badger-multisig/blob/issue/502/scripts/issue/502/top_ups_for_q3.py#L40) & [L43](https://github.com/Badger-Finance/badger-multisig/blob/issue/502/scripts/issue/502/top_ups_for_q3.py#L43) are coming from the fact of funds in this [msig](https://github.com/Badger-Finance/badger-multisig/blob/main/helpers/addresses.py#L44) to be transferred to the *payments_multisig* (mainly all the badger funds).

jalbrekt85 (approved):
lgtm. agree that it is better to just finish the balancer stable deposit at this point.

PR #572, opened 2022-06-20
feat: basic lever_up script for treasury

GalloDaSballo:
WIP

- [X] Deposit into AAVE
- [X] Go Long on X token by one Loop (borrow X / 10_000, buy long token, deposit)
- [] Delever
- [] Multichain
- [] Claim Extra Rewards

### DIFF
--- a/scripts/treasury/aave_leverage.py
+++ b/scripts/treasury/aave_leverage.py
@@ -1,0 +1,88 @@
+"""
+  Given Token In and OUT, Leverage functions
+"""
+import requests
+import json
+from great_ape_safe import GreatApeSafe
+from brownie import Contract
+from helpers.addresses import registry
+
+multisigAddress = registry.eth.badger_wallets.dev_multisig
+
+## Signer
+safe = GreatApeSafe(multisigAddress)
+
+## Mainnet
+AAVE_LENDING_POOL = safe.contract("0x7d2768dE32b0b80b7a3454c06BdAc94A69DDc7A9")
+
+WBTC = safe.contract("0x2260FAC5E5542a773Aa44fBCfeDf7C193bc2C599")
+
+USDC = safe.contract("0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48")
+
+DEPOSIT_TOKEN = WBTC ## We deposit wBTC and Buy it
+BORROW_TOKEN = USDC ## We borrow USDC and use it to buy wBTC
+
+## Returns price in ETH
+PRICE_ORACLE = safe.contract("0xA50ba011c48153De246E5192C8f9258A2ba79Ca9")
+
+## UniV2 Router
+UNIV2_ROUTER = safe.contract("0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D")
+
+
+
+def deposit(amount):
+  """
+    Deposit into AAVE, earns basic interest, LM incentives and allows to lever
+  """
+  WBTC.approve(AAVE_LENDING_POOL, amount)
+  AAVE_LENDING_POOL.deposit(WBTC, amount, safe.address, 0)
+
+
+
+def lever_up(percent_bps):
+  """
+    Given total borrow amounts, we will use `percent_bps` of it to borrow BORROW_TOKEN
+  """
+  ## Get total borrowable
+  info = AAVE_LENDING_POOL.getUserAccountData(safe.address)
+  print("info")
+  print(info)
+
+  available_borrows_eth = info[2]
+  print("available_borrows_eth")
+  print(available_borrows_eth)
+
+  ## From ETH to USDC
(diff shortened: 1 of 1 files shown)


GalloDaSballo:
@jorijnsmit @Hal69k See above for as basic as it get's leverage long script (also allows leverage short, just swap tokens)

GalloDaSballo:
We have a better pricer as well from the [`fair-selling repo`](https://github.com/GalloDaSballo/fair-selling) however sale integration has yet to be finished, but can use that once it's ready

GalloDaSballo:
Added function to repay debt, will work on delever (repay with available funds) tomorrow

scripts/treasury/aave_leverage.py
```diff
@@ -0,0 +1,134 @@
...
+def delever(debt_to_repay):
+  """
+    NOTE: Incomplete
+  """
+  ## Given amount of debt_to_repay, withdraw what is possible and repay
+  info = AAVE_LENDING_POOL.getUserAccountData(safe.address)
+  collateral_in_eth = info[0]
+  debt_in_eth = info[1]
+
+  usdc_to_eth = PRICE_ORACLE.getAssetPrice(BORROW_TOKEN)
```
rayeaster, line 106: Aave use chainlink feeds as oracle by default and fallback to some home-grown `blackbox` oracle on some conditions(https://docs.aave.com/developers/v/2.0/the-core-protocol/price-oracle).

Considering the potential risk exposed by the fallback oracle like this report https://medium.com/@hacxyk/aave-v3s-price-oracle-manipulation-vulnerability-168e44e9e374, we might use chainlink directly and/or other off-chain price feed?

scripts/treasury/aave_leverage.py
```diff
@@ -0,0 +1,134 @@
...
+
+  ## Calculate USDC we can borrow
+  usdc_we_can_borrow = ((available_borrows_eth * 10**(BORROW_TOKEN.decimals())) // usdc_to_eth) * percent_bps // 10_000
+
+  print("usdc_we_can_borrow")
+  print(usdc_we_can_borrow)
+
+  ## Borrow USDC
+  AAVE_LENDING_POOL.borrow(BORROW_TOKEN, usdc_we_can_borrow, 2, 0, safe.address)
```
rayeaster, line 72: some check on `usdc_we_can_borrow`? like `if(usdc_we_can_borrow <= 100 * 10**6): return`

scripts/treasury/aave_leverage.py
```diff
@@ -0,0 +1,134 @@
...
+
+  ## Sell USDC for more wBTC
+  BORROW_TOKEN.approve(UNIV2_ROUTER, usdc_we_can_borrow)
+  swap_result = UNIV2_ROUTER.swapExactTokensForTokens(
+        usdc_we_can_borrow,
+        0, ## NOTE: We could use oracle here
```
rayeaster, line 78: agree. If no minimum output is specified here and we don't use sth like flashbot/eden to protect the script execution, this swap might fall victim of mev bots.

scripts/treasury/aave_leverage.py
```diff
@@ -0,0 +1,134 @@
...
+
+  available_borrows_eth = info[2]
+  print("available_borrows_eth")
+  print(available_borrows_eth)
+
+  ## From ETH to USDC
+  usdc_to_eth = PRICE_ORACLE.getAssetPrice(BORROW_TOKEN)
+
+  ## Calculate USDC we can borrow
+  usdc_we_can_borrow = ((available_borrows_eth * 10**(BORROW_TOKEN.decimals())) // usdc_to_eth) * percent_bps // 10_000
```
rayeaster, line 66: do we need to set an upper limit for `percent_bps` and apply some check before borrow it from Aave? say, never exceed the `70%` of the maximum available to borrow, i.e., `percent_bps < 7_000`?

scripts/treasury/aave_leverage.py
jorijnsmit: i agree, using the direct btcusd chainlink oracle here makes more sense. even if it is just to get rid of the unnecessary eth hop.

GalloDaSballo:
@jorijnsmit @Hal69k wdyt?

scripts/treasury/aave_leverage.py
jorijnsmit: this is missing a division by the ltv, since `available_borrows_eth` is already reduced by the ltv. in this state 1btc levered by 5000bps will result in 1.35btc, since the ltv is 70%.

eg if we want to add a lever of .5x with a ltv of .7, we need `borrowable * .5 / .7`

jorijnsmit:
im refactoring the logic into great ape safe, i think that will serve us well in the future. trying to incorporate as much of the comments left here as well.

gosuto-inzasheru:
closing in favour of #625

PR #576, opened 2022-06-21
chore: add cookies deployer to addresses

btcookies:


petrovska-petro (approved):
confirmed other deployer funding it and logic contract deployment history ✅

PR #579, opened 2022-06-22
chore: add dripper addresses

jorijnsmit:
https://github.com/Badger-Finance/badger-infra/issues/4
https://github.com/Badger-Finance/badger-infra/issues/5


petrovska-petro (approved):
confirmed beneficiaries, starting timestamp, duration, interval, watchlist asset (only badger), owner (techops) and keeper registry address ✅

jalbrekt85 (approved):

Issue #581, opened 2022-06-22
Claim rewards from auraBAL


petrovska-petro:
- [ ] claim aura rewards 

- [ ] perhaps redirect vp to delegate addr

petrovska-petro:
```
vents In This Transaction
--------------------------
├── Aura (0xC0c293ce456fF0ED870ADd98a0828Dd4d2903DBF)
│   ├── Approval
│   │   ├── owner: 0xC47162863a12227E5c3B0860715F9cF721651C0c
│   │   ├── spender: 0x3Fa73f1E5d8A792C80F426fc8F84FBF7Ce9bBCAC
│   │   └── value: 11328161565441513667081
│   ├── Transfer
│   │   ├── from: 0xC47162863a12227E5c3B0860715F9cF721651C0c
│   │   ├── to: 0x3Fa73f1E5d8A792C80F426fc8F84FBF7Ce9bBCAC
│   │   └── value: 11328161565441513667081
│   └── Approval
│       ├── owner: 0xC47162863a12227E5c3B0860715F9cF721651C0c
│       ├── spender: 0x3Fa73f1E5d8A792C80F426fc8F84FBF7Ce9bBCAC
│       └── value: 0
│   
├── Vote Locked Aura (0x3Fa73f1E5d8A792C80F426fc8F84FBF7Ce9bBCAC)
│   └── Staked
│       ├── _user: 0xA9ed98B5Fb8428d68664f3C5027c62A10d45826b
│       ├── _paidAmount: 11328161565441513667081
│       └── _lockedAmount: 11328161565441513667081
│   
├── AuraBalRewardPool (0xC47162863a12227E5c3B0860715F9cF721651C0c)
│   └── RewardPaid
│       ├── user: 0xA9ed98B5Fb8428d68664f3C5027c62A10d45826b
│       ├── reward: 11328161565441513667081
│       └── locked: True
│   
├── Vote Locked Aura (0x3Fa73f1E5d8A792C80F426fc8F84FBF7Ce9bBCAC)
│   ├── DelegateChanged
│   │   ├── delegator: 0xA9ed98B5Fb8428d68664f3C5027c62A10d45826b
│   │   ├── fromDelegate: 0x0000000000000000000000000000000000000000
│   │   └── toDelegate: 0x14F83fF95D4Ec5E8812DDf42DA1232b0ba1015e6
│   └── DelegateCheckpointed
[...]

PR #583, opened 2022-06-22
Claim aura rewards and delegate

petrovska-petro:
Tackles #581 

```
brownie run issue/581/claim_aurbal_rewards
```


jalbrekt85 (approved):
tx trace looks good. might make sense to have an Aura class in the future that makes sure claiming, delegating, etc works properly

Issue #580, opened 2022-06-22
Seed initial liquidity into first graviAURA gauge pools so they start trading


Tritium-VLK:
Balancer recommends at least $20k USD in a pool for it to trade. I think that will work in the 33/33/33 pools, maybe a little more would be good.  We may need a bit more in the badger/graviaura pool as it's an 80/20.  In the end there is enough TCL when we start seeing swaps/arbitrage.

The pools have already been created and are defined in this document: https://docs.google.com/document/d/1fi1_MmLoArubvmulIVZRK08Eo4HPNMOEuvzRhQmtDCE/ as

[AURA/ETH/graviAURA (33/33/33)](https://app.balancer.fi/#/pool/0xa3283e3470d3cd1f18c074e3f2d3965f6d62fff2000100000000000000000267)
To support AURA liquidity during its 
[auraBAL/ETH/graviAURA (33/33/33)](https://app.balancer.fi/#/pool/0x0578292cb20a443ba1cde459c985ce14ca2bdee5000100000000000000000269)
To support both auraBAL and graviAURA liquidity and utility
[BADGER/WBTC/graviAURA (40/40/20)](https://app.balancer.fi/#/pool/0xa0b7be8336eaa4ba1c7bbc23a694de3c0ff83a1000010000000000000000026f)
To drive value to BADGER

Will start the gauge vote today, reference this public ticket, and state that initial liqudity should be in place by next Wednesday or so in terms of planning snapshots.

Tritium-VLK:
sent initial BPT tokens for all pools and some other dust I collected building these to treasury-ops, from tritium-vlk.eth

jorijnsmit:
necessary liq at current coingecko prices:
```
[3944773175542406447104, 5546173279091759104, 3944773175542406447104]
[3944773175542406447104, 5546173279091759104, 1257861635220125777920]
[2366863905325443973120, 4833836858006042574848]
needed aura: 15779092702169625788416 15779.092702169626
needed weth: 11092346558183518208 11.092346558183518
needed bal: 1257861635220125777920 1257.861635220126
needed badger: 4833836858006042574848 4833.836858006042
```

Tritium-VLK:
Note that due to feedback on from our governance, we have changed the badger pool to BADGER/WBTC/graviAURA 40/40/20.  I have updated the summary.  Here is the new pool:
[...]

PR #600, opened 2022-06-27
bootstrap new bpt pools with some liquidity

jorijnsmit:
issue #580. only doing the first two bpts, the badger bpt still has not been decided on.

bit tricky this one, please review carefully. also @jalbrekt85 can you confirm `pool_type=Stable` is a good choice here? i think it only affects the min bpt out calc, but not sure. it is weird to me that not roughly the same amount of bpt tokens are returned for both pool.


jalbrekt85:
>  name for stables seems to always contain _XXXX_ **Stable Pool**

using the name is how I differentiated before I started using the API for getting the pool type, but this is pretty inconsistent because pool deployers don't have to follow the naming conventions

>  stable pool does not have `getNormalizedWeights()` method

i think this would work pretty well. the only point for grabbing the pool type in the first place is for the pool math and the main difference between stable/weighted pools is whether or not it uses the `getNormalizedWeights` or `getAmplificationParameter`

PR #614, opened 2022-06-30
fix: stake bpt in aura instead of balancer

jorijnsmit:
tiny adjustment made in the top up script to prevent staking in balancer and use the aura gauge instead.


petrovska-petro (approved):

sajanrajdev (approved):
- Modified BADGER transfer amount to the Tree dripper to match the required amounts.
- Modified bal3Pool handling to stake on Aura instead of Balancer
- Modified printouts to reflect new staking flow for the bal3Pool

LGTM

jorijnsmit:
linking #502

PR #636, opened 2022-07-05
uni v3 swap method

jalbrekt85:
initial attempt at `uni_v3.swap`, having problems encoding `path` for `router.quoteExactInput` & `router.exactInput` for use in multi hop swaps


great_ape_safe/ape_api/uni_v3.py
```diff
@@ -399,3 +433,39 @@ def transfer_nft(self, token_id, new_owner):
...
+        for item in multihop_path:
+            if web3.isAddress(item):
+                path_encoded += web3.toBytes(hexstr=item)
+            else:
+                path_encoded += int.to_bytes(item, 3, byteorder="big")
+
+        min_out = self.quoter.quoteExactInput(
+            path_encoded,
+            mantissa,
+        ).return_value * (1 - self.slippage)
```
gosuto-inzasheru, line 457: the quote function is actually (weirdly) not defined as a `view`, resulting in brownie performing an actual tx for it if we dont specify `.call`:
```suggestion
        min_out = self.quoter.quoteExactInput.call(
            path_encoded,
            mantissa,
        ) * (1 - self.slippage)
```

great_ape_safe/ape_api/uni_v3.py
```diff
@@ -3,7 +3,10 @@
 from datetime import datetime
 import math
 from pathlib import Path
-from brownie import interface, chain, multicall
+import eth_abi
```
gosuto-inzasheru, line 6:
```suggestion
```
unused

gosuto-inzasheru (changes requested):

gosuto-inzasheru:
working good after these minor changes, at least from what i could see in console testing.

maybe also implement the `exactOutput` func while you are at it? maybe in the same `swap` wrapper by changing params to `mantissa_in=None` and `mantissa_out=None`?

interfaces/uniswap/IQuoter.sol
```diff
@@ -0,0 +1,44 @@
+// SPDX-License-Identifier: UNLICENSED
+// !! THIS FILE WAS AUTOGENERATED BY abi-to-sol v0.6.1. SEE SOURCE BELOW. !!
```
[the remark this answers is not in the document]
jalbrekt85: wasn't aware, very nice

great_ape_safe/ape_api/uni_v3.py
```diff
@@ -48,6 +57,31 @@ def _get_pool(self, position):
...
+            
+            if list(fee_tiers.values()).count(0) == 3:
+                raise Exception(
+                    f"No liquidity found for {path[i].symbol()} - {path[i+1].symbol()}")
+
+            best_tier = max(fee_tiers, key=fee_tiers.get)
+            multihop.append(best_tier)
+            multihop.append(path[i+1].address)
+
+        return multihop
```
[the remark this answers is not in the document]
jalbrekt85: works with any number of tokens in path from what ive seen. the loop iterates over `len(path) - 1` but appends `path[i+1]`, so the last token is always added

great_ape_safe/ape_api/uni_v3.py
jalbrekt85: good call

gosuto-inzasheru (approved):

petrovska-petro (approved):
V neat, tried locally with diff hop amounts and checked against ui the tokens out and amounts seems to align properly ✅ 

@jalbrekt85 wondering for this case that we are leveraging the uniswap packages if we could use directly the interfaces from there in the script, similar like [here](https://eth-brownie.readthedocs.io/en/stable/package-manager.html?highlight=pm#using-packages-in-tests). will remove the need to needing to add the interfaces again, as in those package we added in the dependencies should be available all interfaces already.

PR #644, opened 2022-07-07
Compound leveraging

rayeaster:


gosuto-inzasheru:
closing for now

can alway revisit when the need arises

PR #653, opened 2022-07-08
feat: add wrappers for the different types of claims

gosuto-inzasheru:


great_ape_safe/ape_api/aura.py
```diff
@@ -27,13 +23,23 @@ def __init__(self, safe):
     def _alert_not_relevant():
         print(" ==== NOT RELEVANT METHOD IN AURA CLASS ==== \n")
 
+
+    def claim_all_from_booster(self):
+        self.claim_all(1)
```
petrovska-petro, line 28: i like this approach, gives a bit more of intuition. but would say here not to pass the 1, as it is the default value
```suggestion
        self.claim_all()
```

gosuto-inzasheru: making it explicit is better, again this increases readability for the user.

petrovska-petro: sounds good!

Issue #662, opened 2022-07-13
Voter msig auto-delegate and vote with the stack


petrovska-petro:
The aura claimed from the airdrop was locked and [delegate.badgerdao.eth](https://etherscan.io/address/0x14f83ff95d4ec5e8812ddf42da1232b0ba1015e6). Delegate to our own-self for being able to redirect its VP for future voting rounds

- [x] self-delegate

- [x] vote with stack

petrovska-petro:
```
Events In This Transaction
--------------------------
├── Vote Locked Aura (0x3Fa73f1E5d8A792C80F426fc8F84FBF7Ce9bBCAC)
│   ├── DelegateChanged
│   │   ├── delegator: 0xA9ed98B5Fb8428d68664f3C5027c62A10d45826b
│   │   ├── fromDelegate: 0x14F83fF95D4Ec5E8812DDf42DA1232b0ba1015e6
│   │   └── toDelegate: 0xA9ed98B5Fb8428d68664f3C5027c62A10d45826b
│   ├── DelegateCheckpointed
│   │   └── delegate: 0x14F83fF95D4Ec5E8812DDf42DA1232b0ba1015e6
│   └── DelegateCheckpointed
│       └── delegate: 0xA9ed98B5Fb8428d68664f3C5027c62A10d45826b
│   
└── Gnosis Safe (0xA9ed98B5Fb8428d68664f3C5027c62A10d45826b)
    └── ExecutionSuccess
        ├── txHash: 0x677b6b6cd3acf6e4b11471c8bdf2a5e77bfa700ea853eed97e0c645e99d1b248
        └── payment: 0
```

```
{
│   'ethereum_client': <gnosis.eth.ethereum_client.EthereumClient object at 0x7fac62f6a130>,
│   'safe_address': '0xA9ed98B5Fb8428d68664f3C5027c62A10d45826b',
│   'to': '0x40A2aCCbd92BCA938b02010E17A5b8929b49130D',
│   'value': 0,
│   'data': HexBytes('0x8d80ff0a00000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000079003fa73f1e5d8a792c80f426fc8f84fbf7ce9bbcac000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000245c19a95c000000000000000000000000a9ed98b5fb8428d68664f3c5027c62a10d45826b00000000000000'),
[...]

PR #666, opened 2022-07-15
Delegate aura lock in voter to itself

petrovska-petro:
Tackles #662 

```
brownie run issue/662/voter_delegate_itself
```


gosuto-inzasheru (approved):
work of art

do we already know how to handle the voting for the msig directly?

jalbrekt85 (approved):
lgtm

PR #668, opened 2022-07-15
Cast vote in voter_msig in aura space

petrovska-petro:
Tackles #662 casting voting piece.

```
brownie run snapshot/AURA/vote_proposal {proposa_id}
```

**Note:** at the moment the relayer endpoint seems to be on maintenance

### DIFF
--- a/scripts/snapshot/AURA/vote_proposal.py
+++ b/scripts/snapshot/AURA/vote_proposal.py
@@ -1,0 +1,99 @@
+import requests
+import json
+import time
+
+from eth_account import messages
+from brownie import interface
+
+from great_ape_safe import GreatApeSafe
+from helpers.addresses import r
+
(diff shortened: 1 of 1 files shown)


gosuto-inzasheru (approved):
tough to test this.

i know there is a test env on rinkeby i think: https://demo.snapshot.org/#/, but we might as well test in prod.

if this proves to work correctly, i would suggest moving this to great ape safe, so that in the future we can just `safe.snapshot.vote(prop_id)` (or i guess `safe_tx = safe.snapshot.vote(prop_id)` or something).

jalbrekt85 (approved):
the process of posting voting data seems to be same as #465, which has been tested on live msigs, lgtm. agree with @gosuto-inzasheru that this would be a good script to abstract into great ape safe

Issue #677, opened 2022-07-19
Add `Snapshot` to great ape safe


jalbrekt85:
should handle voting through snapshot api

PR #685, opened 2022-07-26
Snapshot class, method to vote

jalbrekt85:
handles #677, still more to explore with api. drop suggestion/feature request here
idea is that vote can be posted as such through scripts:
```py
voter = GreatApeSafe(r.badger_wallets.treasury_voter_multisig)
safe.init_snapshot(proposal_id)
safe.snapshot.vote("80/20 BADGER/WBTC")
```


jalbrekt85:
closing and reuploading with signed commit

PR #686, opened 2022-07-26
snapshot class, vote method

jalbrekt85:
handles #677, still more to explore with api. drop suggestion/feature request here
idea is that vote can be posted as such through scripts:
```py
voter = GreatApeSafe(r.badger_wallets.treasury_voter_multisig)
safe.init_snapshot(proposal_id)
safe_tx = safe.snapshot.vote("yes") or safe.snapshot.vote({"80/20 BADGER/WBTC": 1})
voter.post_safe_tx(safe_tx=safe_tx)
```
so far only supports single and weighted voting, not sure if other types are needed. full list [here](https://docs.snapshot.org/proposals/voting-types)

### DIFF
--- a/great_ape_safe/ape_api/snapshot.py
+++ b/great_ape_safe/ape_api/snapshot.py
@@ -27,6 +27,8 @@ class Snapshot():
                     choices
                     state
                     type
+                    start
+                    end
                     space {
                         id
                     }
@@ -70,6 +72,13 @@ class Snapshot():
         choices = self.propoosal_data["choices"]
         space = self.propoosal_data["space"]["id"]
         vote_type = self.propoosal_data["type"]
+        start = self.propoosal_data["start"]
+        end = self.propoosal_data["end"]
+        current_time = int(time.time())
+
+        assert current_time >= start and current_time <= end, \
+            "Vote is not within proposal time window"
+
 
         assert isinstance(choice, dict if vote_type == "weighted" else str)
 
@@ -80,11 +89,12 @@ class Snapshot():
             )
         else:
             choice = str(choices.index(choice) + 1)
-            assert choices.index(choice) <= len(choices) + 1, "choice out of bounds"
+            assert choices.index(choice) <= len(choices) + 1, \
+                "choice out of bounds"
 
         payload = {
             "version": version,
-            "timestamp": str(int(time.time())),
+            "timestamp": str(current_time),
             "space": space,
             "type": type,
             "payload": {
--- a/great_ape_safe/ape_api/snapshot.py
+++ b/great_ape_safe/ape_api/snapshot.py
@@ -105,4 +105,6 @@ class Snapshot():
             print(f"Error notifying relayer: {response.text}")
             raise
 
-        return tx_data
+        return self.safe.build_multisig_tx(
+            to=registry.eth.gnosis.sign_message_lib, value=0, data=tx_data, operation=1
+        )
--- a/great_ape_safe/ape_api/snapshot.py
+++ b/great_ape_safe/ape_api/snapshot.py
@@ -1,0 +1,108 @@
+import requests
+import json
+import time
+from rich.console import Console
+
+from eth_account import messages
+from brownie import interface
+from helpers.addresses import registry
+
+
+console = Console()
+
+
+class Snapshot():
+    def __init__(self, safe, proposal_id):
+        self.safe = safe
+
+        self.sign_message_lib = interface.ISignMessageLib(
+            registry.eth.gnosis.sign_message_lib, owner=self.safe.account
+        )
+
+        self.vote_relayer ="https://relayer.snapshot.org/api/message"
+        self.subgraph = "https://hub.snapshot.org/graphql?"
+        self.proposal_query = """
+            query($proposal_id: String) {
+                proposal(id: $proposal_id) {
+                    choices
+                    state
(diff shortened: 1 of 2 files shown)


great_ape_safe/ape_api/snapshot.py
```diff
@@ -0,0 +1,126 @@
...
+        assert isinstance(choice, dict if vote_type == "weighted" else str)
+
+        if vote_type == "weighted":
+            choice = json.dumps(
+                {choices.index(k) + 1: v for k, v in choice.items()},
+                separators=(",", ":")
+            )
+        else:
+            choice = str(choices.index(choice) + 1)
+            assert choices.index(choice) <= len(choices) + 1, \
```
petrovska-petro, line 88: everything in the class looks good!

minor thing i believe in this assert is that `choice` in L87 becomes an *int* and now choices contains something like `["For", "Against"]` so will trigger a `ValueError` 
```suggestion
            assert choice <= len(choices) + 1, \
```

jalbrekt85: yup, good catch

great_ape_safe/ape_api/snapshot.py
```diff
@@ -0,0 +1,126 @@
...
+            }
+        """
+        self.headers = {
+            "Accept": "application/json",
+            "Content-Type": "application/json",
+            "Referer": "https://snapshot.org/",
+        }
+
+        self.proposal_id = proposal_id
+        self.propoosal_data = self._get_proposal_data(self.proposal_id)
```
gosuto-inzasheru, line 44:
```suggestion
        self.proposal_data = self._get_proposal_data(self.proposal_id)
```

great_ape_safe/ape_api/snapshot.py
```diff
@@ -0,0 +1,126 @@
...
+        if not response.ok:
+            print(f"Error getting proposal data: {response.text}")
+            raise
+
+        return response.json()["data"]["proposal"]
+
+
+    def show_proposal_choices(self):
+        # external helper method to view proposal choices
+        choices = self.propoosal_data["choices"]
```
gosuto-inzasheru, line 63:
```suggestion
        choices = self.proposal_data["choices"]
```

great_ape_safe/ape_api/snapshot.py
```diff
@@ -0,0 +1,126 @@
...
+    
+
+    def vote(self, choice, version="0.1.3", type="vote", metadata=""):
+        # given a choice, contruct payload, post to vote relayer and return safe tx
+        # for single vote, pass in choice as str ex: "yes"
+        # for weighted vote, pass in choice(s) as dict ex: {"80/20 BADGER/WBTC": 100}
+        choices = self.propoosal_data["choices"]
+        space = self.propoosal_data["space"]["id"]
+        vote_type = self.propoosal_data["type"]
+        state = self.propoosal_data["state"]
```
gosuto-inzasheru, line 74:
```suggestion
        choices = self.proposal_data["choices"]
        space = self.proposal_data["space"]["id"]
        vote_type = self.proposal_data["type"]
        state = self.proposal_data["state"]
```

great_ape_safe/ape_api/snapshot.py
```diff
@@ -0,0 +1,126 @@
...
+            )
+        else:
+            choice = str(choices.index(choice) + 1)
+            assert int(choice) <= len(choices) + 1, \
+                "choice out of bounds"
+
+        payload = {
+            "version": version,
+            "timestamp": str(int(time.time())),
+            "space": space,
```
gosuto-inzasheru, line 94: nice

petrovska-petro (approved):
LGTM!

gosuto-inzasheru (commented):
just a thought, what about handling the unorthodox `post_safe_tx(safe_tx=safe_tx)` inside the `Snapshot.vote` method?

this way no unusual actions will have to be taken outside of it. only thing is to consider then is to maybe change the method naming a bit to maybe `vote_and_post`.

i dont really see a usecase for batching these or adding them to the regular multicall flow anyway, so might as well deal with it straight away.

jalbrekt85:
> just a thought, what about handling the unorthodox `post_safe_tx(safe_tx=safe_tx)` inside the `Snapshot.vote` method?
> 
> this way no unusual actions will have to be taken outside of it. only thing is to consider then is to maybe change the method naming a bit to something like `vote_and_post`.
> 
> i dont really see a use case for batching these or adding them to the regular multicall flow anyway, so might as well deal with it straight away.

good point, cant really see why it would be batched with other txs anyways. if we ever need the flexibility, something like `vote(post=True)` could also work

great_ape_safe/great_ape_safe.py
```diff
@@ -165,6 +170,8 @@ def take_snapshot(self, tokens):
 
 
     def print_snapshot(self, csv_destination=None):
+        if 'init_snapshot' in self.__dir__():
+            return
```
jalbrekt85, line 174: also note: `print_snapshot` will trigger here and cause exception, added check prevent great ape safe getting confused

gosuto-inzasheru (approved):

petrovska-petro (approved):
sweet, makes sense to include the posting within the class method. i agree that there will be few cases concatenating multiple actions together with voting into a snap

PR #692, opened 2022-07-28
WETH swap split algorithm

sajanrajdev:
Dynamically chooses the split required to swap WETH into BADGER and AURA based on the balances of these tokens obtained directly from bribes. The following cases can be very handy to understand the motivation and how the algorithm was defined:

## TEST CASES ##
**CASE 1: More than 25% of total bribes are BADGER**

Received:
$300 worth of BADGER
$300 worth of Aura
$400 worth of other assets
Total $1000 in Bribes

Current percentages:
30% BADGER
30% AURA

WETH purchase split:
0 BADGER
1 AURA

Total after WETH swap:
$300 BADGER and $700 AURA (30/70)

**CASE 2: More than 75% of total bribes are AURA**

Received:
$100 worth of BADGER
$800 worth of Aura
$100 worth of other assets
Total $1000 in Bribes

Current percentages:
10% BADGER
80% AURA

WETH purchase split:
1 BADGER
0 AURA

Total after WETH swap:
$200 BADGER and $800 AURA (20/80)

**CASE 3: No BADGER is received and less than 75% AURA is received**

Received:
$500 worth of Aura
$500 worth of other assets
Total $1000 in Bribes

Current percentages:
0% BADGER
50% AURA

WETH purchase split:
0.5 BADGER
0.5 AURA

Total after WETH swap:
$250 BADGER and $750 AURA (25/75)

**CASE 4: Less than 25% BADGER is received and less than 75% AURA is received**

Received:
$100 worth of BADGER
$400 worth of Aura
$500 worth of other assets
Total $1000 in Bribes

Current percentages:
10% BADGER
40% AURA

WETH purchase split:
0.3 BADGER
0.7 AURA

Total after WETH swap:
$250 BADGER and $750 AURA (25/75)

NOTES:
- CASE 1 and CASE 2 can't happen simultaneously
- If no BADGER nor AURA are received, the split should be 25/75

## Algorithm
Dynamic WETH split algorithm:
```
    if badger_percentage_of_total_bribes >= 0.25:
        badger_split = 0
        aura_split = 1
    elif aura_percentage_of_total_bribes >= 0.75:
        badger_split = 1
        aura_split = 0
    else:
        badger_split = (BADGER_SHARE * total_bribes_usd_balance - badger_usd_balance) / weth_usd_balance
        aura_split = (AURA_SHARE * total_bribes_usd_balance - aura_usd_balance) / weth_usd_balance
```

## Testing
The following block of code was added at the beginning of the `sell_weth()` function to ensure that specific amounts of the tokens involved would be present:
```
    #### TEST BLOCK #####
    weth_amt = 23.145335346977507 * 1e18 # 40%
    badger_amt = 6802.7210884353735 * 1e18 # 30%
    aura_amt = 9090.909090909092 * 1e18 # 30%

    weth_whale = accounts.at("0xF04a5cC80B1E94C69B48f5ee68a08CD2F09A7c3E", force=True)
    badger_whale = accounts.at("0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e", force=True)
    aura_whale = accounts.at("0x43B17088503F4CE1AED9fB302ED6BB51aD6694Fa", force=True)

    WETH.transfer(PROCESSOR, weth_amt, {"from": weth_whale})
    BADGER.transfer(PROCESSOR, badger_amt, {"from": badger_whale})
    AURA.transfer(PROCESSOR, aura_amt, {"from": aura_whale})
    #### END OF TEST BLOCK ####
```

Different amounts (from CG token/USD estimator) of WETH, BADGER and AURA were chosen to represent the cases from above. These were the results:

**CASE 1:**
BADGER percentage of bribes: 0.29980917244450095
AURA percentage of bribes: 0.29980917244450095
BADGER split: 0
AURA split: 1

**CASE 2:**
BADGER percentage of bribes: 0.09973636984455508
AURA percentage of bribes: 0.8003088101466118
BADGER split: 1
AURA split: 0

**CASE 3:**
BADGER percentage of bribes: 0.0
AURA percentage of bribes: 0.5002474682573235
BADGER split: 0.5002475907990506
AURA split: 0.49975240920094943

**CASE 4:**
BADGER percentage of bribes: 0.09972888019127885
AURA percentage of bribes: 0.3998221469486724
BADGER split: 0.3002726111114323
AURA split: 0.6997273888885678

Results match the expected!

### DIFF
--- a/scripts/badger/process_bribes_graviaura.py
+++ b/scripts/badger/process_bribes_graviaura.py
@@ -1,5 +1,5 @@
 from brownie import interface, web3
-
+from pycoingecko import CoinGeckoAPI
 from great_ape_safe import GreatApeSafe
 from helpers.addresses import r
 
@@ -15,6 +15,7 @@ DEADLINE = 60 * 60 * 3
 # percentage of the bribes that is used to buyback $badger
 # ref: https://forum.badger.finance/t/bip-95-graviaura-regulations/5716
 BADGER_SHARE = 0.25
+AURA_SHARE = 1 - BADGER_SHARE
 
 SAFE = GreatApeSafe(r.badger_wallets.techops_multisig)
 SAFE.init_badger()
@@ -55,29 +56,60 @@ def claim_and_sell_for_weth():
 
 def sell_weth():
     weth_total = WETH.balanceOf(PROCESSOR)
-    badger_share = int(WETH.balanceOf(PROCESSOR) * BADGER_SHARE)
-    aura_share = int(WETH.balanceOf(PROCESSOR) - badger_share)
+    badger_total = BADGER.balanceOf(PROCESSOR)
+    aura_total = AURA.balanceOf(PROCESSOR)
+
+    ## Estimate the amount of BADGER and AURA to buy
+    # Grab prices from coingecko
+    ids = ["weth", "aura-finance", "badger-dao"]
+    prices = CoinGeckoAPI().get_price(ids, "usd")
+
+    # Estimate total USD value of bribes
+    weth_usd_balance = weth_total / 1e18 * prices["weth"]["usd"]
+    badger_usd_balance = badger_total / 1e18 * prices["badger-dao"]["usd"]
+    aura_usd_balance = aura_total / 1e18 * prices["aura-finance"]["usd"]
+    total_bribes_usd_balance = weth_usd_balance + badger_usd_balance + aura_usd_balance
+
+    # Estimate current percentage of BADGER and AURA
+    badger_percentage = badger_usd_balance / total_bribes_usd_balance
+    aura_percentage = aura_usd_balance / total_bribes_usd_balance
+
+    # Estiamte BADGER and AURA shares to swap for (NOTE: aura_split is 1 - badger_split)
+    # If processor contains more BADGER than 25% of total bribes, don't get any more
+    if badger_percentage >= BADGER_SHARE:
+        badger_split = 0
+    # If processor contains more AURA than 75% of total bribes, don't get any more
+    elif aura_percentage >= AURA_SHARE:
+        badger_split = 1
+    # Obtain BADGER split considering the current amount sitting on the Processor
+    else:
+        badger_split = (BADGER_SHARE * total_bribes_usd_balance - badger_usd_balance) / weth_usd_balance
+
+    badger_share = int(weth_total * badger_split)
+    aura_share = int(weth_total - badger_share)
     assert badger_share + aura_share == weth_total
 
-    order_payload, order_uid = SAFE.badger.get_order_for_processor(
-        sell_token=WETH,
-        mantissa_sell=badger_share,
-        buy_token=BADGER,
-        deadline=DEADLINE,
-        coef=COEF,
-        prod=COW_PROD,
-    )
-    PROCESSOR.swapWethForBadger(order_payload, order_uid)
+    if badger_share > 0:
+        order_payload, order_uid = SAFE.badger.get_order_for_processor(
+            sell_token=WETH,
+            mantissa_sell=badger_share,
+            buy_token=BADGER,
+            deadline=DEADLINE,
+            coef=COEF,
+            prod=COW_PROD,
+        )
+        PROCESSOR.swapWethForBadger(order_payload, order_uid)
 
-    order_payload, order_uid = SAFE.badger.get_order_for_processor(
-        sell_token=WETH,
-        mantissa_sell=aura_share,
-        buy_token=AURA,
-        deadline=DEADLINE,
-        coef=COEF,
-        prod=COW_PROD,
-    )
-    PROCESSOR.swapWethForAURA(order_payload, order_uid)
+    if aura_share > 0:
+        order_payload, order_uid = SAFE.badger.get_order_for_processor(
+            sell_token=WETH,
+            mantissa_sell=aura_share,
+            buy_token=AURA,
+            deadline=DEADLINE,
+            coef=COEF,
+            prod=COW_PROD,
+        )
+        PROCESSOR.swapWethForAURA(order_payload, order_uid)
 
     # since the swapWeth methods each set their own approval, multicalling them
     # will make them replace each other. this performs one more final overwrite


scripts/badger/process_bribes_graviaura.py
```diff
@@ -55,29 +56,60 @@ def claim_and_sell_for_weth():
 
 def sell_weth():
     weth_total = WETH.balanceOf(PROCESSOR)
-    badger_share = int(WETH.balanceOf(PROCESSOR) * BADGER_SHARE)
-    aura_share = int(WETH.balanceOf(PROCESSOR) - badger_share)
+    badger_total = BADGER.balanceOf(PROCESSOR)
```
petrovska-petro, line 59: while the calculation for the splits lgtm, i am thinking that `badger_total` will be always equal to zero in the processor balance

due to this [logic](https://etherscan.io/address/0x0d724e8aee6f73b35a596c8c947c92c75eac7818#code#F1#L2075), the badger tokens claimed will never reach the processor, will be emitted into the tree and will not allow to carry this calculations, unless i am missing something

perhaps this calculation could be carry over, but needs to leverage the information of the [endpoint](https://hhand.xyz/reward/1/0x3c0989eF27e3e3fAb87a2d7C38B35880C90E63b5) instead

sajanrajdev: Good catch, this would be the case indeed.

Fetching the BADGER bribes amount from the endpoint should be the way to go and I think we should still do so to ensure that the total BADGER emitted meets the 25/75 split. Will apply this change.

GalloDaSballo: > Good catch, this would be the case indeed.
> 
> Fetching the BADGER bribes amount from the endpoint should be the way to go and I think we should still do so to ensure that the total BADGER emitted meets the 25/75 split. Will apply this change.

NOTE: We need to see what happens on a second bribe round, as I believe `amount` will be an increasing amount (not the amount for that round but the total) due to how HH code is written

GalloDaSballo: Question:

> while the calculation for the splits lgtm, i am thinking that `badger_total` will be always equal to zero in the processor balance
> 
> due to this [logic](https://etherscan.io/address/0x0d724e8aee6f73b35a596c8c947c92c75eac7818#code#F1#L2075), the badger tokens claimed will never reach the processor, will be emitted into the tree and will not allow to carry this calculations, unless i am missing something
> 
> perhaps this calculation could be carry over, but needs to leverage the information of the [endpoint](https://hhand.xyz/reward/1/0x3c0989eF27e3e3fAb87a2d7C38B35880C90E63b5) instead

Wouldn't the `badger_total` be non-zero if BADGER is a bribe?

sajanrajdev: > NOTE: We need to see what happens on a second bribe round, as I believe amount will be an increasing amount (not the amount for that round but the total) due to how HH code is written

@GalloDaSballo, interesting, we should definitely confirm this as we also rely on this quantity to claim bribes and swap them for WETH.

sajanrajdev: > Wouldn't the badger_total be non-zero if BADGER is a bribe?

As Petro pointed out, graviAURA is [hardcoded](https://etherscan.io/address/0x0d724e8aee6f73b35a596c8c947c92c75eac7818#code#F1#L2075) to directly transfer BADGER bribes to the Tree so they wouldn't fall into the processor at any point.

petrovska-petro: > > NOTE: We need to see what happens on a second bribe round, as I believe amount will be an increasing amount (not the amount for that round but the total) due to how HH code is written
> 
> @GalloDaSballo, interesting, we should definitely confirm this as we also rely on this quantity to claim bribes and swap them for WETH.

i will say that if we claim with the right cadence, post-round, it should not become an issue that amounts gets accumulated under the same bribed token cc: @GalloDaSballo

petrovska-petro: > Good catch, this would be the case indeed.
> 
> Fetching the BADGER bribes amount from the endpoint should be the way to go and I think we should still do so to ensure that the total BADGER emitted meets the 25/75 split. Will apply this change.

also i am thinking that the endpoint, likely will clear it out after claiming in step 1, so it may be worthy to save the claimable amount somewhere ( json, introduce as a param in the second step etc), otherwise, i think later on when trying to ping the endpoint, will come back empty data, since there is nothing to claim anymore

sajanrajdev: > also i am thinking that the endpoint, likely will clear it out after claiming in step 1, so it may be worthy to save the claimable amount somewhere ( json, introduce as a param in the second step etc), otherwise, i think later on when trying to ping the endpoint, will come back empty data, since there is nothing to claim anymore

This is what I thought would happen, hard to know until we try but it makes sense. I like the idea of storing on a JSON, will incorporate this.

petrovska-petro (approved):
algo for the split looks good and the workaround for storing the badger amount is neat on the local json, we will just need to make sure every round to push it to the repo to have the records too.

for this particular round, made me think if it is worthy to account for the amount of badger which is minimal and also we're mixing 2 rounds afaik

badger comes from the 1st round, which i believe was a test from @Tritium-VLK  and the usdc comes from this last round. thoughts?

sajanrajdev:
Modified function so that the total BADGER bribes amount acquired is passed to the `sell_weth` function as a parameter. This should be passed in wei. Tested this approach and it worked as expected.

gosuto-inzasheru:
shouldnt in case 1 and 2 some badger be swapped for aura or vice versa in order to obtain a 25/75 split?

i guess in the case of badger for aura that is not possible since it goes straight to the tree.. but aura for badger?

sajanrajdev:
> shouldnt in case 1 and 2 some badger be swapped for aura or vice versa in order to obtain a 25/75 split?
> 
> i guess in the case of badger for aura that is not possible since it goes straight to the tree.. but aura for badger?

@gosuto-inzasheru, as discussed with @Tritium-VLK, the intention is to never sell any BADGER nor AURA and, hence, he will modify the policy to account for these cases. The new policy will stablish that:
```
- Bribes will be sold in a split as needed for BADGER and AURA such as that the total of these assets purchased plus the total received directly as bribes leads to 25/75 rewards split respectively.
- If at any given time, more than 25% of the bribes value receive is BADGER or more than 75% of the bribes received is AURA, no more of these assets will be purchased and the the reminder of the bribes will be sold for the other token
```

gosuto-inzasheru (approved):
good to go then imo!

PR #694, opened 2022-07-29
feat: generalized get order for processor

sajanrajdev:


gosuto-inzasheru:
so `orderId` as computed by the processor comes back different than what the api returns:
computed:
```
0xe7ef6a0304143962d648faaec6d7f538711470a6d1b4e2513981dbff12952e7c0b6198b324e12a002b60162f8a130d6aedabd04c62e7f632
```
api:
```
0x3faf6ebd04356ef261b34b18699f123d4baf6e79437003799469b00fed0365530b6198b324e12a002b60162f8a130d6aedabd04c62e7f632
```

the diff seems to be in the first part, which according to the processor source code is the `orderDigest`. the second part, `owner`, matches.

scripts/badger/process_bribes_bvecvx.py
```diff
@@ -53,6 +53,7 @@ def step0_1(sim=False):
 
     for addr, mantissa in claimed.items():
         order_payload, order_uid = SAFE.badger.get_order_for_processor(
+            PROCESSOR,
             sell_token=SAFE.contract(addr),
             mantissa_sell=mantissa,
```
gosuto-inzasheru, line 58:
```suggestion
            mantissa_sell=int(mantissa),
```
breaks when using `sim=True` otherwise

petrovska-petro: i think it is casting [here](https://github.com/Badger-Finance/badger-multisig/blob/main/great_ape_safe/ape_api/cow.py#L53) later already within the cow class itself

gosuto-inzasheru: aaaaah that line needs to be `mantissa_sell = int(mantissa_sell)`! it is bugged using `==` currently.

PR #697, opened 2022-08-01
Fix the script for voting into the aura proposals

petrovska-petro:
Had this locally since [this](https://discord.com/channels/785315893960900629/999748268420501686/999801064394084352) tx was posted, but never pushed. Leaving it here for the records, and probably will be updated after #686 gets merged


jalbrekt85 (approved):
changes to correct vote weight. lgtm. can also be verified through [here](https://snapshot.mypinata.cloud/ipfs/bafkreibxmezwaojkokriws34jrubqkq5d4c6gjuuiqoktgbdkaris4uqya) (100% for 80/20 badger/wbtc)

gosuto-inzasheru (approved):

PR #712, opened 2022-08-03
fix: update ftm gnosis api url

gosuto-inzasheru:


sajanrajdev (approved):
Confirmed this to be the new API URL for fantomsafe with their team. 

LGTM

petrovska-petro (approved):
new api looks good, was updated and tested also in the safe delegate app ✅

Issue #708, opened 2022-08-03
Scripting for bribing into the Aura & Balancer marketplace in HH conditionally


petrovska-petro:
Should: 

- bribe for 80badger_20wbtc gauge

- allow passing the amount in thru the cli as following args:

```
<badger_bribed_in_aura> <badger_bribed_in_balancer>
```

in case where one of the args is zero, should not fall into that code block and only bribing in one of the markets

PR #718, opened 2022-08-04
bribe through hh script

jalbrekt85:
handles #708

```
brownie run badger/bribe_hh <badger_bribe_in_aura> <badger_bribe_in_balancer>
```

### DIFF
--- a/scripts/badger/bribe_hh.py
+++ b/scripts/badger/bribe_hh.py
@@ -1,0 +1,57 @@
+from decimal import Decimal
+
+from brownie import web3, interface
+
+from great_ape_safe import GreatApeSafe
+from helpers.addresses import r 
+
+
+def main(badger_bribe_in_aura=0, badger_bribe_in_balancer=0):
+    safe = GreatApeSafe(r.badger_wallets.treasury_ops_multisig)
+
+    badger = interface.ERC20(
+        r.treasury_tokens.BADGER, owner=safe.account
+    )
+
+    safe.take_snapshot([badger])
+
+    bribe_vault = interface.IBribeVault(
+        r.hidden_hand.bribe_vault, owner=safe.account
+    )
+
+    balancer_briber = interface.ITokenmakBribe(
+        r.hidden_hand.balancer_briber, owner=safe.account
+    )
+
+    aura_briber = interface.IAuraBribe(
+        r.hidden_hand.aura_briber, owner=safe.account
+    )
+
+    if badger_bribe_in_aura > 0:
+        prop = web3.solidityKeccak(["address"], [badger.address])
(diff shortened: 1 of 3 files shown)


scripts/badger/bribe_hh.py
```diff
@@ -0,0 +1,59 @@
...
+    balancer_briber = interface.ITokenmakBribe(
+        r.hidden_hand.balancer_briber, owner=safe.account
+    )
+
+    aura_briber = interface.IAuraBribe(
+        r.hidden_hand.aura_briber, owner=safe.account
+    )
+
+    if badger_bribe_in_aura > 0:
+        prop = web3.solidityKeccak(["address"], [badger.address])
```
petrovska-petro, line 33: in this case the proposal is built up from two variables the proposal index and the choice number which our gauge represents. [ref](https://etherscan.io/address/0x642c59937a62cf7dc92f70fd78a13cee0aa2bd9c#code#L934)

since this will change from round to round, i advise to added them also as arguments to pass in the cli

latest proposal set [tx](https://etherscan.io/tx/0xe0a89600d65a8716020c21ca099f8925d7e8c8da15945380a210e459066e1939). Proposal index is 11 and choice is 38

petrovska-petro: checking on the graphql endpoint of snapshot. i think to leave it more programmatically solved. it's better to pass perhaps the proposal id where we want to bribe for and deduct from there the proposal index and also the choice number, since we know which choice is our target (`"80/20 BADGER/WBTC"`)

scripts/badger/bribe_hh.py
```diff
@@ -0,0 +1,99 @@
...
+        # grab data from proposals to find out the proposal index
+        response = requests.post(
+            url,
+            json={"query": query_proposals},
+        )
+        # reverse the order to have from oldest to newest
+        proposals = response.json()["data"]["proposals"][::-1]
+        for proposal in proposals:
+            if aura_proposal_id in proposal["id"]:
+                proposal_index = proposals.index(proposal)
```
jalbrekt85, line 73: nice solution, lgtm

Tritium-VLK (dismissed):

scripts/badger/bribe_hh.py
gosuto-inzasheru, line 73:
```suggestion
            if aura_proposal_id == proposal["id"]:
                proposal_index = proposals.index(proposal)
                break
```
- both strings have to be identical
- break out of for loop once index is found

gosuto-inzasheru (commented):
minor adjustments but good to go

scripts/badger/bribe_hh.py
petrovska-petro: good one!

gosuto-inzasheru (approved):

petrovska-petro (approved):

Issue #720, opened 2022-08-05
Migrate bauraBAL to strategy that autocompounds bb-a-usd


sajanrajdev:
auraBAL's strategy was modified to autocompound bb-a-usd rewards. Since the current strategy only holds test assets, it was decided that it was more efficient to migrate off of this one to a freshly deployed new one with the patched logic to avoid Timelock operations. The process to migrate for V1.5 is as follows:
harvest -> withdrawToVault -> setStrategy -> earn (Withdrawing to vault will claim rewards, hence we should harvest beforehand atomically).

Vault proxy: [0x37d9D2C6035b744849C15F1BFEE8F268a20fCBd8](https://etherscan.io/address/0x37d9D2C6035b744849C15F1BFEE8F268a20fCBd8)
Current strategy proxy: [0xED6d51A82065725e283fC035f488213e869FD976](https://etherscan.io/address/0xED6d51A82065725e283fC035f488213e869FD976)
New strategy proxy: [0x32BdF2B35Bbf8dAca3561C3ADE59c941488E9c69](https://etherscan.io/address/0x32BdF2B35Bbf8dAca3561C3ADE59c941488E9c69#readProxyContract)

To be called from devMulti

sajanrajdev:
Changes: https://github.com/Badger-Finance/auraBal-staker/pull/2#event-7136240691

gosuto-inzasheru:
https://etherscan.io/tx/0x9a62dffaaf52c7ca79d9eaff2933154dc646d9e7dc2dc572ca6c72f02d2a9b90

PR #721, opened 2022-08-05
feat: baurabal strat migration script

sajanrajdev:
Tackles issue #720 

Run with: 
```
brownie run scripts/issue/720/migrate_baurabal_strategy.py
```

Simulation:
```
brownie run scripts/issue/720/migrate_baurabal_strategy.py main true
```


shuklaayush (approved):

petrovska-petro (approved):
Migration scripts checks-up necessary steps for migrating to new strat which only sells BbaUsd if above certain threshold set ✅ 

Simulation looks good, `bb_a_usd.balanceOf` assert correctly when the threshold is set to zero, expecting zero bal in the strat ✅

Issue #709, opened 2022-08-03
Claim bribes from voter


petrovska-petro:
Our voter msig will vote with its stack to the 80badger_20wbtc gauge, likely it will be entitled for claiming bribes every round moving forward. Worth having ready the following actions into the script:

- claims from the rewards distributor in hidden hands

- sends the bribes back to trops

PR #726, opened 2022-08-08
claim bribes from voter script

jalbrekt85:
handles #709

```coffeescript
brownie run badger/claim_hh
```

### DIFF
--- a/scripts/badger/claim_hh.py
+++ b/scripts/badger/claim_hh.py
@@ -1,5 +1,3 @@
-import requests
-
 from great_ape_safe import GreatApeSafe
 from helpers.addresses import r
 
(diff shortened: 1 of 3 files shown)


scripts/badger/claim_hh.py
```diff
@@ -0,0 +1,44 @@
...
+
+    aggregate = {"tokens": [], "amounts": []}
+    for item in data:
+        aggregate["tokens"].append(item["token"])
+        aggregate["amounts"].append(item["claimMetadata"]["amount"])
+
+    voter.take_snapshot(tokens=aggregate["tokens"])
+    destination.take_snapshot(tokens=aggregate["tokens"])
+
+    rewards.claim(
```
petrovska-petro, line 25: i will say given that we have this [method](https://github.com/Badger-Finance/badger-multisig/blob/main/great_ape_safe/ape_api/badger.py#L154) in the badger class. will advise to leverage and allow passing an address for where bribes will be claimed

jalbrekt85: good idea

jalbrekt85: updated in [7d83c21](https://github.com/Badger-Finance/badger-multisig/pull/726/commits/7d83c21c22e4cc69971be521696cc371a355d3d1), allows address to be passed into `claim_bribes_hidden_hands` but claims from gravi aura strat by default as to not break other scripts

petrovska-petro (approved):
nice ! 

ready to recycle those badger bribes 🔥

gosuto-inzasheru (approved):
rock n roll

Issue #729, opened 2022-08-09
Deprecate Polygon Vaults


axejintao:
# Summary

All polygon vaults qualify for our deprecation criteria. We should update their state to reflect that as they are already not being maintained.

PR #730, opened 2022-08-09
chore: add script for polygon vault deprecation

axejintao:
# Summary

- deprecates remaining 3 polygon vaults
- solve #729

### DIFF
--- a/scripts/issue/729/deprecate_polygon.py
+++ b/scripts/issue/729/deprecate_polygon.py
@@ -1,0 +1,30 @@
+from great_ape_safe import GreatApeSafe
+from helpers.addresses import registry
+
+TECH_OPS = registry.poly.badger_wallets.techops_multisig
+
+REGISTRY = registry.poly.registry_v2
+
+WBTC_IBBTC_SLP = registry.poly.sett_vaults.bslpibBTCWbtc
+WBTC_USDC_QLP = registry.poly.sett_vaults.bqlpUsdcWbtc
+AWBTC_REN_CRV = registry.poly.sett_vaults.bcrvRenBTC
+
+def deprecate_polygon():
+    safe = GreatApeSafe(TECH_OPS)
+    safe.init_badger()
+
+    registry = safe.contract(REGISTRY)
+
+    safe.badger.demote_vault(WBTC_IBBTC_SLP, 0)
+    safe.badger.demote_vault(WBTC_USDC_QLP, 0)
+    safe.badger.demote_vault(AWBTC_REN_CRV, 0)
+
+    info = registry.productionVaultInfoByVault(WBTC_IBBTC_SLP)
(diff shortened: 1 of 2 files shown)


sajanrajdev (dismissed):
Pushed a change to the Badger.py library so that Eth specific contracts are only initialized when running on ETH, getting an error on Polygon otherwise.

Deprecated the three active vaults on Polygon and confirms this.

LGTM

For other reviewers, run with: 
```
brownie run scripts/issue/729/deprecate_polygon.py deprecate_polygon --network polygon-main-fork
```

gosuto-inzasheru:
i dont see a need for 822a077 @sajanrajdev. i believe the fact that we started using `r` from the registry instead of `registry.<chain>` in for example `badger.py` should be enough. `GreatApeSafe.contract` just returns `None` in the case of an invalid contract. how did that break anything on your side?

scripts/issue/729/deprecate_polygon.py
```diff
@@ -0,0 +1,30 @@
...
+
+TECH_OPS = registry.poly.badger_wallets.techops_multisig
+
+REGISTRY = registry.poly.registry_v2
+
+WBTC_IBBTC_SLP = registry.poly.sett_vaults.bslpibBTCWbtc
+WBTC_USDC_QLP = registry.poly.sett_vaults.bqlpUsdcWbtc
+AWBTC_REN_CRV = registry.poly.sett_vaults.bcrvRenBTC
+
+def deprecate_polygon():
```
gosuto-inzasheru, line 12: would opt for using default `main` here if there is no good reason not to. this saves having to specify the func name when calling the script:
```
brownie run issue/729/deprecate_polygon deprecate_polygon
```
vs
```
brownie run issue/729/deprecate_polygon
```

gosuto-inzasheru (approved):
runs fine, also with `HEAD` at ee470e8.

would be nice to get rid of your commit that follows that one actually @sajanrajdev, unless there is something im missing?

sajanrajdev:
> i dont see a need for [822a077](https://github.com/Badger-Finance/badger-multisig/commit/822a077bbacb10880482de17063e26a93cce6729) @sajanrajdev. i believe the fact that we started using `r` from the registry instead of `registry.<chain>` in for example `badger.py` should be enough. `GreatApeSafe.contract` just returns `None` in the case of an invalid contract. how did that break anything on your side?

@gosuto-inzasheru, Interesting, will remove my commit and dig into what is breaking on my side but you are right, it should be enough.

gosuto-inzasheru (approved):

sajanrajdev (approved):
Script LGTM and was able to run it properly to confirm with the reverted changes. I continue to have issues with it but it is likely due to a local brownie mis config or rpc. 

Approving since it runs properly you for you as it is.

Issue #734, opened 2022-08-11
Stake auraBAL idle in treasury


petrovska-petro:
After influence pod meeting we're discussing that we're missing >100% yield already for almost 2 weeks. The consensus was that was worthy to put it to work as soon as possible and then wd and dog-food into our set

Open questions: 

- if leaving this week already from beta, worth instead depositing there. cc: @sajanrajdev @Tritium-VLK 

- they're going next to guarded state from experimental so likely to take another week or two to switch to prod state

petrovska-petro:
```
snapshot result for 0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e:
                    balance_before         balance_after               balance_delta
symbol                                                                              
auraBAL  23,846.351001653578614506  0.000000000000000000  -23,846.351001653578614506
```

```
{
│   'ethereum_client': <gnosis.eth.ethereum_client.EthereumClient object at 0x7fabf688df10>,
│   'safe_address': '0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e',
│   'to': '0x40A2aCCbd92BCA938b02010E17A5b8929b49130D',
│   'value': 0,
│   'data': HexBytes('0x8d80ff0a000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000f200616e8bfa43f920657b3497dbf40d6b1a02d4608d00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b30000000000000000000000005e5ea2048475854a5702f5b8468a51ba1296efcc00000000000000000000000000000000000000000000050cb699649f19f8c6ea005e5ea2048475854a5702f5b8468a51ba1296efcc000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000048dcb40610000000000000000000000000000'),
│   'operation': 1,
│   'safe_tx_gas': 0,
│   'base_gas': 0,
│   'gas_price': 0,
│   'gas_token': '0x0000000000000000000000000000000000000000',
│   'refund_receiver': '0x0000000000000000000000000000000000000000',
│   'signatures': b'',
│   '_safe_nonce': 53,
│   '_safe_version': '1.3.0',
│   '_chain_id': None
}
```

petrovska-petro:
[...]

PR #740, opened 2022-08-12
Add script to stake directly in aura aurabal

petrovska-petro:
Tackles #734 

```
brownie run issue/734/stake_aurabal
```

**Note:** tried to add also in the strategy script some minor additions to adapt around the fact that we will have the aurabal staked in aura and we need to claim from it also.

Also, we get deducted [here](https://etherscan.io/address/0xead792b55340aa20181a80d6a16db6a0ecd1b827#code#F33#L186) some portion while going thru the wrapper route, which i think is worthy not giving away.


scripts/badger/aura_tcl_strategy.py
```diff
@@ -109,14 +111,16 @@ def main():
         wrapper.deposit(
             bal_to_deposit,  # uint256 _amount
             wrapper_aurabal_out,  # uint256 _minOut
-            False,  # bool _lock
+            True,  # bool _lock. Set as true otherwise, we get deducted a "penalty"
```
gosuto-inzasheru, line 114: not sure im following this change. wont this stake our aurabal, making it impossible to deposit into baurabal..?

maybe it is better to already check vault status here: lock aurabal directly in case the vault isnt open yet, but take the penalty so we can dogfood immediately otherwise.

also makes we realise this would be a good use case for a zap; zap into baurabal straight from bal :)

petrovska-petro: the naming of the argument may be confusing, this does not stake the auraBAL, as long as the `staking_addr` is the zero address. Passing this as `True` we're increasing the amount lock and time for the voting proxy of aura, managing the *veBAL* voting power and getting an incentive by doing so, instead of getting deducted the incentive out of our portion of bpt

gosuto-inzasheru: ah you are right, i forgot that passing `ZERO_ADDRESS` prevents from staking.

gosuto-inzasheru (approved):

jalbrekt85 (approved):
lgtm

Issue #743, opened 2022-08-12
rewrite process_bribes_graviaura to post using raw calldata instead


gosuto-inzasheru:
script as merged in #736 somehow reverts on calling `sellBribeForWeth`:
```
order_payload, order_uid = SAFE.badger.get_order_for_processor(PROCESSOR, SAFE.contract('0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48'), 739356929, WETH, deadline=DEADLINE, coef=COEF)
```
```
PROCESSOR.sellBribeForWeth(order_payload, order_uid)
```
result:
![Screenshot_2022-08-12_at_18 08 24](https://user-images.githubusercontent.com/2835259/184439206-1dee3045-f83d-4fff-a4f6-e42d3b40803c.png)

call trace:
![Screenshot_2022-08-12_at_21 17 05](https://user-images.githubusercontent.com/2835259/184439167-a5370665-5321-469c-b1ee-eb4da0289622.png)

current idea is that the unexpected error stems from a revert which is actually expected (https://etherscan.io/address/0xb27308f9F90D607463bb33eA1BeBb41C27CE5AB6#code#F1#L90). possible solution is to rewrite the script so that we can post the bribe selling as raw calldata, and thus prevent running a simulation onfork. this should circumvent the (expected) revert.

gosuto-inzasheru:
approach in outdated, problem lies with the pricer and its interface with the processor.

PR #744, opened 2022-08-12
feat: sell bribes using raw calldata

gosuto-inzasheru:
solves #743


gosuto-inzasheru:
approach is outdated, problem lies with the pricer and its interface with the processor.

Issue #745, opened 2022-08-12
hh api has a descrepancy between `amount` and `claimable`


gosuto-inzasheru:
![Screenshot 2022-08-12 at 23 14 42](https://user-images.githubusercontent.com/2835259/184445982-e1b596bf-8d74-4d0a-b8d4-6b710dbcdb19.png)

it appears `claimable` is what we are actually after, not `amount`.

PR #746, opened 2022-08-12
fix: actual amount of tokens claimed is stored in `claimable` not in `claimMetadata.amount`

gosuto-inzasheru:
solves #745

### DIFF
--- a/great_ape_safe/ape_api/badger.py
+++ b/great_ape_safe/ape_api/badger.py
@@ -151,7 +151,7 @@ class Badger():
                 claimables.append(token_addr)
         if len(claimables) > 0:
             self.strat_bvecvx.claimBribesFromConvex(claimables)
-    
+
 
     def get_hh_data(self, address=None):
         """
@@ -175,7 +175,7 @@ class Badger():
         aggregate = {"tokens": [], "amounts": []}
         for item in data:
             aggregate["tokens"].append(item["token"])
-            aggregate["amounts"].append(item["claimMetadata"]["amount"])
+            aggregate["amounts"].append(item["claimable"].replace('.', ''))
 
         metadata = [(item['claimMetadata']['identifier'], item['claimMetadata']
                      ['account'], item['claimMetadata']['amount'], item['claimMetadata']['merkleProof'])


gosuto-inzasheru:
ref: https://hhand.xyz/reward/1/0x3c0989eF27e3e3fAb87a2d7C38B35880C90E63b5

petrovska-petro (approved):
nice, hope the discrepancy next time is not the other way around from the endpoint. will assume that what is pass as a proof metadata should be more precise somehow

great_ape_safe/ape_api/badger.py
```diff
@@ -175,7 +175,7 @@ def claim_bribes_hidden_hands(self, claim_from_strat=True) -> dict:
         aggregate = {"tokens": [], "amounts": []}
         for item in data:
             aggregate["tokens"].append(item["token"])
-            aggregate["amounts"].append(item["claimMetadata"]["amount"])
+            aggregate["amounts"].append(item["claimable"].replace('.', ''))
```
sajanrajdev, line 178: By replacing `.` with `''`, we are counting on the amount of decimals provided to be complete and I am not sure it will always be the case. For instance, in the example from the issue, the claimable is `739.356929` but removing the `.` leads to `739356929 wei` which is only: `0.000000000739356929`. Maybe I am missing something?

great_ape_safe/ape_api/badger.py
```diff
@@ -175,7 +175,7 @@ def claim_bribes_hidden_hands(self, claim_from_strat=True) -> dict:
...
 
         metadata = [(item['claimMetadata']['identifier'], item['claimMetadata']
                      ['account'], item['claimMetadata']['amount'], item['claimMetadata']['merkleProof'])
```
sajanrajdev, line 181: Shouldn't we also change it here? Not sure what the purpose of `amount` is then...

great_ape_safe/ape_api/badger.py
gosuto-inzasheru: usdc only has 6 decimals

great_ape_safe/ape_api/badger.py
gosuto-inzasheru: i think the amount sent as metadata is more related to the proof and the merkle tree, not the actual individual claimable.

great_ape_safe/ape_api/badger.py
sajanrajdev: Ah, that's what I was missing... 😬

sajanrajdev (approved):
LGTM

Issue #751, opened 2022-08-16
Ribbon Call Options Deposit


Hal69k:
approve the vault to transfer badger from the multisig + call deposit() on the vault contract https://etherscan.io/address/0x270f4a26a3fe5766ccef9608718491bb057be238

Deposit 100k BADGER into the above vault contract
Needs to be in before august 26th 

Treasury Decision Approval:
![image](https://user-images.githubusercontent.com/90224840/184977315-17091819-2ba0-4080-9039-fcbd1c0d98e7.png)

gosuto-inzasheru:
```
Events In This Transaction
--------------------------
├── Badger (0x3472A5A71965499acd81997a54BBA8D852C6E53d)
│   └── Approval
│       ├── _owner: 0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e
│       ├── _spender: 0x270F4a26a3fE5766CcEF9608718491bb057Be238
│       └── _amount: 100000000000000000000000
│   
├── Ribbon BADGER Treasury Vault (0x270F4a26a3fE5766CcEF9608718491bb057Be238)
│   └── Deposit
│       ├── account: 0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e
│       ├── amount: 100000000000000000000000
│       └── round: 1
│   
├── Badger (0x3472A5A71965499acd81997a54BBA8D852C6E53d)
│   └── Transfer
│       ├── _from: 0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e
│       ├── _to: 0x270F4a26a3fE5766CcEF9608718491bb057Be238
│       └── _amount: 100000000000000000000000
│   
└── Gnosis Safe (0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e)
    └── ExecutionSuccess
        ├── txHash: 0x07e82e00649b8f071fd9faa68b01391eee9bc8a4ee49beabdc126ed22672853a
        └── payment: 0
```
```
{
│   'ethereum_client': <gnosis.eth.ethereum_client.EthereumClient object at 0x12256bf70>,
│   'safe_address': '0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e',
│   'to': '0x40A2aCCbd92BCA938b02010E17A5b8929b49130D',
│   'value': 0,
[...]

PR #755, opened 2022-08-17
script: deposit into ribbon badger vault

gosuto-inzasheru:
script to solve #751


petrovska-petro (approved):
smooth!. worth noting that deployer addr veracity was checked ✅

jalbrekt85 (approved):
lgtm

Issue #757, opened 2022-08-17
Add general scripts to interact with Registry and Guestlists


sajanrajdev:
Add interactive functions to promote, demote, change metadata on the Registry from TechOps directly from the console. These may provide a user interface on cli for ease of use and allow for batching as needed.

Add functions to interact with the Guestlist contract as well as these will be delegated to techOps from now and on.

PR #762, opened 2022-08-18
RegistryV2 and Guestlist scripts library

sajanrajdev:
Tackles issue #757 

Introduces the following scripts to perform the following on the RegistryV2 and Guestlist with agility:

### RegistryV2
- Batch promote vaults from a list
- Batch demote vaults from a list
- Change metadata (the vault's behaviors allowed are limited to a list included in this PR - behaviors outside of this list are currently not being picked up by the UI's SDK).

### Guestlist
- Change the user cap of a given vault
- Change the total cap of a given vault
- Remove the caps for a given vault
- Remove the caps for a batch of vaults

### DIFF
--- a/scripts/badger/guestlist.py
+++ b/scripts/badger/guestlist.py
@@ -1,0 +1,69 @@
+from rich.console import Console
+from great_ape_safe import GreatApeSafe
+from helpers.addresses import get_registry as registry
+from helpers.constants import MaxUint256, AddressZero
+
+C = Console()
+
+# assuming techOps
+r = registry()
+SAFE = GreatApeSafe(r.badger_wallets.techops_multisig)
+SAFE.init_badger()
+
+# Add the vaults addresses to remove caps from
+VAULTS_ARRAY = ["0xf8f5677B6bCecdb9be94AE8f6770a05a6C53C378"]
+
+# Sets a new user cap for a given vault
+def set_vault_user_cap(vault, cap):
+    vault = SAFE.contract(vault)
+    guestlist = vault.guestList()
+    if guestlist != AddressZero:
+        C.print(f"Guestlist: {guestlist}")
+        SAFE.badger.set_guestlist_user_cap(guestlist, cap)
+    else:
+        C.print(f"[red]No guestlist set for this vault![/red]")
+
+    SAFE.post_safe_tx()
+
+# Sets a new total cap for a given vault
+def set_vault_total_cap(vault, cap):
+    vault = SAFE.contract(vault)
+    guestlist = vault.guestList()
+    if guestlist != AddressZero:
+        C.print(f"Guestlist: {guestlist}")
+        SAFE.badger.set_guestlist_total_cap(guestlist, cap)
+    else:
+        C.print(f"[red]No guestlist set for this vault![/red]")
+
+    SAFE.post_safe_tx()
+
+# Sets the user and total cap to MAXUINT256 for a given vault
+def remove_caps_for_vault(vault):
+    vault = SAFE.contract(vault)
+    guestlist = vault.guestList()
+    if guestlist != AddressZero:
+        C.print(f"Guestlist: {guestlist}")
+        SAFE.badger.set_guestlist_total_cap(guestlist, int(MaxUint256))
+        SAFE.badger.set_guestlist_user_cap(guestlist, int(MaxUint256))
+    else:
+        C.print(f"[red]No guestlist set for this vault![/red]")
+
+    SAFE.post_safe_tx()
+
+# Sets the user and total cap to MAXUINT256 for a batch of vaults
+def remove_caps_for_vaults():
+    if VAULTS_ARRAY:
+        for vault in VAULTS_ARRAY:
+            vault = SAFE.contract(vault)
+            C.print(f"Vault: {vault}")
+            guestlist = vault.guestList()
+            if guestlist != AddressZero:
+                C.print(f"Guestlist: {guestlist}")
+                SAFE.badger.set_guestlist_total_cap(guestlist, int(MaxUint256))
+                SAFE.badger.set_guestlist_user_cap(guestlist, int(MaxUint256))
+            else:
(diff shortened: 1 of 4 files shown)


scripts/badger/guestlist.py
```diff
@@ -0,0 +1,69 @@
...
+# Add the vaults addresses to remove caps from
+VAULTS_ARRAY = []
+
+# Sets a new user cap for a given vault
+def set_vault_user_cap(vault, cap):
+    vault = SAFE.contract(vault)
+    guestlist = vault.guestList()
+    if guestlist != AddressZero:
+        C.print(f"Guestlist: {guestlist}")
+        SAFE.badger.set_guestlist_user_cap(guestlist, cap)
```
petrovska-petro, line 22: will advise in the script to only post the tx within the *if* code block, similarly in the methods below
```suggestion
        SAFE.badger.set_guestlist_user_cap(guestlist, cap)
        SAFE.post_safe_tx()
```

scripts/badger/registry.py
```diff
@@ -2,20 +2,99 @@
...
+    if "behavior=" in metadata:
+        new_behavior = metadata.split("behavior=")[1]
+        if new_behavior not in ALLOWED_BEHAVIORS:
+            C.print(f"[red]Wrong vault behavior! Allowed behaviors: {ALLOWED_BEHAVIORS}[/red]")
+            return
+    else:
+        C.print(f"[red]Wrong metadata format[/red]")
+        return
+
+    info = SAFE.badger.registry_v2.productionVaultInfoByVault(vault)
```
petrovska-petro, line 80: in this occasion i think is worthy leveraging the `dict()` capability, mainly to be more explicit on which variable we're interested on, feels more natural to read directly the *'metadata'* than *[3]*. it is probably an opinionated comment, but if you agree would be great to adapt the script to use data all around
```suggestion
    metadata = SAFE.badger.registry_v2.productionVaultInfoByVault(vault).dict()['metadata']
```

scripts/badger/guestlist.py
gosuto-inzasheru: L26 will then need to be removed, since `post_safe_tx` doesnt necessarily end the script.

gosuto-inzasheru: ah actually same goes for all these funcs

scripts/badger/registry.py
```diff
@@ -2,20 +2,99 @@
...
 from helpers.addresses import get_registry as registry
 
-console = Console()
+C = Console()
 
 # assuming techOps
-contract_registry = registry()
-chain_safe = GreatApeSafe(contract_registry.badger_wallets.techops_multisig)
-chain_safe.init_badger()
+r = registry()
```
gosuto-inzasheru, line 8: i think this behaviour was simplified by just importing `r` from `helpers.addresses` no?

scripts/badger/registry.py
```diff
@@ -2,20 +2,99 @@
...
+  'DCA',
+  'Ecosystem',
+  'Ecosystem Helper',
+  'None'
+]
+
+# Add any vaults's addresses that you want to promote or demote
+VAULTS_ARRAY = []
+# Add the new statuses you want to promote or promote each vault to
+NEW_STATUS_ARRAY = []
```
gosuto-inzasheru, line 27: this might make more sense as a dict, ie `{vault1: new_status, vault2: new_status}` etc. but minor detail ofc.

scripts/badger/registry.py
gosuto-inzasheru: i like this, makes it more readable. some goes for L33 and L51 then.

scripts/badger/registry.py
```diff
@@ -2,20 +2,99 @@
...
+        for vault in VAULTS_ARRAY:
+            info = SAFE.badger.registry_v2.productionVaultInfoByVault(vault)
+            new_status = NEW_STATUS_ARRAY[VAULTS_ARRAY.index(vault)]
+            if info[2] < new_status:
+                print(info)
+                SAFE.badger.promote_vault(info[0], info[1], info[3], new_status)
+                assert SAFE.badger.registry_v2.productionVaultInfoByVault(vault)[2] == new_status
+                C.print(f"[green]Promoting {vault} to {new_status}[/green]")
+            else:
+                C.print(f"[red]Promoting {vault} to lower state, current state is {info[2]}[/red]")
```
gosuto-inzasheru, line 41: this is a bit confusing since it is not _actually_ promoting.

makes we wonder if an `update_status` func or something would actually be more efficient than the two current `promote_vaults` and `demote_vaults` funcs.

gosuto-inzasheru (commented):
all pretty minor comments, functionality wont really be affected but more usability/readability. which is whole point of this pr of course 😄

scripts/badger/guestlist.py
sajanrajdev: Good call, applying the change

gosuto-inzasheru (approved):

scripts/badger/guestlist.py
```diff
@@ -0,0 +1,65 @@
...
+    if VAULTS_ARRAY:
+        for vault in VAULTS_ARRAY:
+            vault = SAFE.contract(vault)
+            C.print(f"Vault: {vault}")
+            guestlist = vault.guestList()
+            if guestlist != AddressZero:
+                C.print(f"Guestlist: {guestlist}")
+                SAFE.badger.set_guestlist_total_cap(guestlist, int(MaxUint256))
+                SAFE.badger.set_guestlist_user_cap(guestlist, int(MaxUint256))
+                SAFE.post_safe_tx()
```
petrovska-petro, line 61: everything looks good, just i think this post should happen outside of the *for loop*  to post all of them in a single tx

sajanrajdev: You are right, my bad. Moving down

gosuto-inzasheru (approved):

petrovska-petro (approved):

PR #768, opened 2022-08-22
chore: update august swapr gauges

btcookies:
Note: looks like ibbtc/weth is no longer receiving rewards, deprecate vault


gosuto-inzasheru (approved):
checks out

Issue #760, opened 2022-08-18
Modify aura bribes processing script to exclude BADGER bribes


sajanrajdev:
This issue related to https://github.com/Badger-Finance/badger-projects/issues/140

The [sell_weth](https://github.com/Badger-Finance/badger-multisig/blob/62f2d94b7b80ed41b9b9539d743f89f8e03eb386/scripts/badger/process_bribes_graviaura.py#L77) script currently takes into consideration the amount of BADGER obtained from bribes and uses this amount to estimate the extra amount of BADGER required to obtained from welling wETH in order to meet the 25% split.

From now and on, no more BADGER bribes will be hunt and, therefore, any BADGER claimed will come from the DAO's voting power and shouldn't be accounted for the total user's bribes. This means that BADGER should be excluded from the estimation and, hence, the 25% of the BADGER to be distributed from bribes will come from selling wETH. The script need to be adapted for this.

PR #770, opened 2022-08-23
feat: adapted sell weth script to check badger

sajanrajdev:
Tackles issue #760 

With these changes, only the BADGER sent to the processor will be accounted towards HODLrs hunted bribes. For the time being, BADGER bribes will be redirected to Treasury so, unless someone manually sends BADGER to the processor, no BADGER will be transferred there from the strat. It is likely that the balance of BADGER on the processor will be 0 and 25% of the wETH will be sold for BADGER to compensate.

### DIFF
--- a/scripts/badger/process_bribes_graviaura.py
+++ b/scripts/badger/process_bribes_graviaura.py
@@ -55,29 +55,13 @@ def claim_and_sell_for_weth():
             )
             PROCESSOR.sellBribeForWeth(order_payload, order_uid)
 
-        # If Badger is claimed, we store the amount to be able to pass it to next 
-        # step's script in order to estimate the Badger split to buy from WETH.
-        if addr == BADGER.address:
-            dump_dir = "data/badger/hh_badger_bribes/"
-            file_name = chain.time()
-            os.makedirs(dump_dir, exist_ok=True)
-            with open(f'{dump_dir}{file_name}.json', 'w') as f:
-                bribe_data = {
-                    'address': addr,
-                    'mantissa': mantissa,
-                    'timestamp': file_name
-                }
-                json.dump(bribe_data, f, indent=4, sort_keys=True)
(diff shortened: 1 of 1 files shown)


petrovska-petro (approved):
update the script with the assumption that only badger which will reach the processor are claims from the delegate voting with the DAO's vp into the 80badger_20wbtcc gauge. Currently dependant on upgrading the strat with the [changes](https://github.com/Badger-Finance/vested-aura/pull/14) worked so badger indeed is not sent directly to tree ✅

sajanrajdev:
> update the script with the assumption that only badger which will reach the processor are claims from the delegate voting with the DAO's vp into the 80badger_20wbtcc gauge. Currently dependant on upgrading the strat with the [changes](https://github.com/Badger-Finance/vested-aura/pull/14) worked so badger indeed is not sent directly to tree ✅

Just a small correction: The voter was currently set to vote for the BADGER/wBTC pool using the 10% voting power belonging to the DAO and, hence, any bribes obtained from this will be redirected to Treasury. The voter will not vote for any other pool bribed with BADGER so, for the time being, we can expect no BADGER to be sent to the processor. Should this change in the future and BADGER bribes are sent to the processor for any reason, the script will be able to handle the calculation based on this amount.

petrovska-petro:
> 

Got it, thanks for the clarification 🤝

petrovska-petro:
> Just a small correction: The voter was currently set to vote for the BADGER/wBTC pool using the 10% voting power belonging to the DAO and, hence, any bribes obtained from this will be redirected to Treasury. The voter will not vote for any other pool bribed with BADGER so, for the time being, we can expect no BADGER to be sent to the processor. Should this change in the future and BADGER bribes are sent to the processor for any reason, the script will be able to handle the calculation based on this amount.

Got it, thanks for the clarification 🤝

gosuto-inzasheru (approved):
as discussed during call, makes sense now.

PR #777, opened 2022-08-25
feat: graviaura upgrade queue data

sajanrajdev:


petrovska-petro (approved):
matches tx event info emitted ✅

gosuto-inzasheru (approved):

Issue #492, opened 2022-05-25
how to generate merkle tree proof for `claimEpoch` on opolis?


jorijnsmit:

gosuto-inzasheru:
the proof is currently generated by their (logged in) ui, no further details available yet.

gosuto-inzasheru:
so now that we have migrated away from their ui, it turns out the stream was actually a superfluid stream.

we can currently access it through the safe app ui, but there might also be a direct onchain way of claiming possible.

ref: https://docs.superfluid.finance/superfluid/

sajanrajdev:
Being streamed to TechOps on Polygon. Keep an eye and handle in the future.

PR #784, opened 2022-08-29
chore: add opolis magic wallet

gosuto-inzasheru:


petrovska-petro (approved):
hard to verify at my end, even tho it seems to interact with the right contract staking and downgrading

is it possible to move the staked work to a msig instead?

gosuto-inzasheru:
yes i am checking this with opolis again. currently still blocked by #492

PR #788, opened 2022-08-31
additions to the graviaura bribes processing script

gosuto-inzasheru:
this due to having to avoid using the pricer


shuklaayush (approved):

scripts/badger/process_bribes_graviaura.py
```diff
@@ -58,6 +63,17 @@ def claim_and_sell_for_weth():
     SAFE.post_safe_tx()
 
 
+def ragequit(token_list=r.bribe_tokens_claimable_graviaura.values()):
+    DEV.take_snapshot(token_list)
+    for token in token_list:
```
sajanrajdev, line 68: This assumes that we will always be claiming these bribe tokens. I would instead check if there's a balance of these on the processor and only rage quit on a token if that's the case.

sajanrajdev: Handled [here](https://github.com/Badger-Finance/badger-avatars/pull/9)

gosuto-inzasheru: yea the idea was to use the `token_list` arg in a case like that, but a programmatic check would indeed make things a bit easier :) solved in f73f7e1

gosuto-inzasheru:
btw also added a check to not ragequit aura, those will eventually have to go back to the processor anyway.

scripts/badger/process_bribes_graviaura.py
```diff
@@ -137,6 +154,33 @@ def sell_weth():
...
+    usdc_mantissa = int(usdc_mantissa)
+
+    USDC = interface.ERC20(r.treasury_tokens.USDC, owner=VAULT.account)
+    BADGER = interface.ERC20(r.treasury_tokens.BADGER, owner=VAULT.account)
+    AURA = interface.ERC20(r.treasury_tokens.AURA, owner=VAULT.account)
+    GRAVI_AURA = interface.ITheVault(r.sett_vaults.graviAURA, owner=VAULT.account)
+
+    proc = GreatApeSafe(PROCESSOR.address)
+    proc.take_snapshot([USDC, BADGER, AURA, GRAVI_AURA])
+
```
petrovska-petro, line 167: maybe before placing the order is a good measure to check that indeed we have the usdc balance to avoid what happens in the past, at least we could get an alert in the terminal to be aware

gosuto-inzasheru: yea that is a good idea, although that should happen on the `Cow` class level, not here. i will write a ticket.

petrovska-petro (approved):

sajanrajdev (approved):

Issue #787, opened 2022-08-31
Voter msig allow casting vote in different gauges


petrovska-petro:
Likely in the following round for voting gauges, in the aura proposal the badger/wbtc gauge may not appear since it is >10%. 

So, we may need to adapt the script so it can vote alternatively to the digg gauge for example given a cli param

PR #789, opened 2022-09-01
allow votes in different gauges in vote_proposal script

jalbrekt85:
handles #787 `gauge` can now be passed into `vote_proposal.py` with the value corresponding to `Gauges` enum


jalbrekt85:
changes added from last pr, redoing

PR #790, opened 2022-09-01
allow votes in different gauges in vote_proposal script

jalbrekt85:
handles #787 `gauge` can now be passed into `vote_proposal.py` with the value corresponding to `Gauges` enum


petrovska-petro (approved):
choices included looks good, increasing the flexibiliy of the script!

gosuto-inzasheru (approved):

Issue #799, opened 2022-09-06
Upgrade all v1.5 vaults on mainnet to include fee events


sajanrajdev:
EDIT: This logic is wrong, fixed below. Kept for reference.

Logic with fee events: https://etherscan.io/address/0x55918a84eb7fd1d930e098a8e86d0206549fa8c3#code

Diff: https://www.diffchecker.com/DsExMnb4

dapp-whisperer:
is this ready to close @sajanrajdev

sajanrajdev:
@dapp-whisperer, we have actually just spotted an issue in the newly introduced events so we will have to fix, redeploy and queue the upgrades again. I am working on the fixes, should be ready soon. We can disregard the already queued upgrades though.

sajanrajdev:
New fixed logic: https://etherscan.io/address/0x60c796acb2e0949178086294f03a44c450511784#code

New Diff: https://www.diffchecker.com/LcL04D3E

GalloDaSballo:
Code reviewed: https://github.com/Badger-Finance/badger-multisig/pull/830

sajanrajdev:
Execution posted at non 1080, events trace:
```
Events In This Transaction
--------------------------
├── Gravitationally Bound AURA (0xBA485b556399123261a5F9c95d413B4f93107407)
│   └── (unknown)
│       ├── topic1: 0xbc7cd75a20ee27fd9adebab32041f755214dbc6bffa90cc0225b39da2e5c2d3b
│       ├── topic2: 0x00000000000000000000000060c796acb2e0949178086294f03a44c450511784
│       └── data: 0x00
│
├── IGovernanceTimelock (0x21CF9b77F88Adf8F8C98d7E33Fe601DC57bC0893)
│   └── ExecuteTransaction
│       ├── txHash: 0x14c6427ad10bac71476ebf2170252de111beb44825b7740efc88a56d41b64274
│       ├── target: 0x20Dce41Acca85E8222D6861Aa6D23B6C941777bF
│       ├── value: 0
│       ├── signature: upgrade(address,address)
│       ├── data: 0x000000000000000000000000ba485b556399123261a5f9c95d413b4f9310740700000000000000000000000060c796acb2e0949178086294f03a44c450511784
│       └── eta: 1666256995
│
├── Badger Sett Aura BAL (0x37d9D2C6035b744849C15F1BFEE8F268a20fCBd8)
│   └── (unknown)
[...]

PR #802, opened 2022-09-06
[Vaults v1.5] Upgrade script to include fee events

sajanrajdev:
Tackles issue #799 

Queue with:
```
brownie run scripts/issue/799/upgrade_vaults_v1_5.py
```
Execute with:
```
brownie run scripts/issue/799/upgrade_vaults_v1_5.py main true
```
Simulate with:
```
brownie run scripts/issue/799/upgrade_vaults_v1_5.py main false true
```


scripts/issue/799/upgrade_vaults_v1_5.py
```diff
@@ -0,0 +1,175 @@
...
+    "0x794783dcfcac8c1944727057a3208d8f8bb91506"
+]
+
+
+def main(queue="true", simulation="false"):
+    safe = GreatApeSafe(registry.eth.badger_wallets.dev_multisig)
+    safe.init_badger()
+
+    for key, address in VAULTS.items():
+        address = web3.toChecksumAddress(address)
```
gosuto-inzasheru, line 39: is this needed? should already be taken care of by https://github.com/Badger-Finance/badger-multisig/blob/09d414664dae385de3890e09015dfb3afe27cda7/helpers/addresses.py#L1013

scripts/issue/799/upgrade_vaults_v1_5.py
```diff
@@ -0,0 +1,175 @@
...
+                    prev_gov = vault.governance()
+                    prev_keeper = vault.keeper()
+                    prev_guardian = vault.guardian()
+                    prev_strategist = vault.strategist()
+                    prev_treasury = vault.treasury()
+                    prev_token = vault.token()
+                    prev_symbol = vault.symbol()
+                    prev_balance = vault.balance()
+                    prev_totalSupply = vault.totalSupply()
+                    prev_getPricePerFullShare = vault.getPricePerFullShare()
```
gosuto-inzasheru, line 100: what is the reason writing all this out manually? instead of the programmatic approach such as https://github.com/Badger-Finance/badger-multisig/blob/09d414664dae385de3890e09015dfb3afe27cda7/scripts/issue/759/upgrade_uni_lp_geyser.py#L55-L61

scripts/issue/799/upgrade_vaults_v1_5.py
```diff
@@ -0,0 +1,175 @@
...
+                    key,
+                    address,
+                    simulation,
+                    safe
+                )
+        
+    safe.post_safe_tx()
+
+
+def execute_timelock(timelock, queueTx_dir, key, address, simulation, safe):
```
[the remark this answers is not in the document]
sajanrajdev: You are right. I based of this script from the script to upgrade all Setts to include GAC, which had a lot more off cases. This one doesn't have to be as overcomplicated, will fix.

scripts/issue/799/upgrade_vaults_v1_5.py
sajanrajdev: You are right

scripts/issue/799/upgrade_vaults_v1_5.py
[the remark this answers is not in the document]
sajanrajdev: The reason is that there are multiple vaults being upgraded and, hence, in order to check the storage of each, I need to fetch fetch the storage, execute individually and assert storage. Calling `execute_timelock` just calls execute on all files within the folder in a loop of its own so not ideal for batched upgrades.

gosuto-inzasheru (approved):
- successful local run, except for freeze on preview (gas passed from script to safe_tx_gas will be too low if using `skip_preview=True` to post!)
- not familiar with the solidity changes, but if the diff is indeed https://www.diffchecker.com/VqBX6bzH then all lgtm

petrovska-petro (approved):
- confirmed diff between contracts looks good and contains only the events changes pushed to the mix re perf and wd fees ✅ 

- sim runs smoothly and checkouts for all targeted setts ✅ 

- emits expected events in sim during harvest and withdrawal actions ✅ 

- sets delay 4d 

very excited for this upgrade!

dapp-whisperer (approved):

Issue #793, opened 2022-09-02
confirm assets to be cowswapped are actually present in the wallet


gosuto-inzasheru:
to prevent an issue we had with #788

a big red warning message in console should be enough, i think an assert would be too strong.

Issue #785, opened 2022-08-30
week 36: claim bribes from hh for graviaura emissions


gosuto-inzasheru:
prev #748 

should be able to be done onchain if pricer has indeed been updated

gosuto-inzasheru:
claim and sign swaps to weth:
```
Events In This Transaction
--------------------------
├── Stader (0x30D20208d987713f46DFD34EF128Bb16C404D10f)
│   └── Transfer
│       ├── from: 0x0b139682D5C9Df3e735063f46Fb98c689540Cf3A
│       ├── to: 0x3c0989eF27e3e3fAb87a2d7C38B35880C90E63b5
│       └── value: 65557073912584389948969
│   
├── RewardDistributor (0x0b139682D5C9Df3e735063f46Fb98c689540Cf3A)
│   ├── RewardClaimed
│   │   ├── identifier: 0x2dd1f383d6e6702cd6adab3dcb2c446af489332ef2e5897b957b888e2e3ccc4b
│   │   ├── token: 0x30D20208d987713f46DFD34EF128Bb16C404D10f
│   │   ├── account: 0x3c0989eF27e3e3fAb87a2d7C38B35880C90E63b5
│   │   ├── amount: 65557073912584389948969
│   │   └── updateCount: 5
│   └── RewardClaimed
│       ├── identifier: 0xf480fafe51da4bae2535242cdaca3898066e4dfc5967d1276811be881cd6d18d
│       ├── token: 0x9DDb2da7Dd76612e0df237B89AF2CF4413733212
│       ├── account: 0x3c0989eF27e3e3fAb87a2d7C38B35880C90E63b5
│       ├── amount: 262861900658457050
│       └── updateCount: 2
│   
├── Rocket Pool Protocol (0xD33526068D116cE69F19A9ee46F0bd304F21A51f)
│   └── Transfer
│       ├── from: 0x0b139682D5C9Df3e735063f46Fb98c689540Cf3A
│       ├── to: 0x3c0989eF27e3e3fAb87a2d7C38B35880C90E63b5
│       └── value: 625317724524310805296
│   
├── RewardDistributor (0x0b139682D5C9Df3e735063f46Fb98c689540Cf3A)
│   ├── RewardClaimed
│   │   ├── identifier: 0x329ffd9b41345bc4959db47e686a0416d414457adeaa66a22d5c5d79d6830e1b
[...]

PR #806, opened 2022-09-09
fix: confirm seller has enough assets before cowswapping

gosuto-inzasheru:
solves #793 

also some minor edits that were needed to let last hh claim pass successfully (#785)


great_ape_safe/ape_api/cow.py
```diff
@@ -79,26 +82,25 @@ def _sell(self, sell_token, mantissa_sell, buy_token,
...
+        fee_amount = int(fee_and_quote["fee"]["amount"])
         if mantissa_buy:
             # overwrite quote in case order has a limit
             mantissa_buy = int(mantissa_buy)
             buy_amount_after_fee = mantissa_buy
         else:
-            buy_amount_after_fee = int(int(fee_and_quote['buyAmountAfterFee']) * coef)
+            buy_amount_after_fee = int(int(fee_and_quote["buyAmountAfterFee"]) * coef)
             pricer = interface.IOnChainPricing(
                 interface.IBribesProcessor(registry.eth.cvx_bribes_processor).pricer(),
```
petrovska-petro, line 93: here will should use the pricer accordingly to each processor, currently it is hardcoded for cvx processor only

great_ape_safe/ape_api/cow.py
```diff
@@ -109,37 +111,44 @@ def _sell(self, sell_token, mantissa_sell, buy_token,
...
         }
-        print('ORDER PAYLOAD')
+        print("ORDER PAYLOAD")
         pprint(order_payload)
-        print('')
+        print("")
 
-        r = requests.post(f'{self.api_url}orders', json=order_payload)
+        # confirm seller actually has the assets that are to be swapped
+        if sell_token.balanceOf(origin) < mantissa_sell:
```
petrovska-petro, line 135: nice cli prompt for this situation!

great_ape_safe/ape_api/cow.py
gosuto-inzasheru: fixed in 6a03250

gosuto-inzasheru:
@saj can you have a look at 8568b3b L43-52?

gosuto-inzasheru:
(i know im sneaking in code outside the scope of this issue, but thought i might as well while we are updating the cow stuff)

scripts/badger/process_bribes_bvecvx.py
```diff
@@ -41,10 +41,15 @@ def step0_1(sim=False):
         claimed = {alcx.address: 500e18}
     else:
         claimed = SAFE.badger.claim_bribes_votium(registry.eth.bribe_tokens_claimable)
-        # Handling cvxFXS rewards as bribes (Ref: https://github.com/Badger-Finance/badger-strategies/issues/56)
-        cvxFXS_amount = SAFE.badger.sweep_reward_token(CVX_FXS)
-        if cvxFXS_amount > 0:
-            claimed[CVX_FXS] = claimed.setdefault(CVX_FXS, 0) + cvxFXS_amount
```
sajanrajdev, line 47: Pretty smart way to generalize the sweeping! I can confirm that the badger wrapper function will check if the balance for each token is 0 and, if this is the case, it will simply return 0 so the new logic should handle it properly.

Any new reward tokens emitted to vlCVX will simply need to be added to the bribe_tokens_claimable list. Downside of this is that the script will take a bit longer to run to potentially only handle one known token as everything else will be transferred on claim_bribes_votium. Not a bad tradeoff in favor of generalization!

sajanrajdev: LG!

scripts/badger/process_bribes_graviaura.py
```diff
@@ -47,6 +47,9 @@ def claim_and_sell_for_weth(claim_only=False):
     # NOTE: badger is directly emitted by the strat to tree
     # NOTE: aura is sent to processor, but should not be sold for weth
     for addr, mantissa in claimed.items():
+        if addr == "0x0":
```
sajanrajdev, line 50: Aren't ETH bribes shown as the `hhBribeVault`: https://etherscan.io/address/0x7c2a951d062cc7b7c6f9c7aa7e80a6f20eaa8cab#code#L2069?

According to the [rewards_distributor](https://etherscan.io/address/0x0b139682D5C9Df3e735063f46Fb98c689540Cf3A#readContract), it is: [0x9DDb2da7Dd76612e0df237B89AF2CF4413733212](https://etherscan.io/address/0x9DDb2da7Dd76612e0df237B89AF2CF4413733212)

petrovska-petro: i believe the endpoint which is the one we rely on for the claims here uses different convention for eth

great_ape_safe/ape_api/cow.py
petrovska-petro: v cool way of leveraging `locals()` 👍

petrovska-petro (approved):
adds improvements around processors:

- checks on prompt if origin has enough balance prior of posting order in cow ✅ 

- uses on each processor its own version of pricer for the quoting ✅ 

- adds new tokens which were recently process in aura ecosystem weth & rpl ✅ 

- generalised tokens to be swep from bvecvx strat, checkin if bal > 0 ✅

scripts/badger/process_bribes_graviaura.py
gosuto-inzasheru: yea the hhand.xyz api endpoint uses `"0x0"` for naked eth

petrovska-petro (approved):

scripts/badger/process_bribes_graviaura.py
sajanrajdev: Got it, I figured it could be the case but couldn't confirm myself. LGTM then.

sajanrajdev (approved):

Issue #823, opened 2022-09-19
Prepare migration script for capped gauge change


dapp-whisperer:
The proposed action set for each vault with a capped gauge:

```
* harvest
* withdraw
* setPID
* earn
```

naturally we'd want to confirm key variables after the migration.

after development, should be sent to WATCHPUG to review to validate the migration approach. Ideally this can be ready by EOW.

sajanrajdev:
Call trace:
```
Migrating gauge for Badger Sett 20WBTC-80BADGER
Transaction sent: 0xfecd1ad8013c56d7561291a1143511bc99639527868e39cfa66fda962ed69310
  Gas price: 0.0 gwei   Gas limit: 12000000   Nonce: 1
  IStrategyAuraStaking.harvest confirmed   Block: 15617815   Gas used: 793680 (6.61%)

Transaction sent: 0xaa2269f095b7e14908f1ad02ee8c1ab6b7add174da9d0102ab825149a4828284
  Gas price: 0.0 gwei   Gas limit: 12000000   Nonce: 2
  ITheVault.withdrawToVault confirmed   Block: 15617816   Gas used: 126286 (1.05%)

Transaction sent: 0x4d87fa4ed567db180904503af8e472b9d8bc8698cff6a1336a47580f3c9fae80
  Gas price: 0.0 gwei   Gas limit: 12000000   Nonce: 3
  IStrategyAuraStaking.setPid confirmed   Block: 15617817   Gas used: 54348 (0.45%)

Transaction sent: 0xa966ea15433b6d00d66df8a6208d35f2d2f30cc3539862132a6a34648f22a329
  Gas price: 0.0 gwei   Gas limit: 12000000   Nonce: 4
  ITheVault.earn confirmed   Block: 15617818   Gas used: 395271 (3.29%)

Migration successful!
Migrating gauge for Badger Sett 40WBTC-40DIGG-20graviAURA
Transaction sent: 0x93c40bbb90a590ff710a671ebeb132067b59f0054aacdaddfa17217798ae978c
  Gas price: 0.0 gwei   Gas limit: 12000000   Nonce: 5
  IStrategyAuraStaking.harvest confirmed   Block: 15617819   Gas used: 775235 (6.46%)

Transaction sent: 0xa8aa84b852b49fcc271e52d848819ac2c9c170662e1a2249c2d7c49d83de7ece
  Gas price: 0.0 gwei   Gas limit: 12000000   Nonce: 6
[...]

PR #824, opened 2022-09-21
Aura gauges migration script

sajanrajdev:
Tackles issue #823 

Script that migrates the underlying gauge of Aura strategies to a new one based on the Aura's PID for that gauge.

NOTE: The script will prevent the gauge migration if it finds the new base rewards pool to empty. As per @GalloDaSballo, "We need more than 1e18 ppfs to avoid a rebase of the ppfs over time".

Run with:
```
brownie run scripts/issue/823/migrate_strategy_gauges.py
```

Run simulation with:
```
brownie run scripts/issue/823/migrate_strategy_gauges.py main true
```

### DIFF
--- a/helpers/addresses.py
+++ b/helpers/addresses.py
@@ -316,7 +316,8 @@ ADDRESSES_ETH = {
         "native.bcrvBadger": "0x1905FD2D2D09792eE058C2b46a05F11630a1EcA1",
         "native.graviAURA": "0x3c0989eF27e3e3fAb87a2d7C38B35880C90E63b5",
         "native.bauraBal": "0x32BdF2B35Bbf8dAca3561C3ADE59c941488E9c69",
-        "native.b80BADGER-20WBTC": "0xe43857fE16D18b6633A663389934d6c64D5E81FD",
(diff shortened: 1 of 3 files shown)


dapp-whisperer (approved):
Looks sound to me, checks the variables I'd want to check and simulates the harvest after waiting some time.

gosuto-inzasheru (approved):
looks good saj, both harvest and withdrawall are successful on my end for both vaults.

GalloDaSballo:
Screenshot 2022-09-22 at 16 43 33

GalloDaSballo:
TODO: Test if wrong PID

GalloDaSballo:
Screenshot 2022-09-22 at 16 47 39


Screenshot 2022-09-22 at 16 47 24

I think the script handles it properly, last check is to make sure both gauges are non-zero which I believe has happened.

sajanrajdev:
If needed, the two migrations can be ran separately by commenting out one or the other at the moment of posting. This may be the case as their previous gauges will dry up at different times.

CC @GalloDaSballo @dapp-whisperer

PR #828, opened 2022-09-22
feat: add drippers for q4

gosuto-inzasheru:
deployed for q4. still need funding.


gosuto-inzasheru:
closes #818 and #819

shuklaayush:
Should the duration be 92 days instead of 91?

shuklaayush (approved):

jalbrekt85 (approved):

Issue #834, opened 2022-09-23
bug: `amount` grabbed from hidden hand api appears to have decimal issues


gosuto-inzasheru:
amount claimed from hidden hand in [this tx](https://etherscan.io/tx/0x3f42b8cc5b1125d2af4323a4b68b45d50d6c5fe17dfa9c15aede3076b4c75dc0) was `214.97151695` but only `21.49715169` was passed to the cowswap order. either amount grabbed from the api was wrong or something went wrong in parsing.

https://github.com/Badger-Finance/badger-multisig/blob/765e91ec1676322f6eee0ba0e701649cd01ac363/great_ape_safe/ape_api/badger.py#L159-L193

gosuto-inzasheru:
closed by #856

PR #835, opened 2022-09-23
chore: add a way to sell a specific token present in the processor

gosuto-inzasheru:
there was a bug (#834) for which this func was needed to solve


sajanrajdev (approved):
Simple enough, mimics the same process to sell bribes for ETH but takes a custom address as needed. It ensures that the token isn't AURA or BADGER to avoid reverts. The function was successfully used in prod already.

LGTM

jalbrekt85 (approved):
lgtm

Issue #841, opened 2022-09-29
Retire ops signer


Tritium-VLK:
Retire 0xcf4fF1e03830D692F52EB094c52A5A6A2181Ab3F as an operation wallet.  Stop eth topups.

gosuto-inzasheru:
address | chain | label
-- | -- | --
0x86cbD0ce0c087b482782c181dA8d191De18C8275 | eth | techops_multisig
0x6F76C6A1059093E21D8B1C13C4e20D8335e2909F | eth | politician_multisig
0x9faA327AAF1b564B569Cb0Bc0FDAA87052e8d92c | eth | recovered_multisig
0xD4868d98849a58F743787c77738D808376210292 | eth | ops_multisig
0x042B32Ac6b453485e357938bdC38e0340d4b9276 | eth | treasury_ops_multisig
0xB76782B51BFf9C27bA69C77027e20Abd92Bcf3a8 | eth | ibbtc_multisig
0x30a9c1D258F6c2D23005e6450E72bDD42C541105 | eth | payments_multisig
0x329543f0F4BB134A3f7a826DC32532398B38a3fA | bsc | dev_multisig
0xeb7341c89ba46CC7945f75Bd5dD7a04f8FA16327 | poly | techops_multisig
0x292549E6bd5a41aE4521Bb8679aDA59631B9eD4C | arbitrum | techops_multisig
0xb364bAb258ad35dd83c7dd4E8AC78676b7aa1e9F | arbitrum | dev_multisig
0x781E82D5D49042baB750efac91858cB65C6b0582 | ftm | techops_multisig
0xf109c50684EFa12d4dfBF501eD4858F25A4300B3 | ftm | treasury_ops_multisig

gosuto-inzasheru:
was removed from gas station, eth balance has been returned to ops.

now only needs to be replaced as a signer.

SHAKOTN:
https://arbiscan.io/tx/0x57f13fcf0fb9c90d1609a5f84f846322e68bfc4ee89dd0bf7d88e64b5fb59138

petrovska-petro:
https://etherscan.io/tx/0xb75549d7b4afd16680462ae8c3f557730e12350417fb60b08aec23f9d4fa7f78

petrovska-petro:
https://bscscan.com/tx/0xce52af54e1a378c2bffa0fa5cc6adb86576279641273788f01207e3171170515

petrovska-petro:
[...]

PR #848, opened 2022-10-03
chore: remove executor1 from address book and add new tree root validator

### LINKED ISSUE #844: Add new account as Root Validator on addresses.py
Need to add `0xcA63a374ec5f41dA402dD17b6d90DddFea3d64C4` as a new `ops_root-validator` to be included on gas refills as it will be used as a new badger tree approver

We can move the current `ops_root-validator-new` to deprecated as well

gosuto-inzasheru:
adjusting the address book with this pr and then running `brownie run badger/update_gas_station_watchlist` should close #844 and part of #841

@mitche50 how can we verify the eao address of the validator?


petrovska-petro:
getting merge #849 should be sufficient, which already has two reviews. perhaps should this be close?

gosuto-inzasheru:
yea you are right this is redundant

Issue #845, opened 2022-09-30
Add new root validator to badger tree


mitche50:
We need to sign the Validator role to `0xcA63a374ec5f41dA402dD17b6d90DddFea3d64C4` on the badger tree(`0x660802Fc641b154aBA66a62137e71f331B6d787A`) by calling:
```
grantRole(0x2650881bae229064abd4495ca56feba9ce572f8653a1aad517d81f9e3cf174c0, 0x660802Fc641b154aBA66a62137e71f331B6d787A)
```

gosuto-inzasheru:
note: has to be called from dev msig

gosuto-inzasheru:
```
Events In This Transaction
--------------------------
├── IBadgerTreeV2 (0x660802Fc641b154aBA66a62137e71f331B6d787A)
│   ├── RoleGranted
│   │   ├── role: 0x2650881bae229064abd4495ca56feba9ce572f8653a1aad517d81f9e3cf174c0
│   │   ├── account: 0xcA63a374ec5f41dA402dD17b6d90DddFea3d64C4
│   │   └── sender: 0xB65cef03b9B89f99517643226d76e286ee999e77
│   └── RoleRevoked
│       ├── role: 0x2650881bae229064abd4495ca56feba9ce572f8653a1aad517d81f9e3cf174c0
│       ├── account: 0x71e604d7db051eB7d9d0572dd06DbdB4435c9B1E
│       └── sender: 0xB65cef03b9B89f99517643226d76e286ee999e77
│   
└── Gnosis Safe (0xB65cef03b9B89f99517643226d76e286ee999e77)
    └── ExecutionSuccess
        ├── txHash: 0xb12c866950406a0c02c37adae0d57c3016d3be7f5d3769cba87a3962411e605a
        └── payment: 0
```
```
{
│   'ethereum_client': <gnosis.eth.ethereum_client.EthereumClient object at 0x11e761c10>,
│   'safe_address': '0xB65cef03b9B89f99517643226d76e286ee999e77',
│   'to': '0x40A2aCCbd92BCA938b02010E17A5b8929b49130D',
│   'value': 0,
[...]

PR #849, opened 2022-10-03
script: add new root validator to badger tree

gosuto-inzasheru:
@mitche50 i think there is a typo in your (pseudo) code in #845 

this pr should be correct.

uses address book from #848

### DIFF
--- a/helpers/addresses.py
+++ b/helpers/addresses.py
@@ -67,7 +67,6 @@ ADDRESSES_ETH = {
         "ops_deployer7": "0x7140B5254d80154f9Fc5F86054efB210f3a1e2c6",
         "ops_deployer8": "0x9082b0dD7A72c328833e6461965C9E91Cf59a960",
         "ops_deployer9": "0xd49177c4cF4b8cCdf58475c9a155C4CD8b9F3F31",
-        "ops_executor1": "0xcf4fF1e03830D692F52EB094c52A5A6A2181Ab3F",
         # "ops_executor2": "0x8938bf50d1a3736bdA413510688834540858dAEA",  # can execute from deployer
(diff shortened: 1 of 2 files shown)


sajanrajdev (dismissed):
Confirmed that the pseudocode in the issue is wrong - the address to pass to the grantRole functions should be the new validator's. Hence, this PR is correct and runs properly.

image

Script LGTM

NOTE: @mitche50, the Tree still has VALIDATOR ROLE granted to a few old accounts:
[0x626F69162Ea1556A75Dd4443D87D2fe38dd25901](https://etherscan.io/address/0x626F69162Ea1556A75Dd4443D87D2fe38dd25901)
[0x71e604d7db051eB7d9d0572dd06DbdB4435c9B1E](https://etherscan.io/address/0x71e604d7db051eB7d9d0572dd06DbdB4435c9B1E)
[0x1318d5c0C24830D86Cc27Db13Ced0CED31412438](https://etherscan.io/address/0x1318d5c0C24830D86Cc27Db13Ced0CED31412438)

Should we also remove the ROLE from these?

petrovska-petro (dismissed):
Script adds correctly a new validator, addr of `ops_root-validator_v3` confirmed ownership internally ✅

gosuto-inzasheru:
waiting for response from @mitche50 on removing old validator roles before merging.

mitche50:
[0x71e604d7db051eB7d9d0572dd06DbdB4435c9B1E](https://etherscan.io/address/0x71e604d7db051eB7d9d0572dd06DbdB4435c9B1E) I believe can be removed, the other two should be left as backups in case the new account owner cannot perform approvals.

gosuto-inzasheru:
added call to `renounceRole`

Issue #833, opened 2022-09-23
week 40: claim bribes from hh for graviaura emissions


gosuto-inzasheru:
prev #805

sajanrajdev:
Claim bribes and swap for wETH: 
```
Tx Hash: 0xde52a9749a38eab29c9233263065bdc66ace68a8f38ee36c3e6ecd969af45753
From: 0x0a9af7FAba0d5DF7A8C881e1B9cd679ee07Af8A2
To: 0x86cbD0ce0c087b482782c181dA8d191De18C8275
Value: 0
Function: Gnosis Safe.execTransaction
Block: 15678512
Gas Used: 1849805 / 2086125 (88.7%)

Events In This Transaction
--------------------------
├── RewardDistributor (0x0b139682D5C9Df3e735063f46Fb98c689540Cf3A)
│   └── RewardClaimed
│       ├── identifier: 0x8eee07e6aecfd58a059e9c795e092747309d76151e2ab9562faf60b2d64673fd
│       ├── token: 0xC0c293ce456fF0ED870ADd98a0828Dd4d2903DBF
│       ├── account: 0x3c0989eF27e3e3fAb87a2d7C38B35880C90E63b5
│       ├── amount: 297270305214029220240
│       └── updateCount: 7
│
├── USD Coin (0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48)
│   └── Transfer
│       ├── from: 0x0b139682D5C9Df3e735063f46Fb98c689540Cf3A
│       ├── to: 0x3c0989eF27e3e3fAb87a2d7C38B35880C90E63b5
│       └── value: 8169850361
│
├── RewardDistributor (0x0b139682D5C9Df3e735063f46Fb98c689540Cf3A)
│   ├── RewardClaimed
│   │   ├── identifier: 0x3f5a22cb83118980ed5887bc9ed2b8405706479ce22435bdf0b199c2192702ac
│   │   ├── token: 0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48
│   │   ├── account: 0x3c0989eF27e3e3fAb87a2d7C38B35880C90E63b5
│   │   ├── amount: 8169850361
│   │   └── updateCount: 8
│   ├── RewardClaimed
│   │   ├── identifier: 0xf480fafe51da4bae2535242cdaca3898066e4dfc5967d1276811be881cd6d18d
[...]

PR #856, opened 2022-10-04
fix: bugs in processing script for graviaura bribes

gosuto-inzasheru:


gosuto-inzasheru:
i think this was used by @sajanrajdev in production yesterday in order to solve #833

sajanrajdev (approved):
Hotfixes tested and working in production:
- Function to add 0 to claimable amounts with missing decimals before converting to wei
- Fetching pricer from explorer instead of using interface (contract instance was breaking). This is an idea regardless because different processors may use different pricers with different signatures.
LGTM

petrovska-petro (approved):
Nice usage of *ljust* to fill up till decimal token place and addresses added in book are correct ✅

Issue #853, opened 2022-10-04
increase pricer's slippage to 5%


gosuto-inzasheru:
had a revert last round and also today due to the coefficient passed to the order (.9825) being so close to the slippage the pricer allows (2%)

gosuto-inzasheru:
```
{
│   'ethereum_client': <gnosis.eth.ethereum_client.EthereumClient object at 0x125fe2f40>,
│   'safe_address': '0x86cbD0ce0c087b482782c181dA8d191De18C8275',
│   'to': '0x40A2aCCbd92BCA938b02010E17A5b8929b49130D',
│   'value': 0,
│   'data': HexBytes('0x8d80ff0a000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000f200bab7f98d62479309219b2cc5026d4ad1c6c0567400000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000024f0fa55a900000000000000000000000000000000000000000000000000000000000001f3002dc7693444acd1eca1d6de5b3d0d8584f3870c4900000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000024f0fa55a900000000000000000000000000000000000000000000000000000000000001f30000000000000000000000000000'),
│   'operation': 1,
│   'safe_tx_gas': 103187,
│   'base_gas': 0,
│   'gas_price': 0,
│   'gas_token': '0x0000000000000000000000000000000000000000',
│   'refund_receiver': '0x0000000000000000000000000000000000000000',
│   'signatures': b'',
│   '_safe_nonce': 374,
│   '_safe_version': '1.1.1',
│   '_chain_id': None
}
```

PR #865, opened 2022-10-06
script: increase slippage on both bvecvx and gravi pricers to max

gosuto-inzasheru:
script for #853


petrovska-petro (approved):
bumps up both pricer slippage to max value (< 500 as require) ✅

sajanrajdev (approved):
Sets slippage on both CVX and Aura pricers to 4.99%. Confirmed that 5% is indeed the non-inclusive limit:
image

Script runs properly:
image

LGTM

Issue #859, opened 2022-10-05
supply 100k $badger onto euler money market


gosuto-inzasheru:
- [x] add euler platform to our ape api with (for now) support for [deposit](https://docs.euler.finance/developers/getting-started/contract-reference#deposit) and [withdraw](https://docs.euler.finance/developers/getting-started/contract-reference#withdraw).
- [x] deposit 100k $badger into its [market](https://app.euler.finance/market/0x3472a5a71965499acd81997a54bba8d852c6e53d). treasury decision to do so here: https://discord.com/channels/1011322803414388897/1011333638044008548/1026922185505382450

gosuto-inzasheru:
```
Events In This Transaction
--------------------------
├── Euler (0x27182842E098f60e3D576794A5bFFb0777E025d3)
│   ├── RequestDeposit
│   │   ├── account: 0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e
│   │   └── amount: 100000000000000000000000
│   └── Deposit
│       ├── underlying: 0x3472A5A71965499acd81997a54BBA8D852C6E53d
│       ├── account: 0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e
│       └── amount: 100000000000000000000000
│   
├── Euler Pool: Badger (0x7491989cFBC6Da74141BC8cd187E480c21ECE169)
│   └── Transfer
│       ├── from: 0x0000000000000000000000000000000000000000
│       ├── to: 0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e
│       └── value: 100000000000000000000000
│   
├── Euler (0x27182842E098f60e3D576794A5bFFb0777E025d3)
│   └── AssetStatus
│       ├── underlying: 0x3472A5A71965499acd81997a54BBA8D852C6E53d
│       ├── totalBalances: 100005000000000001203293
│       ├── totalBorrows: 0
│       ├── reserveBalance: 1202941
│       ├── poolSize: 100005000000000001203293
│       ├── interestAccumulator: 1000000000000000000000000000
│       ├── interestRate: 0
│       └── timestamp: 1665409271
│   
└── Gnosis Safe (0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e)
    └── ExecutionSuccess
        ├── txHash: 0x1778b9982f4568dc050acc2f4d73d74041d99c2b1f2af269747fe904c7c04c2b
        └── payment: 0
```
```
{
[...]

PR #868, opened 2022-10-07
euler addition to great ape safe + deposit script

jalbrekt85:
handles #859 

```
brownie run issue/859/euler_deposit
```


petrovska-petro (commented):
Market was deployed by one of our wallets, [tx](https://etherscan.io/tx/0x25a26518a34255a5c11357644d9874167dc8268b356e29190134fe55af99936a). [ebadger](https://etherscan.io/address/0x7491989cFBC6Da74141BC8cd187E480c21ECE169#contracts) is not verified tho atm, but deployed thru the euler's market contract. Is it possible @gosuto-inzasheru that you can verify it?

- addressed added into the book matches with their [dev docs](https://docs.euler.finance/euler-protocol/addresses) ✅ 

- actions for deposit looks good ✅ 

```
├── Euler Pool: Badger (0x7491989cFBC6Da74141BC8cd187E480c21ECE169)
│   └── Transfer
│       ├── from: 0x0000000000000000000000000000000000000000
│       ├── to: 0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e
│       └── value: 100000000000000000000000
```

Note: ref regarding the concept of sub-account [doc](https://docs.euler.finance/getting-started/white-paper#sub-accounts) and why in the class is used *0* as primary

great_ape_safe/ape_api/euler.py
```diff
@@ -0,0 +1,48 @@
...
+        etoken.withdraw(0, amount)
+
+        after_etoken = etoken.balanceOf(self.safe)
+        after_underlying = underlying.balanceOf(self.safe)
+
+        assert after_etoken < before_etoken
+        assert after_underlying > before_underlying + amount
+
+    def withdraw_all(self, underlying):
+        self.withdraw(underlying, self.safe.balanceOf(underlying))
```
petrovska-petro, line 48: here is missing referencing the *etoken* for the wd of the whole balance, instead is trying from the safe which does not exist

```suggestion
        etoken = interface.IEToken(
            self.markets.underlyingToEToken(underlying), owner=self.safe.account
        )
        self.withdraw(underlying, etoken.balanceOf(underlying))
```

petrovska-petro, line 48: i think here the intention is to read the bal of the underlying
```
        self.withdraw(underlying, underlying.balanceOf(self.safe))
```

but it may happen that the underlying balance is way more than the etoken balance in the *etoken*. i think it is worthy perhaps to have a flag in the *withdraw* method which if *True* just to pass as amount *etoken.balanceOf(self.safe)* instead, which should remove the whole position

jalbrekt85: actually no need to query etoken balance in`withdraw_all`, `type(uint).max` ([ref](https://docs.euler.finance/developers/getting-started/contract-reference#withdraw)) can just be passed

jalbrekt85: 5906ba5 (also fixes faulty withdraw assert)

scripts/issue/859/euler_deposit.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def main():
+    vault = GreatApeSafe(r.badger_wallets.treasury_vault_multisig)
+    vault.init_euler()
+    
+    badger = vault.contract(r.treasury_tokens.BADGER)
+    ebadger = interface.IERC20(
+            vault.euler.markets.underlyingToEToken(badger)
+        )
```
gosuto-inzasheru, line 16:
```suggestion
    ebadger = interface.IEToken(
            vault.euler.markets.underlyingToEToken(badger)
        )
```

scripts/issue/859/euler_deposit.py
```diff
@@ -0,0 +1,32 @@
...
+
+    before_ebadger = ebadger.balanceOf(vault) / 1e18
+
+    vault.take_snapshot(tokens=[badger])
```
gosuto-inzasheru, line 20:
```suggestion
    vault.take_snapshot(tokens=[badger, ebadger])
```

gosuto-inzasheru (commented):
minor suggestions, rest looks good.

scripts/issue/859/euler_deposit.py
jalbrekt85: 4fadd7a

gosuto-inzasheru (approved):

great_ape_safe/ape_api/euler.py
```diff
@@ -0,0 +1,49 @@
...
+        etoken.withdraw(0, amount)
+
+        after_etoken = etoken.balanceOf(self.safe)
+        after_underlying = underlying.balanceOf(self.safe)
+
+        assert after_etoken < before_etoken
+        assert after_underlying > before_underlying
+
+    def withdraw_all(self, underlying):
+        self.withdraw(underlying, MaxUint256)
```
petrovska-petro, line 49: nice, confirming the logic: [ref](https://github.com/euler-xyz/euler-contracts/blob/cd3036e0087280365819f99ad531141894d0b7ee/contracts/BaseLogic.sol#L394)

petrovska-petro (approved):

PR #872, opened 2022-10-12
chore: provide correct interfaces for tokens using `MintableForkToken`

gosuto-inzasheru:
`brownie test` doesnt run flawless yet but this is a big improvement already


jalbrekt85 (approved):
nice to have your `generate_interface.py` in the repo, new interface lgtm

Issue #873, opened 2022-10-13
week 41: bribe in HH and Convex


petrovska-petro:
From council consensus: 

![Screenshot 2022-10-13 at 15 46 38](https://user-images.githubusercontent.com/84875062/195614463-cb874ec9-185f-432a-a4a4-1e7ddcb756ae.png)

tldr: split 50:50 between bal and convex market, 8k bribe in each ecosystem. targets:

- 80badger/20wbtc in balancer

- badger/fraxbp in convex

petrovska-petro:
```

**Comment by petrovska-petro** [comment]
bribe in hh aura market:

```
Proposal index: 25
Choice: 30
Proposal hash:  0x45768a1322d624d68f3a4861f29d212d4f45fa493b346276b7199d716f4cb86f
```

```
Events In This Transaction
--------------------------
├── IBribeVault (0x9DDb2da7Dd76612e0df237B89AF2CF4413733212)
│   └── DepositBribe
│       ├── bribeIdentifier: 0x982f52be05303b1e164097477ad460431ed849d11050a3a87e465d50b0c3d548
│       ├── rewardIdentifier: 0x375925b16a8a2bdd6450c2444e49cf243d2803d09b4c185c6d8dac205802e056
│       ├── token: 0x3472A5A71965499acd81997a54BBA8D852C6E53d
│       ├── amount: 8000000000000000000000
│       ├── totalAmount: 8000000000000000000000
│       └── briber: 0x042B32Ac6b453485e357938bdC38e0340d4b9276
│   
├── IAuraBribe (0x642c59937A62cf7dc92F70Fd78A13cEe0aa2Bd9c)
│   └── DepositBribe
│       ├── proposal: 0x45768a1322d624d68f3a4861f29d212d4f45fa493b346276b7199d716f4cb86f
│       ├── token: 0x3472A5A71965499acd81997a54BBA8D852C6E53d
│       ├── amount: 8000000000000000000000
│       ├── bribeIdentifier: 0x982f52be05303b1e164097477ad460431ed849d11050a3a87e465d50b0c3d548
│       ├── rewardIdentifier: 0x375925b16a8a2bdd6450c2444e49cf243d2803d09b4c185c6d8dac205802e056
│       └── briber: 0x042B32Ac6b453485e357938bdC38e0340d4b9276
│   
└── Gnosis Safe (0x042B32Ac6b453485e357938bdC38e0340d4b9276)
    └── ExecutionSuccess
[...]

Issue #847, opened 2022-10-03
possible bug: `proposals` in bribe_hh.py is not complete 


gosuto-inzasheru:
https://github.com/Badger-Finance/badger-multisig/blob/0651b7c1a3536fab76a2dc94b7a63f8deacc6f79/scripts/badger/bribe_hh.py#L69-L74

on the last bribe round i had to add `+ 5` to `proposal_index` in order to generate the correct `prop`. not sure if this is due to gauge migrations, a fluke or something else.

petrovska-petro:
seems that in total rn there are so far 20 proposals in length in aura snap space, but somehow when they initiated this proposal they summed up 5, so i guess to adapt we need that `+5` for now to make it work

latest init [tx](https://etherscan.io/tx/0x65ff41fa88a42832b2e223ac0aac00158126c606c5c1ad25edcb7dc6f210b421) from hh

petrovska-petro:
ended up finding the cause, will push solution in pr

gosuto-inzasheru:
fixed by #874

PR #874, opened 2022-10-13
Update bribing script to contain all ecosystems interested on and fix proposal index

petrovska-petro:
tackles #873 & #847

still waiting on the following [question](https://discord.com/channels/866037423325446165/866039411258425385/1030126781585424424). somehow i believe that nobody bribe just yet cause there is a minor issue with that on the way is set, but i may be wrong. if you know the right approach, please lmk

### DIFF
--- a/helpers/addresses.py
+++ b/helpers/addresses.py
@@ -456,6 +456,7 @@ ADDRESSES_ETH = {
         },
     },
     "votium": {
+        "bribe": "0x19BBC3463Dd8d07f55438014b021Fb457EBD4595",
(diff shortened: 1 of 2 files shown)


petrovska-petro:
for those interested in the answer and are not in the discord:

Screenshot 2022-10-14 at 12 53 30

jalbrekt85 (approved):
script refactored to deposit in all applicable bribe markets. `get_index` function looks good and `web3.keccak` is used in votium to generate correct proposal hash

gosuto-inzasheru (approved):
successful run locally

Issue #878, opened 2022-10-14
migrade chainlink keepers from q3 to q4 drippers


gosuto-inzasheru:
see #569 for similar issue from last quarter.

gosuto-inzasheru:
first step; cancel current upkeeps. wait 50 blocks after this before step 2 (otherwise $link cannot be recovered yet).
```
Events In This Transaction
--------------------------
├── IKeeperRegistry (0x7b3EC232b08BD7b4b3305BE0C044D907B2DF960B)
│   ├── UpkeepCanceled
│   │   ├── id: 127
│   │   └── atBlockHeight: 15781243
│   └── UpkeepCanceled
│       ├── id: 128
│       └── atBlockHeight: 15781243
│   
└── Gnosis Safe (0x86cbD0ce0c087b482782c181dA8d191De18C8275)
    └── ExecutionSuccess
        ├── txHash: 0x3ac0f595e63bed50ba143adf429873d5fd63983c164fda00b62673d799b3c1ed
        └── payment: 0
```
```
{
│   'ethereum_client': <gnosis.eth.ethereum_client.EthereumClient object at 0x10ff1f6a0>,
│   'safe_address': '0x86cbD0ce0c087b482782c181dA8d191De18C8275',
│   'to': '0x40A2aCCbd92BCA938b02010E17A5b8929b49130D',
│   'value': 0,
│   'data': HexBytes('0x8d80ff0a000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000f2007b3ec232b08bd7b4b3305be0c044d907b2df960b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000024c8048022000000000000000000000000000000000000000000000000000000000000007f007b3ec232b08bd7b4b3305be0c044d907b2df960b00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000024c804802200000000000000000000000000000000000000000000000000000000000000800000000000000000000000000000'),
│   'operation': 1,
│   'safe_tx_gas': 189854,
│   'base_gas': 0,
│   'gas_price': 0,
│   'gas_token': '0x0000000000000000000000000000000000000000',
│   'refund_receiver': '0x0000000000000000000000000000000000000000',
│   'signatures': b'',
│   '_safe_nonce': 380,
│   '_safe_version': '1.1.1',
│   '_chain_id': None
}
```

[...]

PR #879, opened 2022-10-14
script: migrate keepers from q3 to q4 drippers

gosuto-inzasheru:
solution to #878

### DIFF
--- a/scripts/issue/878/migrate_drippers_q3_q4.py
+++ b/scripts/issue/878/migrate_drippers_q3_q4.py
@@ -1,0 +1,60 @@
+"""
+ref: https://github.com/smartcontractkit/keeper/blob/master/contracts/UpkeepRegistrationRequests.sol
+"""
+
+from brownie import chain, interface
+
+from great_ape_safe import GreatApeSafe
+from helpers.addresses import r
+
+
+LINK_MANTISSA = 50e18
+UPKEEP_ID_REMBADGER_22Q3 = 127  # https://keepers.chain.link/mainnet/127
+UPKEEP_ID_TREE_22Q3 = 128  # https://keepers.chain.link/mainnet/128
+
+SAFE = GreatApeSafe(r.badger_wallets.techops_multisig)
+RELAYER = SAFE.contract(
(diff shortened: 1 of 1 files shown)


petrovska-petro (approved):
migrations looks good:

- cancels upkeeps 127&128 which are from last quarter ✅ 

- retrieve link balances from those cancel ✅ 

- sets gas limit to 300k as prev jobs ✅ 

- register new drippers and top them up with link balance ✅

scripts/issue/878/migrate_drippers_q3_q4.py
```diff
@@ -0,0 +1,60 @@
...
+    SAFE.chainlink.register_upkeep(
+        "RemBadgerDripper2022Q4", r.drippers.rembadger_2022_q4, 300_000, LINK_MANTISSA
+    )
+
+    # update watchlist on the gas station
+    SAFE.badger.set_gas_station_watchlist()
+
+    if not sim:
+        SAFE.post_safe_tx(call_trace=True)
+    else:
```
sajanrajdev, line 59: Just out of curiosity, why do we include a posting when simulating?

sajanrajdev: Nvm, I guess it is so that the snapshot print, got it.

gosuto-inzasheru: well and sometimes it is good to have `.preview` run and see it run successfully.

sajanrajdev (approved):
- Cancels previous upkeeps and withdraws their outstanding LINK balance
- Registers new upkeeps with the same gas limit as the previous ones and tops them up with 50 LINK
- Simulation runs successfully:
```
snapshot result for 0x86cbD0ce0c087b482782c181dA8d191De18C8275:
               balance_before          balance_after          balance_delta
symbol
LINK    64.326337510456059545  89.878640199184261734  25.552302688728202189
```

LINK balance sheet LG: 62.8 + 62.7 + 64.3 + (50) + (50) = 89.8

Given that only ~12 LINK were spent for the previous quarter, reducing the amounts to use should be good.

LGTM

Issue #626, opened 2022-07-04
week 44: voter msig chores


gosuto-inzasheru:
first iteration run at #895, was opened for tackling week 42

gosuto-inzasheru:
hmm not sure if this is still the way we want to go now. @Tritium-VLK?

petrovska-petro:
module checkup for auto processing expired locks - autorelocker (keeper-ize)

gosuto-inzasheru:
list of all actions that need to happen periodically on the voter:
- claim hh bribes
- claim vlaura rewards
- relock vlaura
- withdraw graviaura
- lock aura
- sweep remaining erc20s to trops

gosuto-inzasheru:
lets run this week after #908 and #889 have been execd to maximise lockable (gravi)aura.

petrovska-petro:
```
snapshot result for 0xA9ed98B5Fb8428d68664f3C5027c62A10d45826b:
                     balance_before         balance_after              balance_delta
symbol                                                                              
auraBAL        0.552030578538825728  0.003696524445213861      -0.548334054093611867
graviAURA  7,045.426373041353371013  0.000000000000000000  -7,045.426373041353371013 
```

```
Events In This Transaction
--------------------------
├── Badger (0x3472A5A71965499acd81997a54BBA8D852C6E53d)
│   └── Transfer
│       ├── _from: 0xA9ed98B5Fb8428d68664f3C5027c62A10d45826b
│       ├── _to: 0x042B32Ac6b453485e357938bdC38e0340d4b9276
│       └── _amount: 160096180590721982539
│   
├── Vote Locked Aura (0x3Fa73f1E5d8A792C80F426fc8F84FBF7Ce9bBCAC)
│   ├── DelegateCheckpointed
│   │   └── delegate: 0xA9ed98B5Fb8428d68664f3C5027c62A10d45826b
│   ├── Withdrawn
│   │   ├── _user: 0xA9ed98B5Fb8428d68664f3C5027c62A10d45826b
│   │   ├── _amount: 15945803103329618623482
│   │   └── _relocked: True
│   ├── DelegateCheckpointed
│   │   └── delegate: 0xA9ed98B5Fb8428d68664f3C5027c62A10d45826b
│   ├── Staked
│   │   ├── _user: 0xA9ed98B5Fb8428d68664f3C5027c62A10d45826b
│   │   ├── _paidAmount: 15945803103329618623482
[...]

PR #885, opened 2022-10-18
Add relocker script for voter_msig

petrovska-petro:
Tackle the vlaura which gets over time unlock in the voter_msig, part of the convo at #626 

run:

```
brownie run badger/voter_processing_unlocks

brownie run badger/voter_chores
```

**Note:** opted for leaving in also the script which simply process the expired locks in case that fits on the timing, and we do need necessarily to trigger the big voter chore. imo is handy

Update on voter chores script steams from the convo at this [thread](https://discord.com/channels/1011322803414388897/1032235652307230771)


scripts/badger/voter_chores.py
```diff
@@ -0,0 +1,52 @@
...
+    GRAVI = voter.contract(r.sett_vaults.graviAURA)
+    AURA = voter.contract(r.treasury_tokens.AURA)
+
+    data = voter.badger.get_hh_data()
+    rewards = [x["token"] for x in data]
+
+    voter.take_snapshot(tokens=rewards)
+    destination.take_snapshot(tokens=rewards)
+
+    voter.badger.claim_bribes_hidden_hands(claim_from_strat=False)
```
gosuto-inzasheru, line 21: this reverts on `VirtualMachineError: revert: Invalid _claims`. maybe add a check here as well?

scripts/badger/voter_chores.py
```diff
@@ -0,0 +1,52 @@
...
+    # add relock chore
+    _, unlockable, _, _ = vlAURA.lockedBalances(voter)
+    if unlockable > 0:
+        vlAURA.processExpiredLocks(True)
+
+        _, unlockable, _, _ = vlAURA.lockedBalances(voter)
+        assert unlockable == 0
+
+    # claim rewards and send to trops
+    vlAURA.getReward(voter)
```
gosuto-inzasheru, line 40: this isnt being tracked by the safe snapshot

petrovska-petro: added in the snap in 5d1f7e4e50ba4dbb2ef9b88b8e7e945d690f88a1

scripts/badger/voter_chores.py
petrovska-petro: include the check to see if there is indeed some data to claim for in 5d1f7e4e50ba4dbb2ef9b88b8e7e945d690f88a1

scripts/badger/voter_chores.py
```diff
@@ -0,0 +1,59 @@
...
+    AURABAL.transfer(destination, AURABAL.balanceOf(voter))
+
+    # wd from gravi and lock aura. gravi is coming as fee os processing bribes
+    gravi_balance = GRAVI.balanceOf(voter)
+    if gravi_balance > 0:
+        GRAVI.withdrawAll()
```
gosuto-inzasheru, line 53: there is not enough liq in the vault to withdraw the current amount. kinda nasty, but i think we should do like a 75% withdrawal of the vault's balance in that case instead.

too bad there isnt a withdrawAsMuchAsPossible()...

scripts/badger/voter_processing_unlocks.py
```diff
@@ -0,0 +1,20 @@
+from great_ape_safe import GreatApeSafe
+from helpers.addresses import r
+
+
+def main():
```
gosuto-inzasheru, line 5: lets incorp this into `voter_chores` as well as to not have redundant code

petrovska-petro: resolved in bef27e9818c2f9d02ff5892f422df72e78f50250

scripts/badger/voter_chores.py
petrovska-petro: push a safer approach in 7914726d9854af03e8c65ec75afa220564a0c462

gosuto-inzasheru (approved):
runs good on my local.

`WD_PROTECTION` might be a bit tight, but we can always adjust to `.9` or whatever when running in production

jalbrekt85 (approved):
covers:
- claim hh bribes
- claim vlaura rewards
- relock vlaura
- withdraw graviaura
- lock aura
- sweep remaining erc20s to trops

lgtm

Issue #888, opened 2022-10-19
Treasury Decision: Rebalance WBTC/WETH Ratio


jalbrekt85:
> [unwind treasury] ibBTC position into WBTC, and swap exactly 30 WBTC for WETH at the given market rate

[full proposal](https://docs.google.com/document/d/1Tq58AUW3GwjVovVTU_QRv1DVBgQnbZTaLIZ2Snz1b6Q)

gosuto-inzasheru:
note prop was changed to 40 instead of 30 wbtc: https://discord.com/channels/1011322803414388897/1032192022532067363/1033091802137378836

gosuto-inzasheru:
```
Events In This Transaction
--------------------------
├── Badger Sett Curve.fi Factory BTC Metapool: ibBTC (0xaE96fF08771a109dc6650a1BdCa62F2d558E40af)
│   └── Transfer
│       ├── from: 0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e
│       ├── to: 0x0000000000000000000000000000000000000000
│       └── value: 33303585245672336694
│   
├── VirtualBalanceRewardPool (0xAAf75a94394F6D06E01CcE62e2545CeFFBFA1E2D)
│   └── Withdrawn
│       ├── user: 0x6D4BA00Fd7BB73b5aa5b3D6180c6f1B0c89f70D1
│       └── amount: 33222255369191370785
│   
├── Curve.fi Factory BTC Metapool: ibBTC Convex Deposit (0xb43ed35B5A3a9544BBEd8039c67AB04AD428deEa)
│   └── Transfer
│       ├── from: 0x4F2b8a15d0Dd58c1eB60bd53e966872828519Cee
│       ├── to: 0x0000000000000000000000000000000000000000
│       └── value: 33222255369191370785
│   
├── Curve.fi ibbtc/sbtcCRV-f Gauge Deposit (0x346C7BB1A7a6A30c8e81c14e90FC2f0FBddc54d8)
│   └── UpdateLiquidityLimit
│       ├── user: 0x989AEb4d175e16225E39E87d0D97A3360524AD80
│       ├── original_balance: 1140496849832863208911
│       ├── original_supply: 1140496849832863208911
│       ├── working_balance: 843252568978719920781
│       └── working_supply: 843252568978719920781
│   
[...]

PR #892, opened 2022-10-21
sell wbtc for weth script

jalbrekt85:
handles #888, now sells 40 wbtc ~submitting as draft - currently 38 naked wbtc that can be used for swap, but not sure if more wbtc should be siphoned from other sources in this pr for other TCL activities as mentioned in proposal~

### DIFF
--- a/scripts/issue/888/rebalance_wbtc_weth.py
+++ b/scripts/issue/888/rebalance_wbtc_weth.py
@@ -1,0 +1,17 @@
+from great_ape_safe import GreatApeSafe
+from helpers.addresses import r
+from rich.console import Console
+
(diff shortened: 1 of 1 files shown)


gosuto-inzasheru:
yes we also want to take the opportunity here to fully unwind the ibbtc position.

gosuto-inzasheru (approved):
perfecto

- full withdrawal to wbtc
- swap exact 40 wbtc to weth

petrovska-petro (approved):
looking smooth, un-dogfood from `ibbtc set` and wd everything thru zapper to wbtc and swaps 40wbtc to weth via cow ✅

PR #922, opened 2022-11-01
Register emission control v2

gosuto-inzasheru:
replaces #921 

- add arbitrum emission control address to address book
- commit in #921 is unverified

PR #921, opened 2022-11-01
chore: add emission control registration

axejintao:
# Summary

- register emission control contracts

Issue #792, opened 2022-09-01
register emission control on eth, arbitrum


axejintao:
# Summary

- emission control is used to manage boost weights currently
- add key `emissionControl` to the badger registry v2

gosuto-inzasheru:
posted to mainnet:
```
Events In This Transaction
--------------------------
├── IBadgerRegistryV2 (0xdc602965F3e5f1e7BAf2446d5564b407d5113A06)
│   ├── AddKey
│   │   └── key: emissionControl
│   └── Set
│       ├── key: emissionControl
│       └── at: 0x31825c0A6278b89338970e3eB979b05B27FAa263
│   
└── Gnosis Safe (0x86cbD0ce0c087b482782c181dA8d191De18C8275)
    └── ExecutionSuccess
        ├── txHash: 0xd0c6b347b961ff500420528f2b064736e717c5b06f5dae1b464bd36929846b84
        └── payment: 0
```
```
{
│   'ethereum_client': <gnosis.eth.ethereum_client.EthereumClient object at 0x10936bbb0>,
│   'safe_address': '0x86cbD0ce0c087b482782c181dA8d191De18C8275',
│   'to': '0x40A2aCCbd92BCA938b02010E17A5b8929b49130D',
│   'value': 0,
│   'data': HexBytes('0x8d80ff0a000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000d900dc602965f3e5f1e7baf2446d5564b407d5113a0600000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000084a815ff15000000000000000000000000000000000000000000000000000000000000004000000000000000000000000031825c0a6278b89338970e3eb979b05b27faa263000000000000000000000000000000000000000000000000000000000000000f656d697373696f6e436f6e74726f6c000000000000000000000000000000000000000000000000'),
│   'operation': 1,
│   'safe_tx_gas': 353456,
│   'base_gas': 0,
│   'gas_price': 0,
│   'gas_token': '0x0000000000000000000000000000000000000000',
│   'refund_receiver': '0x0000000000000000000000000000000000000000',
│   'signatures': b'',
│   '_safe_nonce': 392,
│   '_safe_version': '1.1.1',
│   '_chain_id': None
}
```

gosuto-inzasheru:
and on arb1:
```
Events In This Transaction
--------------------------
[...]

PR #924, opened 2022-11-01
fix: pr #921 #922 but now with signed commits only

gosuto-inzasheru:
closes #921 
closes #922 
solves #792


sajanrajdev (approved):
Registers the Emission Control contract on the registry for mainnet and arbitrum. Runs properly

LGTM

petrovska-petro (approved):
runs smoothly in mainnet and arbitrum to register emission control ✅

PR #932, opened 2022-11-04
Feat/frax hh amend

gosuto-inzasheru:
just increasing readability to keep smoothening the bribe process every round :)


petrovska-petro (approved):
thanks for the brush up!

Issue #940, opened 2022-11-07
Brick upgradability in deprecated vaults


petrovska-petro:
In light of the decentralisation roadmap, we will start bricking the ability to upgrade deprecated vaults. Deprecated vaults can be found in the [registry_v2](https://etherscan.io/address/0xdc602965F3e5f1e7BAf2446d5564b407d5113A06#code), speficially those with **status 0**.

Procedure:

```
proxyAdmin.changeProxyAdmin(<proxy>, ZERO_ADDRESS) 
```

**Note:** the bricking of the upgradability of the vault includes its corresponding strategy also to be brick. i.e: the product as a whole

gosuto-inzasheru:
note: contracts do not accept `ZERO_ADDRESS` to be their admin

petrovska-petro:
**Workaround:** will be re the proxyadmin deploy an empty contract for bricking the upgradability and set it as the new admin

petrovska-petro:
empty contract to use as the new admin: https://etherscan.io/address/0x933FDBAc3773514384c6db30Eb196a0be543D617. cc: @jalbrekt85

petrovska-petro:
```
Found 25 deprecated vaults:
[
    'Badger Sett Curve.fi renBTC/wBTC/sBTC',
    'Badger Sett Curve.fi tBTC/sbtcCrv',
    'Badger SuperSett (Harvest) Curve.fi renBTC/wBTC',
    'Badger Sett Uniswap V2',
    'Badger Sett Uniswap V2',
    'Badger Sett Badger',
    'Badger Sett SushiSwap LP Token',
    'Badger Sett Curve.fi hBTC/wBTC',
    'Badger Sett Curve.fi pBTC/sbtcCRV',
    'Badger Sett Curve.fi oBTC/sbtcCRV',
    'Badger Sett Curve.fi bBTC/sbtcCRV',
    'Badger Sett Curve.fi USD-BTC-ETH',
    'Badger Sett Convex Token',
    'Badger Sett Interest bearing mStable BTC',
    'Badger Sett mBTC/HBTC Feeder Pool',
    'Badger Sett Curve.fi Factory USD Metapool: Magic Internet Money 3Pool',
    'Badger Sett Curve.fi Factory USD Metapool: Frax',
    'Badger Sett SushiSwap LP Token',
    'Badger Sett Digg',
    'Badger Sett SushiSwap LP Token',
    'Badger Sett SushiSwap LP Token',
[...]

PR #948, opened 2022-11-09
brick deprecated vaults script + sim

jalbrekt85:
handles #940 - `NEW_PROXY_ADMIN` will probably have to be changed


petrovska-petro (commented):
Bricking by updating the proxy admin looks good, there is 25 vaults in deprecated state, mix of v1's and v1.5's vaults

does the sim for `0x4b92d19c11435614CD49Af1b589001b7c08cD4D5` works for you properly in the minting?. getting `AttributeError: Contract 'ERC20' object has no attribute 'mint'`

will advise to remove the jsons from the pr, we will push them once we indeed post them in the msig. shortly, will share the addr of the empty contract to use as the "new admin"

scripts/issue/940/brick_deprecated.py
```diff
@@ -0,0 +1,79 @@
...
+                proxy_admin.changeProxyAdmin(vault, NEW_PROXY_ADMIN)
+
+                # arbitrary small amount to ensure funds can be minted
+                amount = 1e8
+                token = MintableForkToken(vault.token(), owner=safe.account)
+                token._mint_for_testing(safe, amount)
+                token.approve(vault, amount)
+
+                # edge case: https://etherscan.io/address/0x4b92d19c11435614CD49Af1b589001b7c08cD4D5#writeProxyContract
+                if vault.address == "0x4b92d19c11435614CD49Af1b589001b7c08cD4D5":
```
gosuto-inzasheru, line 71:
```suggestion
                if vault.address == r.yearn_vaults.byvWBTC:
```

gosuto-inzasheru:
@petrovska-petro just means you have that vault's token initiated locally as a simple `ERC20` interface.

should be solved by running the following in console before running:
```
>>> Contract.from_explorer(Contract('0x4b92d19c11435614CD49Af1b589001b7c08cD4D5').token())
```

gosuto-inzasheru (commented):
indeed these jsons are not to be included; they are not the actual transactions that got queued in the timelock onchain.

sim runs successful for me;
- proxyadmin is set to bricked contract with no content
- verifies that depositing and withdrawing remains possible afterwards

@dapp-whisperer there was a discussion around also disabling deposits for the 1.5 vaults, do we want to include that in these txs?

jalbrekt85:
> will advise to remove the jsons from the pr, we will push them once we indeed post them in the msig. shortly, will share the addr of the empty contract to use as the "new admin"
> 
> empty contract addr: https://etherscan.io/address/0x933FDBAc3773514384c6db30Eb196a0be543D617

addressed in ad0be0d

petrovska-petro (approved):

gosuto-inzasheru (approved):

Issue #967, opened 2022-11-16
Deprecate ibbtc and ren vaults and stop emissios for ibbtc crv


sajanrajdev:
- Deprecate vaults on Registry
- Modify ibBTC's emissions schedule to cancel emissions
- Sweep Q4 BADGER dripper for the Tree
- Transfer deficit + bcvxCRV BADGER deificits to the Tree
- Transfer the rest of BADGER to Trops

sajanrajdev:
Events traace:
```
Transaction was Mined
---------------------
Tx Hash: 0x7147a8e90279bf8adae03363cf31c2da126fe167964ce2fc6864a2a80bdba048
From: 0x0a9af7FAba0d5DF7A8C881e1B9cd679ee07Af8A2
To: 0x86cbD0ce0c087b482782c181dA8d191De18C8275
Value: 0
Function: Gnosis Safe.execTransaction
Block: 15990651
Gas Used: 419932 / 596344 (70.4%)

Events In This Transaction
--------------------------
├── IBadgerRegistryV2 (0xdc602965F3e5f1e7BAf2446d5564b407d5113A06)
│   ├── DemoteVault
│   │   ├── author: 0x86cbD0ce0c087b482782c181dA8d191De18C8275
│   │   ├── version: v1
│   │   ├── metadata: name=ibBTC/crvsBTC,protocol=Convex,behavior=None
│   │   ├── vault: 0xaE96fF08771a109dc6650a1BdCa62F2d558E40af
│   │   └── status: 0
│   └── DemoteVault
│       ├── author: 0x86cbD0ce0c087b482782c181dA8d191De18C8275
│       ├── version: v1
│       ├── metadata: name=renBTC/wBTC,protocol=Convex,behavior=None
│       ├── vault: 0x6dEf55d2e18486B9dDfaA075bc4e4EE0B28c1545
│       └── status: 0
│
├── AdminUpgradeabilityProxy (0x0A4F4e92C3334821EbB523324D09E321a6B0d8ec)
│   └── UnlockScheduleModified
│       ├── index: 39
│       ├── beneficiary: 0xaE96fF08771a109dc6650a1BdCa62F2d558E40af
│       ├── token: 0x3472A5A71965499acd81997a54BBA8D852C6E53d
│       ├── totalAmount: 0
│       ├── start: 1668620976
│       ├── end: 1668620976
│       ├── duration: 0
│       ├── timestamp: 15990651
│       └── blockNumber: 1668699106
[...]

PR #968, opened 2022-11-16
feat: deprecate and stop emissions script

sajanrajdev:
Tackles #967 

Run with:
```
brownie run scripts/issue/967/deprecate_and_stop_emissions.py
```

Run upkeep withdraw with:
```
brownie run scripts/issue/967/deprecate_and_stop_emissions.py withdraw_link
```

### DIFF
--- a/scripts/issue/967/deprecate_and_stop_emissions.py
+++ b/scripts/issue/967/deprecate_and_stop_emissions.py
@@ -31,8 +31,6 @@ ORIGINAL_DURATION_WEEKS = Decimal(
 )
 TOTAL_AMOUNT = 12637714285714284415007
 AMOUNT_PER_WEEK = Wei(f"{1552} ether")
-REMAINING_WEEKS = Decimal((ORIGINAL_SCHEDULE_END - int(time.time())) / 604800)
-BCVXCRV_BADGER_DEFICIT = int(Decimal(AMOUNT_PER_WEEK * REMAINING_WEEKS))
 
 
 def main():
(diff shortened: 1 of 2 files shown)


scripts/issue/967/deprecate_and_stop_emissions.py
```diff
@@ -0,0 +1,123 @@
...
+
+# Tree bcvxCRV BADGER deficit
+ORIGINAL_SCHEDULE_START = 1667606400
+ORIGINAL_SCHEDULE_END = 1672531200
+ORIGINAL_DURATION_WEEKS = Decimal(
+    (ORIGINAL_SCHEDULE_END - ORIGINAL_SCHEDULE_START) / 604800
+)
+TOTAL_AMOUNT = 12637714285714284415007
+AMOUNT_PER_WEEK = Wei(f"{1552} ether")
+REMAINING_WEEKS = Decimal((ORIGINAL_SCHEDULE_END - int(time.time())) / 604800)
```
petrovska-petro, line 34: i think here is more precise to use `lastTimestamp` from the dripper when the upkeep was called last time, otherwise will be a gap of emissions which will not get filled which is from the last upkeep ts and the current ts

sajanrajdev: Good call, fixing...

petrovska-petro: nvm, it is accounted with the tree deficit 👍

scripts/issue/967/deprecate_and_stop_emissions.py
```diff
@@ -0,0 +1,126 @@
...
+    safe.post_safe_tx()
+
+
+def withdraw_link(simulation="false"):
+    safe = GreatApeSafe(registry.eth.badger_wallets.techops_multisig)
+    safe.take_snapshot([LINK])
+    link = safe.contract(LINK, interface.ILinkToken)
+
+    if simulation == "true":
+        safe.init_chainlink()
```
petrovska-petro, line 117: cl class initiation needs to be moved to L112 area, otherwise when it is not sim will not find registry addr

sajanrajdev: Good catch, fixed.

petrovska-petro (approved):
lgtm, scripts sets correctly:

- deprecated status for ibbtc and renbtc vaults in registry to reflect in ui

- cancels upkeep job and retrieves link funds

- nullifies current schedule for ibbtc, stopping emissions completely for that vault its last schedule aka current

- sweps funds from dripper, covers deficit and emissions for bcvxcrv till end of q4

- rest of badger funds sent to trops

jalbrekt85 (approved):
lgtm

gosuto-inzasheru:
note that if the keeper performs an upkeep between posting and execing, the numbers will be wrong on the exec.

next upkeep will probaby happen tomo around noon utc: https://etherscan.io/address/0x9032618620fedac90990d4ba562ab0a86f5abae7#tokentxns

PR #977, opened 2022-11-22
chore: upload json files of queue tx for bricking upgradability

petrovska-petro:
related to #948 

files required to later run the execution of queued txs, also includes the delay change to 7 days as it was requested by @dapp-whisperer at posting time and naming changes in the folder created to signal better the purpose for the records


jalbrekt85 (approved):
lgtm

PR #980, opened 2022-11-23
emergency stable swap script

### LINKED ISSUE #978: create emergency script to quickly swap stables from one to the other
currently vault is holding usdc, usdt and dai in roughly equal parts. in an emergency case of one of these three currencies depegging or whatever, we would like to be able to act quickly and swap one stable into the other. 3pool is probably best, but a backup of using cowswap should be included as well.

i would suggest main taking following params: `symbol_in`, `symbol_out`, amount (optional), 3pool/cowswap (bool, 3pool being default)

prob best to create a new folder for this in `scripts/warroom` or something.

jalbrekt85:
closes #978

### DIFF
--- a/scripts/warroom/emergency_swap.py
+++ b/scripts/warroom/emergency_swap.py
@@ -19,7 +19,11 @@ def main(symbol_in, symbol_out, amount=None, use_3pool="true"):
 
     token_in = tokens[symbol_in]
     token_out = tokens[symbol_out]
-    amount = token_in.balanceOf(vault) if not amount else int(amount) * 10 ** token_in.decimals()
+    amount = (
+        token_in.balanceOf(vault)
+        if not amount
+        else int(amount) * 10 ** token_in.decimals()
+    )
 
     vault.take_snapshot(tokens=[token_in, token_out])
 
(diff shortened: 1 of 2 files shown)


scripts/warroom/emergency_swap.py
```diff
@@ -0,0 +1,34 @@
...
+
+    if use_3pool == "true":
+        vault.init_curve()
+        vault.curve.swap(token_in, token_out, amount)
+
+    else:
+        vault.init_cow(prod=prod)
+        vault.cow.market_sell(token_in, token_out, amount, 60 * 60 * 4)
+
+    vault.print_snapshot()
```
petrovska-petro, line 34: missing posting tx, rest looks good

petrovska-petro (approved):

scripts/warroom/emergency_swap.py
```diff
@@ -0,0 +1,35 @@
...
+
+    tokens = {
+        "usdc": vault.contract(r.treasury_tokens.USDC),
+        "usdt": vault.contract(r.treasury_tokens.USDT),
+        "dai": vault.contract(r.treasury_tokens.DAI),
+    }
+
+    token_in = tokens[symbol_in]
+    token_out = tokens[symbol_out]
+    amount = token_in.balanceOf(vault) if not amount else int(amount)
```
gosuto-inzasheru, line 22: i would suggest to add the decimals of the respective token here programmatically, so that the user can pass `100_000` as an amount an not have to worry about whether it is 6 or 18 decimals.

jalbrekt85: good idea

gosuto-inzasheru (approved):
nice one

petrovska-petro (approved):

Issue #979, opened 2022-11-23
Treasury Decision: Increase the amount of TCL in the BADGER/FRAXBP by $500k total notional value. 


jalbrekt85:
> Increase the amount of TCL in the BADGER/FRAXBP by $500k total notional value. 

Full Document:
https://docs.google.com/document/d/1QFQ5YsOAMmX_Uy4-rXpp1ixKYrSh7ZhhdsWBRPANIME/edit?usp=sharing

petrovska-petro:
instead of having two diff locking positions, we can leverage the following method: https://etherscan.io/address/0xa895b89d74a6bc23a284a0526e123ea776674cf5#code#L917

kek_id can be found in this events: https://etherscan.io/tx/0x74f4492c78385b633fdecb2200efdd99f7377dbf2e0b94468c06682fb78961c1#eventlog

jalbrekt85:
```py
Events In This Transaction
--------------------------
├── Badger (0x3472A5A71965499acd81997a54BBA8D852C6E53d)
│   └── Approval
│       ├── _owner: 0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e
│       ├── _spender: 0x5De4EF4879F4fe3bBADF2227D2aC5d0E2D76C895
│       └── _amount: 92250922509225100000000
│   
├── USD Coin (0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48)
│   └── Approval
│       ├── owner: 0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e
│       ├── spender: 0x5De4EF4879F4fe3bBADF2227D2aC5d0E2D76C895
│       └── value: 250000000000
│   
├── Badger (0x3472A5A71965499acd81997a54BBA8D852C6E53d)
│   └── Transfer
│       ├── _from: 0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e
│       ├── _to: 0x5De4EF4879F4fe3bBADF2227D2aC5d0E2D76C895
│       └── _amount: 92250922509225100000000
│   
├── USD Coin (0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48)
│   ├── Transfer
│   │   ├── from: 0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e
│   │   ├── to: 0x5De4EF4879F4fe3bBADF2227D2aC5d0E2D76C895
│   │   └── value: 250000000000
[...]

PR #985, opened 2022-11-28
increase tcl in badger/fraxbp position script

jalbrekt85:
handles #979


gosuto-inzasheru (approved):
looks good!

runs successfully, amounts leaving wallet make sense. `vault.convex.stake_lock` asserts increase in locked balance.

petrovska-petro (approved):
nice, topups in the private vault with 250k of stable and its equivalent of badger increasing tcl ✅ 

v clean way of importing seeding 🧠

PR #988, opened 2022-11-29
refactor: claimable token list votium can be compiled programmatically

gosuto-inzasheru:
note that in order for this to work you need to have to [`Votium` repo](https://github.com/oo-00/Votium) (cloned) in `data/Votium`

this means there is no more need to maintain the `bribe_tokens_claimable` entry in the address book. on a new round, just run:
```
$ cd data/Votium;git pull;cd ../..;brownie run badger/process_bribes_bvecvx step0_1
```


jalbrekt85 (approved):
clever method to get tokens programmatically, lgtm

Issue #995, opened 2022-11-30
Treasury decision: harvesting strategy for convex-frax tcl


petrovska-petro:
> Swap all yield harvested from Frax gauges to DAI 

Details: https://docs.google.com/document/d/1VHwaeDi6LeyNGzio1OTOtvGrVpVV7pxMRhBgv-M0UWE/edit

petrovska-petro:
```
snapshot result for 0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e:
                  balance_before              balance_after             balance_delta
symbol                                                                               
FXS         0.000000000000000000   4,985.896017614070455272  4,985.896017614070455272
CRV     6,631.324873915028450394  10,823.822250955366227149  4,192.497377040337776755
CVX       285.146969578346223366     454.417434125994569302    169.270464547648345936 
```

```
{
│   'ethereum_client': <gnosis.eth.ethereum_client.EthereumClient object at 0x7fc4dbe5cfd0>,
│   'safe_address': '0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e',
│   'to': '0x40A2aCCbd92BCA938b02010E17A5b8929b49130D',
│   'value': 0,
[...]

PR #996, opened 2022-11-30
Process convex-frax strategy rewards to dai

petrovska-petro:
tackles #995 

```
brownie run badger/convex_frax_tcl_strategy
```

### DIFF
--- a/scripts/badger/convex_frax_tcl_strategy.py
+++ b/scripts/badger/convex_frax_tcl_strategy.py
@@ -1,0 +1,55 @@
+from great_ape_safe import GreatApeSafe
+from helpers.addresses import r
+
+# flag
+prod = False
+
+
+# slippage
+COEF = 0.98
+
+
+def main():
+    vault = GreatApeSafe(r.badger_wallets.treasury_vault_multisig)
+    vault.init_cow(prod)
+
+    # tokens involved
+    dai = vault.contract(r.treasury_tokens.DAI)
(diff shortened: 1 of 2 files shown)


scripts/badger/convex_frax_tcl_strategy.py
```diff
@@ -0,0 +1,55 @@
...
+    # contracts
+    private_vault = vault.contract(r.convex.frax.private_vaults.badger_fraxbp)
+
+    # snap
+    tokens = [fxs, crv, cvx]
+    vault.take_snapshot(tokens)
+
+    # NOTE: vault contains crv & cvx > 0, to make sure it is only the strat rewards check balances before
+    crv_balance_before = crv.balanceOf(vault)
+    cvx_balance_before = cvx.balanceOf(vault)
```
gosuto-inzasheru, line 31: while technically correct, i would not exclude these tokens and just sell all of it.

looking at the origin, it actually originates from the badgerfraxbp being staked on the convex gauge, from before there was a frax gauge available. so fits well in the purpose of this script imo.

it is not considerable, it cleans up the vault and i would not know what to do with it otherwise.

petrovska-petro: selling whole balance change included in 67307fa33055e1e8249c124ba8645de85061148c

jalbrekt85 (approved):
claims rewards and sells for dai, lgtm

PR #997, opened 2022-11-30
add lockAdditional to convex class and use in script

jalbrekt85:
handles https://github.com/Badger-Finance/badger-multisig/issues/979#issuecomment-1330945310


gosuto-inzasheru (approved):

petrovska-petro (approved):
kek_id looks good and script topups lock correctly ✅

PR #1001, opened 2022-12-02
refactor great ape safe inits

jalbrekt85:
have been wanting to do this for a long time, but wasn't able to materialize it until now

- dynamically create `GreatApeSafe.init_{api_name}` methods from an api dict (`{init name: Module}`)
- dynamically build api dict from modules in `great_ape_safe/ape_api/`

consequentially, adding an api to GreatApeSafe now just requires to add a module to `great_ape_safe/ape_api/`.
naming patterns are the same with the previous manual interface naming, thus being backwards compatible with all existing scripts.

for naming it is assumed you are following [pep-8](https://peps.python.org/pep-0008/#class-names) guidelines
`init name`: `init_{file name}`
`Module`: class name

e.g:
`great_ape_safe/ape_api/uni_v3.py `
```py
class UniV3:
```
can be accessed through:
```py
safe.init_uni_v3(); safe.uni_v3
```

### DIFF
--- a/great_ape_safe/ape_api/__init__.py
+++ b/great_ape_safe/ape_api/__init__.py
@@ -1,0 +1,18 @@
+import glob
+import os
+
+
+files = glob.glob("great_ape_safe/ape_api/[!_]*.py")
+module_names = [os.path.splitext(file)[0].replace("/", ".") for file in files]
+
+ape_apis = {}
+for module_path in module_names:
+    module = __import__(module_path, fromlist=[module_path.capitalize()])
+    class_name = module_path.split(".")[-1].capitalize()
+    init_name = module_path.split(".")[-1]
+    if "_" in class_name:
+        # e.g: Uni_v3 should be UniV3
+        i = class_name.index("_") + 1
+        class_name = class_name[:i] + class_name[i:].capitalize()
+        class_name = class_name.replace("_", "")
+    ape_apis[init_name] = getattr(module, class_name)
--- a/great_ape_safe/ape_api/__init__.py
+++ b/great_ape_safe/ape_api/__init__.py
@@ -2,20 +2,10 @@ import glob
 import os
 
 
-def module_path_to_class_name(module_path):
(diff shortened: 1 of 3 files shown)


gosuto-inzasheru (commented):
how do we handle params, like in the case of cow of snapshot?

gosuto-inzasheru:
one option is to not accepts params; for cow prod and staging have been merged into the same env anyway and for snapshot i think there is a way around providing the prop id on initialisation.

jalbrekt85:
> how do we handle params, like in the case of cow of snapshot?

good point, updated to accept be able to accept both pos & kw args: e1119ad

gosuto-inzasheru (approved):
tested locally:
```
>>> from great_ape_safe import GreatApeSafe
>>> safe=GreatApeSafe('techops.badgerdao.eth')
>>> safe.init_snapshot('0xc950294fd8c00b2901026b62b568b01923e62166fca1e96d738f4a27bdf26faf')
>>> safe.snapshot.proposal_id
'0xc950294fd8c00b2901026b62b568b01923e62166fca1e96d738f4a27bdf26faf'
```

nice space and time safer :)

petrovska-petro (approved):
very neat cleanup and reorg 🔥 

tested smoothly:

```
>>> from great_ape_safe import GreatApeSafe
>>> safe=GreatApeSafe('techops.badgerdao.eth')
>>> safe.init_uni_v3()
>>> safe.uni_v3.Q128
340282366920938463463374607431768211456
>>> safe.uni_v3.factory
<IUniswapV3Factory Contract '0x1F98431c8aD98523631AE4a59f267346ea31F984'>
>>> safe.uni_v3.factory.address
'0x1F98431c8aD98523631AE4a59f267346ea31F984'
>>> safe.init_snapshot('0xc950294fd8c00b2901026b62b568b01923e62166fca1e96d738f4a27bdf26faf')
>>> safe.snapshot.format_choice('80/20 BADGER/WBTC')
20
```

Issue #986, opened 2022-11-28
Update snapshot class


petrovska-petro:
Breaking changes: https://snapshot.mirror.xyz/vuManI14DW8u2zhrlskndNgQcXOTbKIelQvkgmxOG2k

- [] modified class to EIP-712 signing  
- [] run-test on any of `test_msigs`

PR #1006, opened 2022-12-07
Update snapshot class for matching eip7712 specs and data structured

petrovska-petro:
Tackles #986

### DIFF
--- a/great_ape_safe/ape_api/snapshot.py
+++ b/great_ape_safe/ape_api/snapshot.py
@@ -3,8 +3,9 @@ import json
 import time
 from rich.console import Console
 
-from eth_account import messages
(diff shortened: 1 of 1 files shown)


gosuto-inzasheru:
discussed with @petrovska-petro; will be easier to help debug and review with a test case in `tests/` which now fails but should pass.

petrovska-petro:
> discussed with @petrovska-petro; will be easier to help debug and review with a test case in `tests/` which now fails but should pass.

push test in 6e80c8dc773d8b780d38a2b7cfc79798088f8a4d

run:

```
brownie test tests/snapshot/test_eip712_hash.py -s 
```

PR #1025, opened 2022-12-14
fix: hotfixes for new structure of `ACTIVE_TOKENS`

gosuto-inzasheru:
following changes in #988 , should have been included there


petrovska-petro (approved):
Remove the addrs of cvx and weth to not be processed in `step1`. Reads tokens addrs from [json file](https://github.com/oo-00/Votium/blob/main/merkle/activeTokens.json) ✅

jalbrekt85 (approved):

PR #1027, opened 2022-12-15
feat: also consider strategist when whitelisting

gosuto-inzasheru:
func only looked at governance and not strategist, while both are considered by the contract when approving contract access:
Screenshot 2022-12-15 at 12 45 29

### DIFF
--- a/great_ape_safe/ape_api/badger.py
+++ b/great_ape_safe/ape_api/badger.py
@@ -302,6 +302,12 @@ class Badger:
                 sett.approveContractAccess(candidate_addr)
                 assert sett.approved(candidate_addr)
                 return True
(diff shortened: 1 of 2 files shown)


petrovska-petro:
will consider also updating this [script](https://github.com/Badger-Finance/badger-multisig/blob/main/scripts/badger/whitelist.py#L10) in this pr to allow being called by `techops` if possible

sajanrajdev (dismissed):
Simple enough, allows for function to be called by strategist.

LGTM

petrovska-petro (approved):
Allows whitelisting via strategist in the class and updates general script to either trigger the WL action via gov or strategist ✅

jalbrekt85 (approved):

Issue #1014, opened 2022-12-09
conclude team vesting


gosuto-inzasheru:
of the 2.1m badger set aside for the team [here](https://etherscan.io/address/0xe4aa1d8aaf8a50422bc5c7310deb1262d1f6f657#tokentxns), still [`728820123211567732115678`](https://etherscan.io/tx/0xe02cfa561e73d54163c383ab13b9ed7d970d12b2c1cb9014730b6f7862ac82ce) needs to be distributed before the eoy:
- 60% of that amount to `0x4300dB1B4D8F417Ccf6672CEa75174b39d9A5e4e`
- 40% to `0xECE2C248630C7eE5A1A2C2Ea780233b615E6862D`

to make sure the addresses can properly receive $badger, the first tx should be a test tx of 1 $badger to each, followed by a second tx with the remainders.

jalbrekt85:
`test_transfer`
```py
{
│   'ethereum_client': <gnosis.eth.ethereum_client.EthereumClient object at 0x10b6c1100>,
│   'safe_address': '0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e',
│   'to': '0x40A2aCCbd92BCA938b02010E17A5b8929b49130D',
│   'value': 0,
[...]

PR #1030, opened 2022-12-15
conclude team vesting script

jalbrekt85:
handles #1014


dapp-whisperer:
this script should be handled in two phases:
- a test transaction to both addresses with a much smaller amount
- the remainder

petrovska-petro (approved):
scripts carries correctly the test transfer and reminder amounts with right split, and verify the balances after the transfer onchain ✅

gosuto-inzasheru (approved):

PR #1033, opened 2022-12-16
feat: divest from cvx through ladder orders

gosuto-inzasheru:


petrovska-petro (approved):
runs smoothly rebalancing the pool at 50:50 ratio of *bvecvx:cvx*, currently around 14k cvx required to add as liquidity, also open orders with 6 days deadline with 5%, 7.5% and 10% price target above spot price for usdc with vault as destination ✅

jalbrekt85 (approved):
lgtm

Issue #1031, opened 2022-12-15
wbtc big dao migration


gosuto-inzasheru:
> joe koh, [15 Dec 2022 at 23:23:47]:
> Hi everybody!
> 
> BitGo has kicked off the claimOwnership function for the New Big DAO multisig here: https://etherscan.io/tx/0xe1a0069a37390e287ca399a4ee79036660ac235da08cb86753045fe5d9e97acc (txId: `0`)
> 
> From your NEW Big DAO address, please broadcast the following HEX DATA:
`0xc01a8c840000000000000000000000000000000000000000000000000000000000000000` 
> 
> to the Big DAO Contract Address: `0xB33f8879d4608711cEBb623F293F8Da13B8A37c5`
>
> You can verify this HEX yourselves by going to the Big DAO Contract page: https://etherscan.io/address/0xB33f8879d4608711cEBb623F293F8Da13B8A37c5#writeContract and using function “5. confirmTransaction” and input the txid `0`. See following screenshot:
>
> ![photo_2022-12-15 23 42 51](https://user-images.githubusercontent.com/2835259/207936321-16431ff5-7fc9-4be8-9c6a-91dafabaaab1.jpeg)
> 
> Please confirm once completed. We need 6 more votes (BitGo + Kyber have are in)

gosuto-inzasheru:
queued transaction `0` on the new wbtc dao msig calls `0x4e71e0c8` (`claimOwnership`) on `0xCA06411bd7a7296d7dbdd0050DFc846E95fEBEB7`, the wbtc token controller. on that controller, the new wbtc dao msig is indeed the `pendingOwner`: 

<img width="404" alt="Screenshot 2022-12-15 at 23 51 28" src="https://user-images.githubusercontent.com/2835259/207937773-6d227ab0-afe1-43ab-a8f5-9815bbd43b55.png">

the proposed calldata signs this queued tx:
```
>>> wbtcdao = Contract('0xB33f8879d4608711cEBb623F293F8Da13B8A37c5')
>>> wbtcdao.decode_input('0xc01a8c840000000000000000000000000000000000000000000000000000000000000000')
("confirmTransaction(uint256)", [0])
```

jalbrekt85:
```py
Events In This Transaction
--------------------------
├── MultiSigWalletWithDailyLimit (0xB33f8879d4608711cEBb623F293F8Da13B8A37c5)
│   └── Confirmation
[...]

PR #1039, opened 2022-12-18
wbtc dao migration script

jalbrekt85:
handles #1031 
```
brownie run issue/1031/wbtc_dao_migration
```

### DIFF
--- a/helpers/addresses.py
+++ b/helpers/addresses.py
@@ -593,9 +593,7 @@ ADDRESSES_ETH = {
     "frax": {
         "BADGER_FRAXBP_GAUGE": "0x5a92EF27f4baA7C766aee6d751f754EBdEBd9fae",
     },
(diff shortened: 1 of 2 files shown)


gosuto-inzasheru (approved):
lgtm. next iteration could be to add a class or something to the great ape safe module that can interact with these [legacy multisigs](https://github.com/gnosis/MultiSigWallet/blob/master/contracts/MultiSigWalletWithDailyLimit.sol). maybe something like `safe.gnosis.confirm(id=0)`?

idk lets see how often these interactions occur :)

petrovska-petro (approved):
smooth, techops will be confirming the new owner of the wbtc controller contract, which is this [msig](https://etherscan.io/address/0xB33f8879d4608711cEBb623F293F8Da13B8A37c5#readContract) ✅

Issue #1045, opened 2022-12-19
withdraw from all lp positions that hold renbtc and swap to wbtc


gosuto-inzasheru:
i count almost $15k worth of renbtc positions in treasury ops worth withdrawing from:
- naked renbtc
- ibbtc/sbtcCRV-f
- crvRenWSBTC
- crvRenWBTC
- mstable lps (2x)

negligible:
- pbtc
- tbtc
- pbtc2
- bbtc

gosuto-inzasheru:
probably worth combining with #884

petrovska-petro:
```
snapshot result for 0x042B32Ac6b453485e357938bdC38e0340d4b9276:
                        balance_before          balance_after           balance_delta
symbol                                                                               
ETH              28.619595083099736545  15.019595083099736545  -13.600000000000000000
WBTC              6.276852910000000000   7.157324670000000000    0.880471760000000000
crvRenWSBTC       0.103865435494786303   0.000000000000000000   -0.103865435494786303
crvRenWBTC        0.081714735802236960   0.000000000000000000   -0.081714735802236960
ibbtc/sbtcCRV-f   0.465596920478966793   0.000000000000000000   -0.465596920478966793
fPmBTC/HBTC       0.134068808178073103   0.000000000000000000   -0.134068808178073103
imBTC             0.830583609141352465   0.000000000000000000   -0.830583609141352465 
```

petrovska-petro:
https://etherscan.io/tx/0xb5d59152d248705013e79953e23d49e7ed077c407a0574dc8c479bb4e6a10eda

Issue #884, opened 2022-10-18
quarter 1: topup gas station


gosuto-inzasheru:
prev #880

gosuto-inzasheru:
this might also be a good oppo to unwrap the little bit of $weth there

petrovska-petro:
tx contains the top-up of gas station with 13.6 eth: https://etherscan.io/tx/0xb5d59152d248705013e79953e23d49e7ed077c407a0574dc8c479bb4e6a10eda

PR #1047, opened 2022-12-19
feat: clean btc lps into wbtc and topup gas station

petrovska-petro:
Tackles #1045 & #884 conditionally

```
brownie run treasury/wbtc_fee_collector main 5000000000000000000
```


scripts/treasury/wbtc_fee_collector.py
```diff
@@ -0,0 +1,87 @@
...
+        min_out_wbtc_from_im_btc,
+        wbtc,
+        trops,
+        r.treasury_tokens.mBTC,
+        True,
+    )
+
+    # 3. Top-up gas station
+    # NOTE: not added the withdraw from `weth` since it breaks the flow of the scripts by reverting
+    if top_eth_mantissa > 0:
```
gosuto-inzasheru, line 84: think this needs a cast? `Decimal` works well on the cli arg strings

scripts/treasury/wbtc_fee_collector.py
```diff
@@ -0,0 +1,87 @@
...
+        True,
+        min_out_wbtc_from_im_btc,
+        wbtc,
+        trops,
+        r.treasury_tokens.mBTC,
+        True,
+    )
+
+    # 3. Top-up gas station
+    # NOTE: not added the withdraw from `weth` since it breaks the flow of the scripts by reverting
```
gosuto-inzasheru, line 83: why doest weth.withdraw revert..?

gosuto-inzasheru: ah it is that thing due to some weird eth fallback behaviour in the gnosis safe contract.

sure i guess we can skip for now without having to build calldata manually...

gosuto-inzasheru (changes requested):

scripts/treasury/wbtc_fee_collector.py
petrovska-petro: good catch, pushed in 5c3a8483fa67b7a8cc0fc7246d30f445fe5ca19a

scripts/treasury/wbtc_fee_collector.py
petrovska-petro: exactly that issue with the gnosis contract 

```
receive() external payable {
        emit SafeReceived(msg.sender, msg.value);
    }
```

gosuto-inzasheru (approved):

jalbrekt85 (approved):
nice, lgtm

Issue #1048, opened 2022-12-20
execute on tcd-21


gosuto-inzasheru:
decision asks for a deposit of all our ether in tranche c (1022.96 - 90 = 932.96) plus matching badger (~491k) into the badger/reth pool.

doing this before redepositing current tcl positions into the new aura gauges would enable us to combine those deposits

ref: https://discord.com/channels/1011322803414388897/1011333638044008548/1050566433299505262

gosuto-inzasheru:
https://etherscan.io/tx/0x732506a05467289bdf19a1233ede55b46f9af334c47f9e0b30832699c7ad7770

PR #1049, opened 2022-12-20
feat: swap weth to reth and deposit into badgerreth

gosuto-inzasheru:
tbh i put the native reth deposit in there because i took a deep dive into the code and didnt want to loose that knowledge

but the room to deposit there atm is <.1eth, so i dont see it as being realistic that we will take that path

maxing out whatever room there is is difficult due to time between posting and execution; too much chance to get frontrun and the tx reverting

see #1048 for details and amounts


petrovska-petro (approved):
Script exec the following:

- considers current reth bal in msig already leveraging exchange rate reth:eth and deducts for the target eth amount to be deposited ✅ 

- decided to swap weth to reth or directly depositing eth into rocket eco depending of network condition, the later very unlikely scenario as its current state ✅ 

- calculates matching badger amount leveraging cg api pricing ✅ 

- deposits into balancer pool, but it does not stake in balancer gauge directly, waiting to depositing back treasury funds into aura ✅

jalbrekt85 (approved):
amounts lgtm
```py
                               balance_before                 balance_after                balance_delta
symbol                                                                                                  
rETH                    12.553624153997098192         17.251866796933356026         4.698242642936257834
WETH                   941.699344603880876279          0.000000000000000000      -941.699344603880876279
BADGER           2,834,141.398451194476730152  2,341,386.570426297038149416  -492,754.828024897438580736
50BADGER-50rETH          0.000000000000000000     40,878.956712821551429584    40,878.956712821551429584
```

Issue #1043, opened 2022-12-19
unstake badger/wbtc lp from convex and remove liquidity from curve


gosuto-inzasheru:
part of #1042

gosuto-inzasheru:
also dogfooded i believe

jalbrekt85:
```py
{
│   'ethereum_client': <gnosis.eth.ethereum_client.EthereumClient object at 0x13e855730>,
│   'safe_address': '0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e',
│   'to': '0x40A2aCCbd92BCA938b02010E17A5b8929b49130D',
│   'value': 0,
│   'data': HexBytes('0x8d80ff0a000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000002c400c36442b4a4522e871399cd717abdd847ab11fe8800000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000084fc6f78650000000000000000000000000000000000000000000000000000000000028c86000000000000000000000000d0a7a8b98957b9cd3cfb9c0425abe44551158e9e00000000000000000000000000000000ffffffffffffffffffffffffffffffff00000000000000000000000000000000ffffffffffffffffffffffffffffffff00c36442b4a4522e871399cd717abdd847ab11fe8800000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000084fc6f786500000000000000000000000000000000000000000000000000000000000306ce000000000000000000000000d0a7a8b98957b9cd3cfb9c0425abe44551158e9e00000000000000000000000000000000ffffffffffffffffffffffffffffffff00000000000000000000000000000000ffffffffffffffffffffffffffffffff00ec1c717a3b02582a4aa2275260c583095536b61300000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000004853828b60050f3752289e1456bfa505afd37b241bca23e685d000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000645b36389c0000000000000000000000000000000000000000000000319ce7165c9f20f76a0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000'),
│   'operation': 1,
│   'safe_tx_gas': 0,
│   'base_gas': 0,
│   'gas_price': 0,
[...]

PR #1051, opened 2022-12-20
badger/wbtc withdraw script

jalbrekt85:
handles #1043

### DIFF
--- a/scripts/issue/1043/unwind_badger_wbtc.py
+++ b/scripts/issue/1043/unwind_badger_wbtc.py
@@ -1,0 +1,23 @@
+from brownie import interface
+
+from great_ape_safe import GreatApeSafe
+from helpers.addresses import r
+
+
+def main():
+    vault = GreatApeSafe(r.badger_wallets.treasury_vault_multisig)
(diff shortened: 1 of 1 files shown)


petrovska-petro (approved):
script claims univ3 fees and un-dogfood the whole badger-wbtc curve lp from the sett and withdraws the lp out into the underlyings ✅

gosuto-inzasheru (approved):
runs successfully onfork

not sure about these though, but not a showstopper:
```
├── IUniswapV3Pool (0xe15e6583425700993bd08F51bF6e7B73cd5da91B)
│   └── Burn
│       ├── owner: 0xC36442b4a4522E871399CD717aBDD847Ab11FE88
│       ├── tickLower: 320100
│       ├── tickUpper: 326940
│       ├── amount: 0
│       ├── amount0: 0
│       └── amount1: 0
```
maybe those events occur even when just collecting for a single id?

jalbrekt85:
> maybe those events occur even when just collecting for a single id?

yup, seems like it does, calling: `vault.uni_v3.collect_fee(167046)` still emits the same events

Issue #1053, opened 2022-12-21
voter msig needs a minimum balance of bvecvx


gosuto-inzasheru:
the pending dfd msig tx performs an onchain balance verification: #221 

voter msig currently does not meet that threshold

- final balance needed: `69884615940476827071326`
- pending bvecvx: `19234917575004743000000`
- current balance: `41175460547350534228174`
- gap: final - pending - current = `9474237818121549843152`

with 17k of $cvx in that contract (waiting for cowswap order to be filled), my suggestion is to temporary deposit 10k cvx into bvecvx, exec the dfd tx and withdraw directly after.

script will need to post two txs: deposit 10k cvx and withdraw 10k bvecvx (one for before the dfd exec and one for after)

gosuto-inzasheru:
https://etherscan.io/tx/0x16f2973766efcab629a12316b13b267327a67d73de88af05240d4f1728c1731d

PR #1057, opened 2022-12-22
feat: temp increase bvecvx balance on voter msig

gosuto-inzasheru:
solution for #1053


petrovska-petro (approved):
deposits 20k cvx into bvecvx to allow dfd msig tx to be exec successfully and post also wd of the same amount deposited, lgtm ✅

jalbrekt85 (approved):
deposits 20k cvx to cover deficit and withdraws after dfd exec, lgtm

dapp-whisperer (approved):

Issue #1061, opened 2022-12-23
Withdraw from Ribbon vault


Hal69k:
Based on the parameters outlined in the original purposely the 100k BADGER should be withdrawn from the Ribbon vault as soon as possible.

![image](https://user-images.githubusercontent.com/90224840/209332613-b60cee85-76ff-4db2-a33f-40a8d967f52c.png)

gosuto-inzasheru:
```
Events In This Transaction
--------------------------
├── Ribbon BADGER Treasury Vault (0x270F4a26a3fE5766CcEF9608718491bb057Be238)
│   ├── Redeem
│   │   ├── account: 0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e
│   │   ├── share: 100000000000000000000000
│   │   └── round: 1
│   ├── Transfer
│   │   ├── from: 0x270F4a26a3fE5766CcEF9608718491bb057Be238
│   │   ├── to: 0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e
│   │   └── value: 100000000000000000000000
│   ├── InitiateWithdraw
│   │   ├── account: 0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e
│   │   ├── shares: 100000000000000000000000
│   │   └── round: 5
│   └── Transfer
│       ├── from: 0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e
│       ├── to: 0x270F4a26a3fE5766CcEF9608718491bb057Be238
│       └── value: 100000000000000000000000
│   
└── Gnosis Safe (0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e)
    └── ExecutionSuccess
        ├── txHash: 0xb7c2836dc5c34011823393e26410ceb5932cb8bc31921342564b2ab074e29de5
        └── payment: 0
```

gosuto-inzasheru:
withdrawal initiated: https://etherscan.io/tx/0x650bb0d5ca81d66677fc63acacee20fce417519637ec983d6f9ae8938008f4df

gosuto-inzasheru:
`brownie run issue/1061/withdraw_ribbon exec_wd`

gosuto-inzasheru:
```
Events In This Transaction
--------------------------
├── Ribbon BADGER Treasury Vault (0x270F4a26a3fE5766CcEF9608718491bb057Be238)
│   ├── Withdraw
[...]

PR #1062, opened 2022-12-23
feat: initiate withdraw from ribbon

gosuto-inzasheru:
solution to #1061


petrovska-petro (approved):
scripts kicks on the process of wd funds from ribbon ✅ 

will req another call to [completeWithdraw](https://etherscan.io/address/0xc92e6b70eb6456171d32c3b5904386c05ec983ff#code#F1#L748), currently in round 5

jalbrekt85 (approved):
Initiates a withdrawal of 100k that can be processed once the round completes, lgtm

Issue #926, opened 2022-11-03
quarter 1: maintain badger dripper for 2023


gosuto-inzasheru:
prev #829 

once drippers are deployed:
- [x] cancel q4 dripper (https://automation.chain.link/mainnet/142)
- [x] register rem_dripper and tree_dripper 2023 upkeep on new chainlink registry
- [x] fund rem_dripper with 300k badger and tree_dripper with 400k badger

note: make sure there is no serious deficit/surplus of badger in the tree! if there is, correct by adjusting tree dripper amount.

gosuto-inzasheru:
nonce 446:
```
Events In This Transaction
--------------------------
├── IKeeperRegistry (0x7b3EC232b08BD7b4b3305BE0C044D907B2DF960B)
│   └── UpkeepCanceled
│       ├── id: 142
│       └── atBlockHeight: 16382625
│   
├── IKeeperRegistry (0x02777053d6764996e594c3E88AF1D58D5363a2e6)
│   └── UpkeepCanceled
│       ├── id: 14270489937535892760600590423564455802049459622872924518248134412357549386527
│       └── atBlockHeight: 16382625
│   
└── Gnosis Safe (0x86cbD0ce0c087b482782c181dA8d191De18C8275)
    └── ExecutionSuccess
        ├── txHash: 0x03fd4690e5e149a574d66841524950d921f997ab820d1949cd430c3775f9c206
        └── payment: 0
```

gosuto-inzasheru:
cancellations: https://etherscan.io/tx/0x87ea6497fbe57c027aa57522ac59744080b438b9dd53a55d1fbaa7c91dc2aa95#eventlog

gosuto-inzasheru:
register new upkeeps and adjust gas watchlist:
```
                                                            addrs      min_bals   min_top_ups
ops_deployer           0xDA25ee226E534d868f0Dd8a459536b03fEE9079b  5.000000e+18  5.000000e+17
ops_deployer2          0xeE8b29AA52dD5fF2559da2C50b1887ADee257556  2.000000e+18  5.000000e+17
ops_deployer3          0x283C857BA940A61828d9F4c09e3fceE2e7aEF3f7  2.000000e+18  5.000000e+17
ops_deployer4          0xef42D748e09A2d9eF89238c053CE0B6f00236210  2.000000e+18  5.000000e+17
[...]

PR #1078, opened 2023-01-06
chore: maintain old registry in codebase

gosuto-inzasheru:
needed for cancelling jobs that are still registered on the old registry (ie for #926)


petrovska-petro (approved):
confirmed that [dripperq4](https://automation.chain.link/mainnet/142) was register with old registry and have funds to be wd after cancelling ✅

jalbrekt85 (approved):

PR #1081, opened 2023-01-09
fix: dont pass fee in case of limit order

gosuto-inzasheru:


petrovska-petro (approved):
confirmed for LMTs that fee_amount should be set to zero

docs: https://docs.cow.fi/tutorials/submit-limit-orders-via-api/limit-order-structure

jalbrekt85 (approved):
simple fix, lgtm

Issue #1122, opened 2023-02-03
tcd#23: gtc univ3 position


gosuto-inzasheru:
- sell 14.5% of current gtc holdings for ether
- deposit all gtc an needed eth into a 30bps univ3 pool with range [-10%,+100%]

ref: https://discord.com/channels/1011322803414388897/1011333638044008548/1070407564048736266

jalbrekt85:
```py
{
│   'ethereum_client': <gnosis.eth.ethereum_client.EthereumClient object at 0x13a918ca0>,
│   'safe_address': '0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e',
│   'to': '0x40A2aCCbd92BCA938b02010E17A5b8929b49130D',
│   'value': 0,
[...]

PR #1127, opened 2023-02-07
create uni-v3 gtc position

jalbrekt85:
handles #1122, includes a `get_amount_out` method in the uni-v3 class to help dynamically set ranges

### DIFF
--- a/scripts/issue/1122/create_gtc_position.py
+++ b/scripts/issue/1122/create_gtc_position.py
@@ -1,0 +1,41 @@
+from great_ape_safe import GreatApeSafe
+from helpers.addresses import r
+
+LOWER_RANGE_MULTIPLIER = 0.9
+UPPER_RANGE_MULTIPLIER = 2
+GTC_SELL_PCT = 0.16
+
+
+def main():
+    vault = GreatApeSafe(r.badger_wallets.treasury_vault_multisig)
+    vault.init_uni_v3()
+
+    gtc = vault.contract(r.treasury_tokens.GTC)
+    weth = vault.contract(r.treasury_tokens.WETH)
+
+    vault.take_snapshot(tokens=[gtc, weth])
+
+    sell_amount = int(gtc.balanceOf(vault) * GTC_SELL_PCT)
+    vault.uni_v3.swap([gtc, weth], sell_amount)
+    vault.print_snapshot()
+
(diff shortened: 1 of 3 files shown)


scripts/issue/1122/create_gtc_position.py
```diff
@@ -0,0 +1,41 @@
...
+
+    range_0 = gtc_rate * LOWER_RANGE_MULTIPLIER / 1e18
+    range_1 = gtc_rate * UPPER_RANGE_MULTIPLIER / 1e18
+
+    vault.uni_v3.mint_position(
+        r.uniswap.v3pool_eth_gtc,
+        range_0,
+        range_1,
+        weth.balanceOf(vault),
+        gtc.balanceOf(vault),
```
gosuto-inzasheru, line 37: this will need to be done dusty, since outcome of swap is not 100% deterministic in our case

gosuto-inzasheru (changes requested):
code looks clean! just be wary of diff between posting and execing :)

scripts/issue/1122/create_gtc_position.py
```diff
@@ -0,0 +1,41 @@
+from great_ape_safe import GreatApeSafe
+from helpers.addresses import r
+
+LOWER_RANGE_MULTIPLIER = 0.9
+UPPER_RANGE_MULTIPLIER = 2
+GTC_SELL_PCT = 0.16
```
petrovska-petro, line 6: this i think should be 14.5% instead
```suggestion
GTC_SELL_PCT = 0.145
```

jalbrekt85: updated in fdcaa9d

scripts/issue/1122/create_gtc_position.py
jalbrekt85: good catch, using `0.97`: 517e2e9

petrovska-petro: should not be only dusty for weth?

jalbrekt85: yup, dont see why its needed for gtc

jalbrekt85: > should not be only dusty for weth?

e5833d1

petrovska-petro (approved):

PR #1131, opened 2023-02-08
Hotfix/claim hh api

gosuto-inzasheru:


petrovska-petro (approved):
updates endpoint, since prev got deprecated ✅ 

adds right addr of stg in the book ✅ 

increase slippage in both incentive processing scripts ✅ 

reads balance of claimed token directly in the processor, which can help to process dust which prev rounds were not pick by cow ✅

jalbrekt85 (approved):
lgtm

Issue #1134, opened 2023-02-09
adjust swap/lock ratio of aura in aura strategy


gosuto-inzasheru:
should be set to swap100%/lock0% as per consensus reached here: https://discord.com/channels/1011322803414388897/1011333638044008548/1072761788761911316

PR #1136, opened 2023-02-09
chore: not lock any aura

gosuto-inzasheru:
script for #1134


petrovska-petro (approved):

jalbrekt85 (approved):
lock 0 by default, lgtm

Issue #1022, opened 2022-12-13
fix sigs sim for tenderly url


gosuto-inzasheru:
generated by us: https://dashboard.tenderly.co/gosuto/msig-sim/simulator/4c81b279-3477-48ff-be81-062d701d390d
generated by safe: https://dashboard.tenderly.co/public/safe/safe-apps/simulator/8394dea4-132e-4a95-a667-40398b85dca0

gosuto-inzasheru:
also getting the following error sometimes: `HTTPError: 400 Client Error: Bad Request for url: https://api.tenderly.co/api/v1/account/gosuto/project/msig-sim/simulate`

gosuto-inzasheru:
lastly i think we should consider (also) calling this debank api endpoint: https://docs.open.debank.com/en/reference/api-pro-reference/wallet#enhanced-transaction-pre-execution

it gives erc20 and $ value deltas from before and after tx

PR #1137, opened 2023-02-10
Fix tenderly sim support

petrovska-petro:
Tackles #1022 

## What was the issue?

Regarding the `400 Client Error` my findings was that sometimes despite of the payload being correct and if you test it later, for example, in postman, it will work properly (given that the block is now in the past). But the caveat is that when we run it locally for some cases seems that `chain.height` is running in front of what tenderly is using as a pending block, so using something as `chain.height - 3` will do the job. However, if we leave out that param completely from the payload we will be using automatically the pending block at their end.

The other cases when our sim link was reverting with `Hash not approved` is due to the address we were using as a `from`, since we were using the first owner address from the safe and it is not always true that specific address is the one being the caller as per how [ape-safe](https://github.com/banteg/ape-safe/blob/master/ape_safe.py#L313-L324) works internally in the preview and receipt generation, that's why it is more robust to directly use the `receipt.sender` overall to avoid that kind of revert.

## Debank api support

Was playing locally to see if we could support the diff in $ val, but encountered that we cannot override storage in that case, left a `TODO` note for reference and added the code just not to leave it out in case that we find a workaround now or in the future they support the overrides

### DIFF
--- a/great_ape_safe/great_ape_safe.py
+++ b/great_ape_safe/great_ape_safe.py
@@ -191,8 +191,6 @@ class GreatApeSafe(ApeSafe):
             safe_tx, receipt = self._set_safe_tx_gas(
                 safe_tx, events, call_trace, reset, log_name, gas_coef
             )
-        if tenderly:
-            self._generate_tenderly_simulation(receipt, safe_tx.safe_tx_gas)
         if debank:
             self._debank_pre_execution(receipt, safe_tx.safe_tx_gas)
         if replace_nonce:
@@ -201,6 +199,8 @@ class GreatApeSafe(ApeSafe):
             pprint(safe_tx.__dict__)
         if hasattr(self, "snapshot"):
             self.print_snapshot(csv_destination)
+        if tenderly:
+            self._generate_tenderly_simulation(receipt, safe_tx.safe_tx_gas)
         if post:
             self.post_transaction(safe_tx)
 


great_ape_safe/great_ape_safe.py
```diff
@@ -253,23 +256,63 @@ def _generate_tenderly_simulation(self, receipt, gas):
...
-            "gas": gas,
+            "gas": 1_500_000 if gas == 0 else gas,
             "save": True,
             "save_if_fails": True,
+            # storage ref: https://github.com/safe-global/safe-contracts/blob/main/contracts/libraries/SafeStorage.sol
             "state_objects": {self.address: {"storage": {"0x04": "0x01"}}},
         }
         r = requests.post(api_url, headers=header, json=tx_payload)
-        r.raise_for_status()
+        if not r.ok:
```
gosuto-inzasheru, line 276: this is automatically included in calling `r.raise_for_status()`

petrovska-petro: you're [right](https://github.com/psf/requests/blob/main/requests/models.py#L756), feels redundant, remove in 0b62fc678bf1d546ca6d6ca6ba7713fe28c32887

great_ape_safe/great_ape_safe.py
```diff
@@ -192,6 +193,8 @@ def post_safe_tx(
             )
         if tenderly:
             self._generate_tenderly_simulation(receipt, safe_tx.safe_tx_gas)
```
gosuto-inzasheru, line 195: minor, but if this is moved to right before `if post` you dont need to scroll up to see the url

gosuto-inzasheru (commented):
great! this will save a lot of time copypasting all the console output to individual issues.

the debank idea would be of great added value, too bad it doesnt let itself be tricked :)

pro tip: cmd+doublclick in macos terminal opens a url in your browser :D

gosuto-inzasheru:
hmm if we would all use the same project on tenderly, we could start adding some labels to addresses to make the sims easier to read. here i do a tree claim for ibbtc msig and have labelled the ibbtc msig in wallets: https://dashboard.tenderly.co/gosuto/msig-sims/simulator/0b5d18c6-7109-4cd1-ab85-8ff3988b2a9c

jalbrekt85 (commented):
lgtm so far, doesn't look like theres an easy to solve that debank gs025 error from happening

great_ape_safe/great_ape_safe.py
petrovska-petro: moved it in bcb8d9f5c0b30e32ad5a2a147dc96a34d47e61dd

petrovska-petro: what are the thoughts on auto-opening to save those clicks 🤔 , something like:

```
import webbrowser

webbrowser.open(url)
```

petrovska-petro:
> hmm if we would all use the same project on tenderly, we could start adding some labels to addresses to make the sims easier to read. here i do a tree claim for ibbtc msig and have labelled the ibbtc msig and tree in wallets: https://dashboard.tenderly.co/gosuto/msig-sims/simulator/0b5d18c6-7109-4cd1-ab85-8ff3988b2a9c
> 
> Screenshot 2023-02-13 at 11 56 49

thats interesting, if we have a tenderly org where all of us has access to it to include labels etc, i think we should move then to use that same project

great_ape_safe/great_ape_safe.py
jalbrekt85: i like it, even more convenient. if chosen to use it, i would put it into a try/except block so that the whole tx doesn't crash just because it isn't able to open up in a web browser for whatever reason. also would want to check compatibility between different OS's

gosuto-inzasheru: tbh i think it is a bit overkill. cmd+doubleclick is not that much effort. main consideration for me here is that a lot of times when running a script i might not be interested in seeing the tenderly. then having to close the browser might actually create unnecessary actions in the workflow.

gosuto-inzasheru (approved):

jalbrekt85 (approved):

PR #1138, opened 2023-02-13
cowswap rate screening

jalbrekt85:
handles #1130

current flow:
check if involved assets exist on cg:
**yes** -> get cg rates and compare to cow, if diff is > than `pct_diff_threshold`, then confirm cow rate with user else continue as normal
**no** -> confirm cow rate with user

### DIFF
--- a/great_ape_safe/ape_api/cow.py
+++ b/great_ape_safe/ape_api/cow.py
@@ -111,7 +111,7 @@ class Cow:
             except KeyError:
                 has_cg_price = False
                 cow_sell_rate = (buy_amount_after_fee / 10 ** buy_decimals) / (
-                    mantissa_sell / 1e18
+                    mantissa_sell / 10 ** sell_decimals
                 )
                 if not Confirm.ask(
                     f"No cg rate found. Continue with cow rate of {cow_sell_rate} {sell_symbol}/{buy_symbol}?"
--- a/great_ape_safe/ape_api/cow.py
+++ b/great_ape_safe/ape_api/cow.py
@@ -33,8 +33,7 @@ class Cow:
 
         # determine api url based on current chain id and `prod` parameter
         chain_label = {1: "mainnet", 4: "rinkeby", 100: "xdai"}
-        prefix = "https://api.cow.fi/" if prod else "https://barn.api.cow.fi/"
-        self.api_url = f"{prefix}{chain_label[chain.id]}/api/v1/"
+        self.api_url = f"https://api.cow.fi/{chain_label[chain.id]}/api/v1/"
 
         self.cg = CoinGeckoAPI()
         self.cg_coin_list = {k["symbol"]: k["id"] for k in self.cg.get_coins_list()}
@@ -137,8 +136,8 @@ class Cow:
                 pct_diff = (cow_sell_rate - cg_sell_rate) / cg_sell_rate
 
                 if abs(pct_diff) > self.pct_diff_threshold:
-                    print(f"cow rate: {cow_sell_rate} {buy_symbol}/{sell_symbol}")
-                    print(f"cg rate: {cg_sell_rate} {buy_symbol}/{sell_symbol}")
+                    print(f"cow rate: {cow_sell_rate} {sell_symbol}/{buy_symbol}")
+                    print(f"cg rate: {cg_sell_rate} {sell_symbol}/{buy_symbol}")
                     direction = "lower" if pct_diff < 0 else "higher"
                     if not Confirm.ask(
                         f"cow rate is {direction} than cg by {round(abs(pct_diff), 2)}%, continue?"
@@ -205,29 +204,30 @@ class Cow:
             ):
                 raise
 
-        r = requests.post(self.api_url + "orders", json=order_payload)
-        if not r.ok:
-            r.raise_for_status()
+        if self.prod:
+            r = requests.post(self.api_url + "orders", json=order_payload)
+            if not r.ok:
+                r.raise_for_status()
 
-        order_uid = r.json()
-        print("ORDER RESPONSE")
-        pprint(order_uid)
-        print("")
+            order_uid = r.json()
+            print("ORDER RESPONSE")
+            pprint(order_uid)
+            print("")
 
-        # dump order to json and add staging label if necessary
-        path = "logs/trading/prod/" if self.prod else "logs/trading/staging/"
-        os.makedirs(path, exist_ok=True)
-        with open(f"{path}{order_uid}.json", "w+") as f:
-            f.write(json.dumps(order_payload))
+            # dump order to json and add staging label if necessary
+            path = "logs/trading/prod/"
+            os.makedirs(path, exist_ok=True)
+            with open(f"{path}{order_uid}.json", "w+") as f:
+                f.write(json.dumps(order_payload))
 
-        if origin != self.safe.address:
-            # can only sign if origin is safe
-            return order_payload, order_uid
+            if origin != self.safe.address:
+                # can only sign if origin is safe
+                return order_payload, order_uid
 
-        # pre-approve the order on-chain, as set by `signingScheme`: presign
-        # (otherwise signature would go in api order payload)
-        # https://docs.cow.fi/smart-contracts/settlement-contract/signature-schemes
-        self.settlement.setPreSignature(order_uid, True)
+            # pre-approve the order on-chain, as set by `signingScheme`: presign
+            # (otherwise signature would go in api order payload)
+            # https://docs.cow.fi/smart-contracts/settlement-contract/signature-schemes
+            self.settlement.setPreSignature(order_uid, True)
 
     def allow_relayer(self, asset, mantissa):
         """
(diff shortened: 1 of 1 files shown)


gosuto-inzasheru (commented):
i like it. 5% default threshold makes sense since onchain pricer has 5% slippage tolerance.

only thing still missing imo is that in a testing situation i would like to block all posts to the api, just to prevent spamming the order book. like rn there is no way for me to test run lets say `badger/convex_frax_tcl_strategy` without firing calls to the cow api.

maybe a flag? like `safe.init_cow(testrun=True)`? in which case no api posts are made. or lets repurpose the prod/staging flag for that? the whole posting to staging api has become useless.

jalbrekt85:
> maybe a flag? like `safe.init_cow(testrun=True)`? in which case no api posts are made. or lets repurpose the prod/staging flag for that? the whole posting to staging api has become useless anyway.

ya good idea, don't see why not. we can repurpose the `prod` flag so as to ensure compatibility with previous scripts.

great_ape_safe/ape_api/cow.py
```diff
@@ -95,6 +101,46 @@ def _sell(
...
+            buy_symbol, buy_decimals = buy_token.symbol(), buy_token.decimals()
+            sell_symbol, sell_decimals = sell_token.symbol(), sell_token.decimals()
+
+            try:
+                buy_token_id = self.cg_coin_list[buy_symbol.lower()]
+                sell_token_id = self.cg_coin_list[sell_symbol.lower()]
+            except KeyError:
+                has_cg_price = False
+                cow_sell_rate = (buy_amount_after_fee / 10 ** buy_decimals) / (
+                    mantissa_sell / 1e18
```
petrovska-petro, line 114:
```suggestion
                    mantissa_sell / 10 ** sell_decimals
```

great_ape_safe/ape_api/cow.py
```diff
@@ -95,6 +101,46 @@ def _sell(
...
+            if has_cg_price:
+                prices = self.cg.get_price([buy_token_id, sell_token_id], "usd")
+                buy_token_price = prices[buy_token_id]["usd"]
+                sell_token_price = prices[sell_token_id]["usd"]
+
+                cow_sell_rate = (buy_amount_after_fee / 10 ** buy_decimals) / (
+                    mantissa_sell / 1e18
+                )
+                cg_sell_rate = (
+                    (((mantissa_sell / 1e18) * sell_token_price) / buy_token_price)
```
petrovska-petro, line 130:
```suggestion
                    (((mantissa_sell / 10 ** sell_decimals) * sell_token_price) / buy_token_price)
```

great_ape_safe/ape_api/cow.py
```diff
@@ -95,6 +101,46 @@ def _sell(
...
+                    * coef
+                    / (mantissa_sell / 1e18)
+                )
+                pct_diff = (cow_sell_rate - cg_sell_rate) / cg_sell_rate
+
+                if abs(pct_diff) > self.pct_diff_threshold:
+                    print(f"cow rate: {cow_sell_rate} {sell_symbol}/{buy_symbol}")
```
petrovska-petro, line 137: here is not the rate expressed instead on: `{buy_symbol}/{sell_symbol}`?

jalbrekt85: yup, fixed in 5aa4523

jalbrekt85:
somehow i introduced a bug in the cg rate calc

great_ape_safe/ape_api/cow.py
gosuto-inzasheru: i dont think so. `cg_sell_rate` is calced as sell/buy on L129. eg:
selling 300 fxs
buying 2692 dai
rate is ~.11
ref https://explorer.cow.fi/orders/0x2ce41c34009c6b9b978764546a660bf165e2061db042cca8e2e2d48aa90dafebd0a7a8b98957b9cd3cfb9c0425abe44551158e9e63eb2898

jalbrekt85: ya realized this too after further testing, reverted in a61edf2

jalbrekt85:
> somehow i introduced a bug in the cg rate calc

fixed in a61edf2

jalbrekt85:
> > maybe a flag? like `safe.init_cow(testrun=True)`? in which case no api posts are made. or lets repurpose the prod/staging flag for that? the whole posting to staging api has become useless anyway.
> 
> ya good idea, don't see why not. we can repurpose the `prod` flag so as to ensure compatibility with previous scripts.

a61edf2

great_ape_safe/ape_api/cow.py
```diff
@@ -94,7 +99,51 @@ def _sell(
             mantissa_buy = int(mantissa_buy)
             buy_amount_after_fee = mantissa_buy
         else:
-            buy_amount_after_fee = int(int(fee_and_quote["quote"]["buyAmount"]) * coef)
+            buy_amount = int(fee_and_quote["quote"]["buyAmount"] * coef)
```
gosuto-inzasheru, line 102:
```suggestion
            buy_amount = int(int(fee_and_quote["quote"]["buyAmount"]) * coef)
```

great_ape_safe/ape_api/cow.py
```diff
@@ -94,7 +99,51 @@ def _sell(
...
+                pct_diff = (cow_sell_rate - cg_sell_rate) / cg_sell_rate
+
+                if abs(pct_diff) > self.pct_diff_threshold:
+                    print(
+                        f"cow rate (before slippage): {cow_sell_rate} {sell_symbol}/{buy_symbol}"
+                    )
+                    print(f"coingecko rate: {cg_sell_rate} {sell_symbol}/{buy_symbol}")
+                    direction = "lower" if pct_diff < 0 else "higher"
+                    if not Confirm.ask(
+                        f"cow rate is {direction} than cg by {round(abs(pct_diff), 2)}%, continue?"
```
gosuto-inzasheru, line 144:
```suggestion
                        f"cow rate is {direction} than cg by {round(abs(pct_diff * 100), 2)}%, continue?"
```

great_ape_safe/ape_api/cow.py
```diff
@@ -94,7 +99,51 @@ def _sell(
             mantissa_buy = int(mantissa_buy)
             buy_amount_after_fee = mantissa_buy
         else:
-            buy_amount_after_fee = int(int(fee_and_quote["quote"]["buyAmount"]) * coef)
+            buy_amount = int(fee_and_quote["quote"]["buyAmount"] * coef)
+            buy_amount_after_fee = int(buy_amount * coef)
```
gosuto-inzasheru, line 103: multiplying by `coef` twice now

jalbrekt85: oops; 0b831ac

great_ape_safe/ape_api/cow.py
gosuto-inzasheru: petrovska is right here, since:
```
cow_sell_rate = (buy_amount / buy_divisor) / (
    mantissa_sell / sell_divisor
)
```
eg
```
cow rate (before slippage): 9.972969031236483 FXS/DAI
coingecko rate: 10.032728902261416 FXS/DAI
cow rate is lower than cg by 0.6%, continue? [y/n]: 
```

which should be DAI/FXS (ie 10 dai per fxs)

gosuto-inzasheru:
rest lgtm now, just the symbols' order

great_ape_safe/ape_api/cow.py
jalbrekt85: we are just comparing rates, not sure how important which asset is being bought/sold is on the ordering here. 
i think the confusion comes from how we are interpreting "ASSET1/ASSET2". coingecko uses this interpretation for example:
[ETH/USDC -> 1550](https://www.coingecko.com/en/coins/ethereum/usd)
[USDC/ETH -> 0.00064](https://www.coingecko.com/en/coins/usd-coin/eth)
basically saying ASSET1 is equal to this many ASSET2

the other way to interpret is like you said as "this many ASSET1 is equal to ASSET 2". this way seems more intuitive if you are reading it more literally from left to right, but the other interpretation is what ive seen exchanges use 🤷

regardless on interpretation, I think what we are also trying to avoid is having it spit out something like 0.0006 ETH/USDC as opposed to 1550 ETH/USDC (this depends on if we are selling or buying the stablecoin currently). so maybe the ordering should be decided programmatically: 
if price of ASSET1 > ASSET2 then have the rate of ASSET1 denominated in ASSET2 else other way around

petrovska-petro: i see the rationale, somehow got mix-up on how the rate was calculate and how it will be more intuitive to express it, but like keeping it up like cg interprets it

would not say that its req to have a programmatically way to decide how the ratio is presented, depends on the order itself sometimes may be looking like 0.xxxx, others 1.xxx etc depending on the pricing of asset sold vs bought

petrovska-petro (approved):

gosuto-inzasheru (approved):

Issue #1133, opened 2023-02-09
UpkeepManager integration


petrovska-petro:
The [UpkeepManager](https://github.com/Badger-Finance/badger-avatars/blob/main/src/upkeeps/UpkeepManager.sol) contract has passed the SC pod reviews and its ready for integration. Required steps:

- [x] Contract deployment with techops as governance
- [x] Initializing base upkeep setup for testing with dummy member
- [x] Cancel dummy member in upkeep manager
- [x] Cancel current upkeeps to migrate to upkeep manager, update watchlist and update keeper registry in gas st
- [x] After >50 blocks add members from existing upkeep jobs which in the future will require auto top-ups 
<img width="1372" alt="Screenshot 2023-02-09 at 18 32 54" src="https://user-images.githubusercontent.com/84875062/217788150-8284ef48-8a32-4965-91cd-09d7850fb1a8.png">

Have not found so far a way to enforce the wd of link funds without [cancelling](https://docs.chain.link/chainlink-automation/automation-economics) to sim an underfunded upkeep id, perhaps for round test makes sense to deploy a dummy keeper compliant contract

petrovska-petro:
contract deployed and verify: https://etherscan.io/address/0x4c02f0160dc0387b13bcb5e1728c780649e109ac

petrovska-petro:
init upkeep manager: https://etherscan.io/tx/0x00481adda6394d80f7bbcf536ccb51c8c01079039360228c66b0ea282f225a90

<img width="1374" alt="Screenshot 2023-02-20 at 22 47 59" src="https://user-images.githubusercontent.com/84875062/220143014-6fa6222a-b858-42f5-a9bb-84ca124748d3.png">

petrovska-petro:
beautiful to see on action the top-ups already due to the gas spike and given that techops send only a bit of link funds to the upkeep manager

https://etherscan.io/tx/0xb0d87c66d0e04a3f47ace41fffe01b16217625aa1a20c7e40a899b8a586d0431 -> for dummy tree dripper 
https://etherscan.io/tx/0x706cfccff2e32d9f3f2ff40f7620a96ed7f321854bd3341b8a21c44b3f764ca0 -> for upkeep id master

petrovska-petro:
cancelattion of dummy member: https://dashboard.tenderly.co/testing-ape/project/simulator/168bb7e7-63f3-48c3-9468-7fa50d4831d7

petrovska-petro:
[...]

PR #1147, opened 2023-02-17
chore: add exec to addr book for gas station

gosuto-inzasheru:
prep for #1133 (updates gas watchlist) and #1144 (make sure acl actors have gas)


petrovska-petro (approved):
confirmed ops_exec15 is a techops signer ✅

sajanrajdev (approved):
- Addition of new executor to address book for gas station covering. Confirmed address.
- Add exception to tx post to only generate Tenderly simulation when skip preview is not enabled.

LGTM

Issue #1157, opened 2023-02-28
bug: cow.py sometimes gets wrong cg id


jalbrekt85:
cg id is retrieved from token symbol, but multiple tokens can have the same symbol, making the _correct_ id ambiguous

PR #1158, opened 2023-02-28
query cg price api by contract address in cow.py

jalbrekt85:
bug explained in: #1157

### DIFF
--- a/great_ape_safe/ape_api/cow.py
+++ b/great_ape_safe/ape_api/cow.py
@@ -6,7 +6,6 @@ from pprint import pprint
 
 from brownie import Contract, chain, interface, web3
 from rich.prompt import Confirm
-import requests
 
 from helpers.addresses import registry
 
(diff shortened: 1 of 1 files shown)


petrovska-petro (commented):
agree that using using the contract addr ensures that there is not duplicates of symbol ✅ 

btw, in L3 already there is imported `import requests`, no req in L9

great_ape_safe/ape_api/cow.py
```diff
@@ -6,7 +6,7 @@
 
 from brownie import Contract, chain, interface, web3
 from rich.prompt import Confirm
-from pycoingecko import CoinGeckoAPI
+import requests
```
gosuto-inzasheru:
```suggestion
```

gosuto-inzasheru (approved):
bug occurred with aura, tested below:
```
>>> from great_ape_safe.ape_api import cow
>>> from great_ape_safe import GreatApeSafe
>>> c = cow.Cow(GreatApeSafe('treasuryvault.badgerdao.eth'))
>>> c.get_cg_price('0xC0c293ce456fF0ED870ADd98a0828Dd4d2903DBF')
3.07
```
lgtm!

petrovska-petro (approved):

Issue #1161, opened 2023-03-03
Add Bunni class in `ape_api` 


petrovska-petro:
- [x] Class should enable to:
  - Add liquidity
  - Remove liquidity
  - Stake
  - Unstake
  - Claim rewards

Considerations:

- [x] While claiming the rewards important visualising current discount of options provided by the [oracle](https://etherscan.io/address/0x9d43ccb1aD7E0081cC8A8F1fd54D16E54A637E30#readContract), and ideally assert in the scenario where the `multiplier` has malicious setting or alert in cli

- [x] Worth also updating the [script](https://github.com/Badger-Finance/badger-multisig/blob/main/scripts/badger/bribe_ecosystems.py) to allow posting incentives for bunni hh

PR #1170, opened 2023-03-13
bunni class + tests

jalbrekt85:
#1161 - WIP

### DIFF
--- a/great_ape_safe/ape_api/uni_v3.py
+++ b/great_ape_safe/ape_api/uni_v3.py
@@ -92,6 +92,42 @@ class UniV3:
                 path_encoded += int.to_bytes(item, 3, byteorder="big")
         return path_encoded
 
+    def calc_min_amounts(
+        self, pool, token0_amount, token1_amount, lower_tick, upper_tick
+    ):
+        sqrtRatioX96, currentTick, _, _, _, _, _ = pool.slot0()
+        sqrtRatio_lower_tick = getSqrtRatioAtTick(lower_tick)
+        sqrtRatio_upper_tick = getSqrtRatioAtTick(upper_tick)
+
+        amount0Min = 0
+        amount1Min = 0
+
+        liquidity = maxLiquidityForAmounts(
+            sqrtRatioX96,
+            sqrtRatio_lower_tick,
+            sqrtRatio_upper_tick,
+            token0_amount,
+            token1_amount,
+        )
+
+        if currentTick < lower_tick:
+            # calc amount0Min
+            amount0Min = getAmount0Delta(
+                sqrtRatio_lower_tick, sqrtRatio_upper_tick, liquidity
+            )
+            amount1Min = 0
+        elif currentTick < upper_tick:
+            # calc both
+            amount0Min = getAmount0Delta(sqrtRatioX96, sqrtRatio_upper_tick, liquidity)
+            amount1Min = getAmount1Delta(sqrtRatio_lower_tick, sqrtRatioX96, liquidity)
+        else:
+            # calculate amount1Min
+            amount0Min = 0
+            amount1Min = getAmount1Delta(
+                sqrtRatio_lower_tick, sqrtRatio_upper_tick, liquidity
+            )
+        return amount0Min, amount1Min
+
     def burn_token_id(self, token_id, burn_nft=False):
         """
         It will decrease the liquidity from a specific NFT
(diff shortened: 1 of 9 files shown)


great_ape_safe/ape_api/uni_v3.py
petrovska-petro, line 1: while updating this class i think is worth updating how the fee tier in the minting method gets pass

currently is [hardcoded](https://github.com/Badger-Finance/badger-multisig/blob/feat/bunni/great_ape_safe/ape_api/uni_v3.py#L356). it can be read directly from the pool instead

petrovska-petro:
> * expectedPaymentAmount

the calculation for exercising the option would not be something like:

```
price = oracle.getPrice() / 1e18
expected_payment_amount = self.olit.balanceOf(self.safe) * price
```

great_ape_safe/ape_api/uni_v3.py
jalbrekt85: good catch

jalbrekt85:
> > expectedPaymentAmount
> 
> the calculation for exercising the option would not be something like:
> 
> ```
> price = oracle.getPrice() / 1e18
> expected_payment_amount = self.olit.balanceOf(self.safe) * price

nice. by making sure to round up, this works smoothly

jalbrekt85:
> > min amounts during withdrawal
> 
> we could leverage their [lens](https://etherscan.io/address/0xb73F303472C4fD4FF3B9f59ce0F9b13E47fbfD19#readContract) contract, given our amount of shares we should be able to calc how much token0 and token1 we own from the total of a specific key considering the ppfs i think
> 
> test [ref](https://github.com/ZeframLou/bunni/blob/main/src/tests/BunniHub.t.sol#L233-L254)

overlooked this sc, but does indeed calc correct min amounts using ppfs, nice.

jalbrekt85:
> > expectedPaymentAmount
> 
> the calculation for exercising the option would not be something like:
> 
> ```
> price = oracle.getPrice() / 1e18
> expected_payment_amount = self.olit.balanceOf(self.safe) * price
> ```
> 
> > min amounts during withdrawal
> 
> we could leverage their [lens](https://etherscan.io/address/0xb73F303472C4fD4FF3B9f59ce0F9b13E47fbfD19#readContract) contract, given our amount of shares we should be able to calc how much token0 and token1 we own from the total of a specific key considering the ppfs i think
> 
> test [ref](https://github.com/ZeframLou/bunni/blob/main/src/tests/BunniHub.t.sol#L233-L254)

fixed both calcs in eb07bb5

great_ape_safe/ape_api/uni_v3.py
jalbrekt85: eb07bb5

tests/bunni/test_flow.py
```diff
@@ -0,0 +1,73 @@
...
+
+    assert gauge.balanceOf(safe) < before_gauge
+    assert bunni_token.balanceOf(safe) > before_bunni
+
+
+def test_compound(bunni, badger, wbtc):
+    badger.approve(bunni.router, 0)
+
+    # simulate some trading so we have some fees to compound
+    bunni.swap([badger, wbtc], 1000e18)
```
petrovska-petro, line 70: the intention here i think is to use another class (univ3), it reverts locally

petrovska-petro: nvm i see the inheritance, does it revert at your end?

jalbrekt85: works on my end, all tests passing currently

petrovska-petro: not blocking, since its not inherited from bunni class itself, but locally for me was reverting

Screenshot 2023-03-15 at 01 05 17

Screenshot 2023-03-15 at 01 06 57

petrovska-petro: reset my config and delete `deployments.db` still the revert persist somehow

tried with `min_out=0` for the wbtc -> badger swap. can confirm that the test passes successfully ✅

petrovska-petro (approved):
Great work on the class! 

- allows to successfully interact with bunni with safety min amount measures
- claim rewards from specific gauge and prompts the discount when below 50
- exercise option based on balancer oracle pricing

tests/bunni/test_flow.py
gosuto-inzasheru: im getting the same. somehow the second `swap` call is causing this. if however, i only make one swap, it reverts on `assert added_liq > 0`; `self.hub.compound(self.bunni_key).return_value` comes back as `(0, 0, 0)`

gosuto-inzasheru (approved):
great contrib. adding tests is def a great addition here, if not a must.

lets see it in action!

tests/bunni/test_flow.py
jalbrekt85: very strange. obviously these swaps will happen naturally on mainnet, so not breaking. maybe we need some univ3 tests

PR #1171, opened 2023-03-13
Precaution measure - euler badger market

petrovska-petro:
```
brownie run badger/withdraw_euler_position
```


GalloDaSballo:
LGTM

Calls `withdraw` on the token with 98% of the balance available at time of posting

dapp-whisperer (approved):

Issue #1166, opened 2023-03-09
add remaining gtc to univ3 while it is out of range


gosuto-inzasheru:
https://app.uniswap.org/#/pool/435187

https://etherscan.io/token/0xde30da39c46104798bb5aa3fe8b9e0e1f348163f?a=0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e

jalbrekt85:
```py
{
│   'ethereum_client': <gnosis.eth.ethereum_client.EthereumClient object at 0x14a0d7ac0>,
│   'safe_address': '0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e',
│   'to': '0x40A2aCCbd92BCA938b02010E17A5b8929b49130D',
│   'value': 0,
│   'data': HexBytes('0x8d80ff0a000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000001b200de30da39c46104798bb5aa3fe8b9e0e1f348163f00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000044095ea7b3000000000000000000000000c36442b4a4522e871399cd717abdd847ab11fe8800000000000000000000000000000000000000000000042e11fdf98ea32a10ea00c36442b4a4522e871399cd717abdd847ab11fe88000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000c4219f5d17000000000000000000000000000000000000000000000000000000000006a3f3000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000042e11fdf98ea32a10ea0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000418ab3b7490e7c10a8000000000000000000000000000000000000000000000000000000000640fd1a30000000000000000000000000000'),
│   'operation': 1,
│   'safe_tx_gas': 0,
│   'base_gas': 0,
│   'gas_price': 0,
│   'gas_token': '0x0000000000000000000000000000000000000000',
│   'refund_receiver': '0x0000000000000000000000000000000000000000',
│   'signatures': b'',
│   '_safe_nonce': 132,
│   '_safe_version': '1.3.0',
│   '_chain_id': None
}
snapshot result for 0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e:
                   balance_before         balance_after               balance_delta
symbol                                                                             
[...]

PR #1172, opened 2023-03-13
deposit remaining gtc in uni v3 position

jalbrekt85:
handles #1166


gosuto-inzasheru (approved):
looks good, lets get it going before it is in range again

petrovska-petro (approved):
Top-ups correctly the range with all gtc in the addr ✅

PR #1181, opened 2023-03-21
sweep badger from voter to treasury

jalbrekt85:
check for badger to sweep from voter in weekly aura tcl strat script


petrovska-petro (approved):
moves badger fees from the incentives processors to vault msig ✅

gosuto-inzasheru (approved):

Issue #1189, opened 2023-03-23
Refine max amount to wd from euler emerg script


petrovska-petro:
Currently the script was done for the emerg situation and the calculation for the amount to wd is a bit naive. Consider:

- utilisation %
- ownership % of the market

PR #1192, opened 2023-03-23
feat: add two approach for wd: softer and max-out

petrovska-petro:
Tackles #1189 

~The pr tries to tackle imo two different types of withdrawals that we can have, the softer considering ownership and utilisation ratios and the max-out approach, which solely cares about what is available in the market as idle capital and check the minimum between our total “ebalance” and the available amount.~

~Somehow, made me realised that even at the time of emerg, we kind of tackle the best outcome scenario for us, only it could get reverted if other depositors tried to wd at the same time interval between [t_tx_posting, t_tx_exec] and its more than the `EXEC_SLIPPAGE` reduction.~

Let me know if you think there is a neater approach or I am missing some caveat. read comments below

--------

Uses the approach of the min amount between the max avail amount in the market and the max user balance

### DIFF
--- a/interfaces/euler/IDebtToken.sol
+++ b/interfaces/euler/IDebtToken.sol
@@ -1,0 +1,66 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity ^0.8.0;
+
+/// @notice Tokenised representation of debts
+interface IDebtToken {
+    /// @notice Debt token name, ie "Euler Debt: DAI"
+    function name() external view returns (string memory);
+
+    /// @notice Debt token symbol, ie "dDAI"
+    function symbol() external view returns (string memory);
+
+    /// @notice Decimals of underlying
+    function decimals() external view returns (uint8);
+
+    /// @notice Address of underlying asset
+    function underlyingAsset() external view returns (address);
+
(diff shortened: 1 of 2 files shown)


gosuto-inzasheru (commented):
you cant withdraw more than what is idle in the market, right? and neither does it make sense to try to 'max out' the whole idle market if the user isnt able to withdraw that much?

so instead of a boolean, i think you can just calc both `max_market_withdrawal` and `max_user_withdrawal` and take the `min()` of those two.

no need for slippage then either imo, if another user withdraws before us for sure it is going to be more than 1-5% of the amount we were about to withdraw.

do you test run this on a particular block btw?

gosuto-inzasheru (approved):
yes looking good now and more robust for different types of market conditions

since amount of actors and actions in this market are so low, i dont think we will run into frontrun problems with this, or if we do it would not have been avoidable by using a slippage.

```
                       balance_before                 balance_after               balance_delta
symbol                                                                                         
BADGER   1,609,031.199754034482789187  1,691,717.313034791075270975   82,686.113280756592481788
eBADGER    100,000.000000000000000000     18,014.500805577505110672  -81,985.499194422494889328
```

jalbrekt85 (approved):
nice, agree this is more robust, lgtm

Issue #1180, opened 2023-03-21
week 21: sell all influence tokens from fees


gosuto-inzasheru:
need to rerun #1010 before end of quarter:
> claim from tree and sell all accumulated influence tokens to stablecoins:
    (b)cvxcrv
    (b)aurabal
    (gravi)aura
    (bve)cvx

gosuto-inzasheru:
@petrovska-petro what do you think is the chance of running this through the fee processor module next week or so?

petrovska-petro:
> @petrovska-petro what do you think is the chance of running this through the fee processor module next week or so?

the fee-module we divided in two iterations to avoid bloating the initial scope, if the [PR-7](https://github.com/Badger-Finance/badger-safe-modules/pull/7) gets merged which was the initial step to wd from the positions, the second iteration was meant to be hook it up with cowswap selling. So, this time still better to run the script as in q4

gosuto-inzasheru:
what is blocking pr7? camos review? could be interesting to try to get it deployed and use it no..?

petrovska-petro:
> what is blocking pr7? camos review? could be interesting to try to get it deployed and use it no..?

One pendant review from @sajanrajdev, he will get it done today ideally, then we can proceed with integration

gosuto-inzasheru:
note: sell to usdc instead of dai, since dai is our biggest stablecoin position already

jalbrekt85:
```py
{
│   'ethereum_client': <gnosis.eth.ethereum_client.EthereumClient object at 0x13629f3a0>,
│   'safe_address': '0x042B32Ac6b453485e357938bdC38e0340d4b9276',
│   'to': '0x40A2aCCbd92BCA938b02010E17A5b8929b49130D',
│   'value': 0,
[...]

PR #1195, opened 2023-03-25
q1 claim and sell influence

jalbrekt85:
adjusts `claim_and_sell_influence.py` for #1180


gosuto-inzasheru (commented):
missing bvecvx action here.

let me make sure what treasury wants to do with it; either unwrap and sell to usdc or otherwise i think it should go to treasury voter, either as a deposit into the stability pool or just for vote weight holdings.

i ran into some coingecko error as well, had to implement the pro api. will pr into this branch.

rest looks good

petrovska-petro (approved):

Issue #1190, opened 2023-03-23
Include boolean flag in emerg scripts


petrovska-petro:
Allows to:

1. Unstake to lp position

or 

2. Withdraw all the way down to the underlings

PR #1196, opened 2023-03-26
calc curve withdraw amounts, add unwrap option to emergency scripts

### LINKED ISSUE #1188: Add slippage control in Curve class
Currently it is defaulted to zero min amount out. [ref](https://github.com/Badger-Finance/badger-multisig/blob/main/great_ape_safe/ape_api/curve.py#L169)

jalbrekt85:
closes #1188 #1190

### DIFF
--- a/great_ape_safe/ape_api/curve.py
+++ b/great_ape_safe/ape_api/curve.py
@@ -1,8 +1,9 @@
-from argparse import ArgumentError
-import numpy as np
-from helpers.addresses import registry
 from brownie import Contract, ZERO_ADDRESS, interface
 from brownie.exceptions import VirtualMachineError
+import numpy as np
+
+from great_ape_safe.ape_api.helpers.coingecko import get_cg_price
+from helpers.addresses import registry
 
 
 class Curve:
@@ -24,8 +25,9 @@ class Curve:
     def _get_coins(self, lp_token):
         # get coin addresses from registry for a specific `lp_token`
         pool = self._get_pool_from_lp_token(lp_token)
-        true_length = self.registry.get_n_coins(pool)[0]
-        return [pool.coins(i) for i in range(true_length)]
+        # true_length = self.registry.get_n_coins(pool)[0]
+        length = self._get_n_coins(lp_token)
+        return [pool.coins(i) for i in range(length)]
 
     def _get_registry(self, pool):
         # get corresponding registry of lp token, either the 'normal' one,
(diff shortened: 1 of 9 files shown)


great_ape_safe/ape_api/curve.py
```diff
@@ -163,10 +165,20 @@ def withdraw(self, lp_token, mantissa):
...
-        n_coins = self._get_n_coins(lp_token)
-        # TODO: slippage and stuff
-        minima = list(np.zeros(n_coins))
+        coins = self._get_coins(lp_token)
+
+        pool_usd_values = [
+            get_cg_price(x)
+            * (pool.balances(i) / 10 ** self.safe.contract(x).decimals())
+            for i, x in enumerate(coins)
+        ]
```
jalbrekt85, line 173: ideally here we would use `pool.price_oracle` instead of coingecko. the problem is for some functions, i couldnt find a clear mapping from coin address to its index. for example, for [tricrypto2 pool](https://etherscan.io/address/0xa895B89D74a6BC23a284a0526e123ea776674cF5#readContract) index `0`:
- `pool.price_oracle(0)` -> wbtc price
- `pool.coins(0)` -> usdt address

is there an intuitive way to do this programmatically?

petrovska-petro: one think to consider is that not all pools has as a feature the internal oracle, e.g: [bveCVX_CVX_f](https://etherscan.io/address/0x04c90C198b2eFF55716079bc06d7CCc4aa4d7512#readContract) 

what are your thoughts on doing the dry-run of the method itself with [call](https://eth-brownie.readthedocs.io/en/stable/api-network.html#ContractTx.call) and use the returned amounts as the minima (those amounts given back can be seen as non-manipulated) and in the particular cases where curve method may not return anything, which i recall there was some cases fallback to the cg/oracle approach?

jalbrekt85: great idea, that should work. will make it foolproof too

jalbrekt85: 69fbf12

jalbrekt85:
also added changes from #1200 to upgrade cg api to pro in this pr since i refactored the `get_cg_price` into its own helper function

petrovska-petro (approved):
Great to add-on wd min amount protection in the curve class ✅ 

- add cg api key and separate helper 
- modify curve class with minimas via dry-run and fallback when pool does not return into cg approach estimations
- adds a flag to allow fully unwrap positions in emerg optionally


Note: make me wonder if historically on the prev txs we ever got rekt meaningfully with any of the normal wd with zero minimums 🤔

gosuto-inzasheru (approved):
lgtm, confirmed `minima` gets populated during tests

PR #1197, opened 2023-03-26
organize ape apis in separate class

jalbrekt85:
also undos changes so that the `init_` methods are defined explicitly. rationale here: https://discord.com/channels/1011322803414388897/1011333454740340756/1088208597986578582

### DIFF
--- a/great_ape_safe/ape_api/__init__.py
+++ b/great_ape_safe/ape_api/__init__.py
@@ -39,8 +39,8 @@ class ApeApis:
     def init_balancer(self):
         self.balancer = Balancer(self)
 
-    def init_bunni(self):
-        self.bunni = Bunni(self)
+    def init_bunni(self, bunni_token_addr=None, pool_addr=None, range0=None, range1=None):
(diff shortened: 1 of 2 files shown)


gosuto-inzasheru (approved):
agree that explicit is cleaner

great_ape_safe/ape_api/__init__.py
```diff
@@ -1,11 +1,94 @@
...
+    def init_chainlink(self):
+        self.chainlink = Chainlink(self)
+
+    def init_compound(self):
+        self.compound = Compound(self)
+
+    def init_convex(self):
+        self.convex = Convex(self)
+
+    def init_cow(self):
```
petrovska-petro, line 54: for the cases like cow or bunni that may be expecting another param this may not work, no?

e.g : `prod` or `bunni_token_addr`

jalbrekt85: good catch 92ab8c4

petrovska-petro (approved):

gosuto-inzasheru (approved):
nice one i missed that!

PR #1200, opened 2023-03-27
feat: add coingecko pro api to cow

gosuto-inzasheru:


jalbrekt85:
added your changes from this pr to #1196 instead, see: https://github.com/Badger-Finance/badger-multisig/pull/1196#issuecomment-1486121267. this can be closed

Issue #1211, opened 2023-03-31
Adapt incentive script to accept amount in dollar for bunni hh


petrovska-petro:
Calculation of total badger occurs internally in the script

PR #1212, opened 2023-03-31
feat: allow passing arg as dollar denominated for bunni incentives

petrovska-petro:
Tackles #1211 

Since the TCD allows for $1-$5k/week, updating the script in this way will facilitate the runs imo


jalbrekt85 (approved):
nice addition, lgtm

gosuto-inzasheru (approved):
runs good onchain!

PR #1217, opened 2023-04-03
speed up aura/convex claiming process

### LINKED ISSUE #1205: investigate speeding up convex/aura claiming process
likely culprit is the potentially long looping process:
https://github.com/Badger-Finance/badger-multisig/blob/8952336e584f80a11ba2a000b302fc05613fa597/great_ape_safe/ape_api/convex.py#L99-L102

one solution is to allow to pass reward addresses to `claim_all` so that searching process can be skipped

### LINKED ISSUE #1206: `wd_unlocked_aura` func for voter needs to handle graviaura withdrawal

jalbrekt85:
closes #1205

edit: also closes #1206

### DIFF
--- a/great_ape_safe/ape_api/aura.py
+++ b/great_ape_safe/ape_api/aura.py
@@ -23,24 +23,24 @@ class Aura(Convex):
     def _alert_not_relevant():
         print(" ==== NOT RELEVANT METHOD IN AURA CLASS ==== \n")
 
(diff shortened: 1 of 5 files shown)


gosuto-inzasheru (approved):
runs good, withdraws all graviaura in voter

petrovska-petro (approved):
great performance improvement 🏃 🏃 ✅

Issue #1220, opened 2023-04-04
buy necessary $grt for utility


gosuto-inzasheru:
https://docs.google.com/document/d/1G5M1jPpLkmH0CbzDatlZdpVl4cP9wDQQ8DtEee9ygLI/edit#heading=h.7dxtka40mkdz

<img width="833" alt="Screenshot 2023-04-05 at 02 04 41" src="https://user-images.githubusercontent.com/2835259/229865540-dbd4e806-8128-4aac-9ded-0980c66ba43e.png">

PR #1222, opened 2023-04-06
feat: buy grt

gosuto-inzasheru:
in order to solve #1220 

note: destination is techops, since it will be the account holder on the graph dashboard.


gosuto-inzasheru:
also note i included the poetry pr already since otherwise i cant run scripts on this laptop

petrovska-petro (approved):
confirmed grt addr and order details based on TCD ✅

jalbrekt85 (approved):
lgtm

Issue #1216, opened 2023-04-03
Create `bunni_tcl_strategy` script 


petrovska-petro:
Rewards are starting to flow in badgerwbtc gauge:

<img width="243" alt="Screenshot 2023-04-03 at 19 35 25" src="https://user-images.githubusercontent.com/84875062/229485916-47ae22b1-d898-4e9f-9422-b286195a91ba.png">

From the [proposal](https://forum.badger.finance/t/27-deposit-500k-in-bunni/5960#proposal-2), perhaps is worth having both possibilities for usdc or weth depending on a flag 

Note: due to rewards volume and main [pool](https://app.balancer.fi/#/ethereum/pool/0x9232a548dd9e81bac65500b5e0d918f8ba93675c000200000000000000000423) best processing for weth

PR #1223, opened 2023-04-06
feat: include bunni tcl strat script

petrovska-petro:
Tackles #1216 

```
brownie run scripts/badger/bunni_tcl_strategy
```


gosuto-inzasheru:
getting an assertion error on `bunny.py#L249`, or if commented out `IOptionsToken.exercise` reverts on a `typed error: 0xd193bd20`.

jalbrekt85:
> getting an assertion error on `bunny.py#L249`, or if commented out `IOptionsToken.exercise` reverts on a `typed error: 0xd193bd20`.

is due to an oversight on slippage control. tests were with small claims so the `math.ceil` was enough to pass. updated to use a wider slipp calc so to account for our large claims: 9801892

gosuto-inzasheru:
@jalbrekt85 why comment out the assert though?

gosuto-inzasheru (approved):
runs successful now.

harvests bunni and converts to weth!

Issue #1234, opened 2023-04-17
sign sweep txs on wbtc big dao msig


gosuto-inzasheru:
see https://github.com/WrappedBTC/DAO/issues/17#issuecomment-1500847880

gosuto-inzasheru:
i think signing tx id 1, 2 and 3 atomically would be most efficient

gosuto-inzasheru:
posted nonce 519 to techops: https://dashboard.tenderly.co/gosuto/msig-sims/simulator/abcf9777-f9ae-4560-90f6-2b212ee0c925

gosuto-inzasheru:
execd: https://etherscan.io/tx/0xdccb74645f0dbddcf79e4cf20cb3b7cdfa66bfb52231660fc2a1e4583833ae43

PR #1240, opened 2023-04-20
feat: sign txs on wbtc dao msig

gosuto-inzasheru:
solution to #1234


petrovska-petro (approved):
confirms pendant txs (1,2) which currently have 2 confirmations. funds when claimed will be transfer to the owner of the controller which is the `r.wbtc.dao_multisig` ✅

jalbrekt85 (approved):

Issue #1105, opened 2023-01-23
sunset arb1:dev deprecated msig


gosuto-inzasheru:
`"dev_multisig_deprecated": "0x468A0FF843BC5D185D7B07e4619119259b03619f"` is not a gnosis safe, and is not being used anymore.

we should sunset it the same way we are doing with others (atomically?);
- add dev msig to owners
- set threshold to 1
- remove all other owners

gosuto-inzasheru:
NOTE: #1139 needs to be completed _before_ this deprecation can take place!

gosuto-inzasheru:
NOTE: ignore nonce 196-203, these were erroneously posted

gosuto-inzasheru:
- make dev msig owner: https://arbiscan.io/tx/0xa0687de01253f740a6549e69600965c42f11a2711e24ad7f25ed71e847b65870#eventlog
- set threshold to 1: https://arbiscan.io/tx/0xe399d0b3bbd3e875504c5aa68d595546ac0b46c610a1d95edc7a7406d24c2c5e#eventlog
- remove owners:
  - https://arbiscan.io/tx/0x91c43541d236aaf2ea661d3c7e4d4883babdd88cb6d7b0d3e7f8a42e16587c6d#eventlog
  - https://arbiscan.io/tx/0x4a8b318dbff322d342d887f676d9e8b721b5b8487f8fae3a45b52381de223fc3#eventlog
  - https://arbiscan.io/tx/0xa9e818aee407d67db23f4578fec237660c3e66c6d6150880439efa34f9acb734#eventlog
  - https://arbiscan.io/tx/0xc045b3effa34253fbddb970e107d1c9aff1cf530467ab950e832a7d2a99c09d0#eventlog
  - https://arbiscan.io/tx/0x1faaac29e5dc0f12c85ddb7758eb665b27c6b711927554950c0088f202709fd2#eventlog
  - https://arbiscan.io/tx/0xae79d57251b6fe9b1558282c55d4f9cb3ae525a679a2d9e1abbd7894f919dab9#eventlog
  - https://arbiscan.io/tx/0xc502f63d988b35e85c31e6c18a2f0312506d79b1ac428a672b99c2d112e3f70c#eventlog

<img width="363" alt="Screenshot 2023-05-19 at 22 00 56" src="https://github.com/Badger-Finance/badger-multisig/assets/2835259/5cefede6-766f-47c4-aa41-1eb254b65b96">
[...]

PR #1249, opened 2023-04-27
feat: sunset deprecated dev msig on arb1

gosuto-inzasheru:
@petrovska-petro this needs to be posted in a different way then usual.

the script currently assumes one of the owners can be loaded into brownie and post all txs that way. maybe not the safest way. another option would be to add frame support.

how did we do this in the past?

solution to #1105

### DIFF
--- a/scripts/issue/1105/sunset_arb1_dev_deprecated.py
+++ b/scripts/issue/1105/sunset_arb1_dev_deprecated.py
@@ -1,0 +1,73 @@
+from brownie import accounts, interface
+
+from helpers.addresses import r
+
+
+def sim_exec(safe):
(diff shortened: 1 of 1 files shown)


petrovska-petro (commented):
logic overall looks good, will say that if one of the current owner is an `deployer` could do the each submission of this tx in order, otherwise frame is also a possibility

gosuto-inzasheru:
ah yes i see `deployer4` being an owner here. perfect.

petrovska-petro (approved):
sim asserts ensures right config to deprecated the msig, leaving in the sim the signer as the last owner to be removed. ensure only 1 as threshold and current dev_msig as unique owner ✅ 

it can be posted from `ops_deployer4`  ✅

PR #1258, opened 2023-05-03
chore: updates needed for latest gravi claim

gosuto-inzasheru:
- coef of `.96` created too much slippage and caused the onchain pricer to revert. this is fixed with a tighter slippage.
- update votium submodule
- add alcx to gravi's claimable tokens


petrovska-petro (approved):
alcx addr confirmed ✅

Issue #1005, opened 2022-12-07
investigate old geyser 


gosuto-inzasheru:
`0xd34C1d3853214bf049B760Ef48A580bfa7A9c8a1`

i think this is an old geyser for the sushi wbtcbadger strat. still holds ~6k badger; there might be a way to retrieve these tokens.

gosuto-inzasheru:
the solution used by https://github.com/Badger-Finance/badger-multisig/pull/539 might be relevant here

gosuto-inzasheru:
there seems to be a discrepancy in the accounting of this contract, which makes it impossible for even the `approvedStaker` to claim the remaining badger through the existing funcs:
```
>>> geyser = Contract('0xd34C1d3853214bf049B760Ef48A580bfa7A9c8a1')
>>> geyser.balanceOf(geyser.approvedStaker())
86036837300173489
>>> geyser.totalSupply()
86036837300173489
>>> badger = Contract(geyser.rewardsToken())
>>> badger.balanceOf(geyser)
6045500000000000000000
```

only option i see is upgrading, and removing the following line in `recoverERC20`:
```
require(tokenAddress != address(rewardsToken), "Cannot withdraw the rewards tokens");
```

gosuto-inzasheru:
queued: https://etherscan.io/tx/0xeee34de5f5b629487660f93bf20e8479fb12c8996abe5295391b4e4e2924ae53#eventlog

gosuto-inzasheru:
would be nice to amend script so that it also send all ragequited assets in dev msig to trops: #1292

gosuto-inzasheru:
posted exec of timelock at nonce 1104:
```
snapshot result for 0x042B32Ac6b453485e357938bdC38e0340d4b9276:
                   balance_before              balance_after             balance_delta
symbol                                                                                
BADGER  15,692.684648998711160785  21,738.184648998711160785  6,045.500000000000000000
RPL          0.000000000000000000       9.119570178825518772      9.119570178825518772
DAI     47,815.084585421722279017  48,064.208734817731947656    249.124149396009668639
STG          0.000000000000000000     354.968280236651675314    354.968280236651675314
BAL          0.000000000000000000      15.118642899356518613     15.118642899356518613
```

sajanrajdev:
Timelock expired without execution. Needs repost.

sajanrajdev:
Queue was posted once more with 4 day ETA replacing the expired execution tx at Nonce 1104.
[...]

PR #1260, opened 2023-05-09
feat: upgrade slp geyser and recover badger from it

gosuto-inzasheru:
solves #1005 

@sajanrajdev and or @petrovska-petro could you have a look before i go ahead and deploy the new logic?

### DIFF
--- a/scripts/issue/1005/sweep_slp_geyser.py
+++ b/scripts/issue/1005/sweep_slp_geyser.py
@@ -25,14 +25,13 @@ def main(queue=False, sim=False, recover=False):
     trops = GreatApeSafe(r.badger_wallets.treasury_ops_multisig)
     badger = safe.contract(r.treasury_tokens.BADGER)
     geyser = StakingRewardsSignalOnly.at(r.slp_geyser, owner=safe.account)
-    geyser_proxy = interface.IAdminUpgradeabilityProxy(r.slp_geyser, owner=safe.account)
+    geyser_proxy_address = r.slp_geyser
     dev_proxy_admin_address = r.badger_wallets.devProxyAdmin
 
     safe.init_badger()
     trops.take_snapshot([badger])
 
-    # TODO: still need to deploy to mainnet
-    geyser_new_logic = StakingRewardsSignalOnly.deploy({"from": safe.account})
+    geyser_new_logic_address = r.logic.slp_geyser
 
     if queue:
         # queue up upgrade to timelock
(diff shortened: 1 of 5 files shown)


petrovska-petro (commented):
workaround makes sense, it removes the `require` from the `recoverERC20` codebase to allow sweeping badger tokens into admin_role aka devmsig 

diff: https://www.diffchecker.com/m8ZzJuHC/

gosuto-inzasheru:
thanks. will deploy, commit and take pr out of draft

petrovska-petro (commented):
confirm deployment from `ops_deployer8` and contract content identical to the local version in `contracts/` directory ✅ 

when running `brownie run scripts/issue/1005/sweep_slp_geyser queue` there seems to be a minor issue with the conditionals, since also it is trying to directly to execute the upgrade in L55

Screenshot 2023-05-17 at 12 50 15

gosuto-inzasheru:
thanks, solved in c5d9843

scripts/issue/1005/sweep_slp_geyser.py
```diff
@@ -0,0 +1,64 @@
...
+
+    # execute upgrade
+    if sim:
+        timelock = accounts.at(safe.badger.timelock, force=True)
+        proxyAdmin = interface.IProxyAdmin(dev_proxy_admin_address, owner=timelock)
+        proxyAdmin.upgrade(geyser_proxy_address, geyser_new_logic_address)
+
+    # recover badger from geyser and forward from dev msig to trops
+    if recover:
+        safe.badger.execute_timelock("data/badger/timelock/upgrade_slp_geyser/")
```
petrovska-petro, line 57: otherwise will be trigger during the `sim` call

```suggestion
if not sim:
    safe.badger.execute_timelock("data/badger/timelock/upgrade_slp_geyser/")
```

gosuto-inzasheru: thanks. 0b33c41

petrovska-petro (approved):

sajanrajdev (approved):
## Review
- Confirmed there's 6045.5 BADGER sitting on the old Geyser
- Confirmed new logic deloyed by ops_deployer8
- [Diff](https://www.diffchecker.com/renSFTy1/) confirms that only change introduced is the removal of the protected reward token from sweeping 
- Script to be executed in two parts (LGTM):
    - Queue upgrade to new logic
    - Atomically execute upgrade, sweep BADGER and transfer BADGER to trops 
- Note: Script will transfer all BADGER balance from dev to trops which should be fine since Dev doesn't currently hold any BADGER and shouldn't hold anyways.
- Script includes a snapshot and a simulation of the upgrade. Running the simulation confirms the expected behavior:

image

LGTM

Issue #1207, opened 2023-03-29
Avatars Integration (Aura & Convex)


petrovska-petro:
Code for [aura-avatar](https://github.com/Badger-Finance/badger-avatars/tree/main/src/aura) and [convex-avatar](https://github.com/Badger-Finance/badger-avatars/tree/main/src/convex) reached a solid state to enter into a testing phase program

Integration steps:

- [x] Avatar deployments with trops as owner
- [x] Scripting seeding and emerg. wds
- [x] Mirror final setup - test all operations in trops phase in actual mainnet, have multiple aura positions (grab small amounts from treasury if needed)
- [x] Upkeep mngr. upkeep registration
- [x] Update `sellBpsAuraToUsdc` via [set](https://etherscan.io/address/0xc4035528fd975f68efff5d0a726fdc6975a57ba4#code#F16#L240) (#1236) 
- [x] Emerg. wd test run
- [x] Deposit 1/10 funds of aura vault positions in the ecosystems to the aura avatar
- [x] Transfer ownership
- [x] Avatar pid migration (pid: 18 and 19)
- [ ] second independent audit of the avatar contracts
- [ ] deposit of 9/10 of deposits of the funds into the aura avatar 
- [ ] deposit all convex position into the convex avatar as final step

petrovska-petro:
aura avatar: https://etherscan.io/address/0x1ab341f6b71721a25ac092fd41480299843be7fa
convex avatar: https://etherscan.io/address/0x21a940194bdfea81c7802da85e5fcb56dffe0739

petrovska-petro:
replicate locally by running `forge test --gas-report` in [main](https://github.com/Badger-Finance/badger-avatars)

<img width="1007" alt="Screenshot 2023-03-30 at 21 30 14" src="https://user-images.githubusercontent.com/84875062/228836771-02b9efaa-1d6d-42ed-9103-991f296556e7.png">

<img width="946" alt="Screenshot 2023-03-30 at 21 30 35" src="https://user-images.githubusercontent.com/84875062/228836804-d056c6ba-aabe-42af-a283-b8897ce839a7.png">

petrovska-petro:
```
snapshot result for 0x042B32Ac6b453485e357938bdC38e0340d4b9276:
                    balance_before               balance_after               balance_delta
symbol                                                                                    
BADGER  122,982.666934232015207430  110,525.919529387725069126  -12,456.747404844290138304
[...]

PR #1261, opened 2023-05-09
feat: transfer 1/3 tcl into trops for avatar deposits

petrovska-petro:
tackles #1207, point of moving 1/3 of funds

**NOTE:** convex tcl cannot be moved in chunks, since it will require the whole wd of the kekid as a whole not percetanges

### DIFF
--- a/scripts/issue/1207/transfer_tcl_funds.py
+++ b/scripts/issue/1207/transfer_tcl_funds.py
@@ -1,0 +1,78 @@
+from great_ape_safe import GreatApeSafe
+from helpers.addresses import r
+
+# Percetange of tcl to be transfer to recipient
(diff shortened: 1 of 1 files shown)


gosuto-inzasheru (approved):
- `tcl_to_trops`: withdraw a third of the bpts in aura gauges and forward them to trops
- `avatar_deposits`: deposit the bpts in trops to the aura avatar

i am also getting a `GS013` on gas estimation on the last trops tx, not sure why. `skip_preview=True` seems to fix it (although no guarantee for whether this will occur onchain or not)

jalbrekt85 (approved):
withdraws and transfers third lp from vault, deposits into avatar from trops, lgtm

Issue #1253, opened 2023-04-28
unload assets from badger deployer


gosuto-inzasheru:
it is currently holding too many funds.
- [ ] remove from gas station watchlist
- [ ] sweep erc20s to treasury ops (`0x042B32Ac6b453485e357938bdC38e0340d4b9276`)
- [ ] send 3 ether to gas station (`0x04C0205b9D900a49597d9a0345aEAD6D73307C3B`)

gosuto-inzasheru:
posted a gas station update, nonce 528:
```
                                                            addrs      min_bals   min_top_ups
ops_deployer           0xDA25ee226E534d868f0Dd8a459536b03fEE9079b  2.000000e+18  5.000000e+17
ops_deployer2          0xeE8b29AA52dD5fF2559da2C50b1887ADee257556  2.000000e+18  5.000000e+17
ops_deployer3          0x283C857BA940A61828d9F4c09e3fceE2e7aEF3f7  2.000000e+18  5.000000e+17
ops_deployer4          0xef42D748e09A2d9eF89238c053CE0B6f00236210  2.000000e+18  5.000000e+17
ops_deployer8          0x9082b0dD7A72c328833e6461965C9E91Cf59a960  2.000000e+18  5.000000e+17
ops_executor6          0x66496eBB9d848C6A8F19612a6Dd10E09954532D0  1.000000e+18  5.000000e+17
ops_executor13         0x95c1D2014909c04202fa73820B894b45F054F25e  1.000000e+18  5.000000e+17
ops_executor14         0xE78e3E1668D42FfCa767e22e57d7d249e02B5F0e  1.000000e+18  5.000000e+17
ops_executor15         0xaC7B5f4E631b7b5638B9b41d07f1eBED30753f16  1.000000e+18  5.000000e+17
ops_root-validator_v3  0xcA63a374ec5f41dA402dD17b6d90DddFea3d64C4  2.000000e+18  5.000000e+17
ops_botsquad           0xF8dbb94608E72A3C4cEeAB4ad495ac51210a341e  5.000000e+18  5.000000e+17
ops_botsquad_cycle0    0x1a6D6D120a7e3F71B084b4023a518c72F1a93EE9  2.000000e+18  5.000000e+17
upkeep_manager         0x4c02f0160Dc0387b13bCb5e1728C780649E109Ac  5.000000e+18  4.000000e+18
```
[...]

PR #1266, opened 2023-05-10
chore: remove min balance exception for `ops_deployer`

gosuto-inzasheru:
first step in #1253


petrovska-petro (approved):
reduces gas top-up bal for `ops_deployer` to 2ETH ✅

jalbrekt85 (approved):

PR #1272, opened 2023-05-18
feat: sell arbitrary token stuck in processor

gosuto-inzasheru:
this week the weth to aura swap expired. using `buy_aura_or_badger` we can now fix such situations, by passing any erc20 address (weth in this case)

also made a slight change in `cow.py` so that we can actually read api's error response before raising (the requests' response content contains more info that the raise):
```
FEE AND QUOTE PAYLOAD:
{'buyToken': '0xC0c293ce456fF0ED870ADd98a0828Dd4d2903DBF',
 'from': '0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e',
 'kind': 'sell',
 'sellAmountBeforeFee': '3023728579871303680',
 'sellToken': '0xba100000625a3754423978a60c9317c58a424e3D'}

FEE AND QUOTE RESPONSE:
{'errorType': 'SellAmountDoesNotCoverFee', 'description': 'The sell amount for the sell order is lower than the fee.', 'data': {'fee_amount': '0x4f0dbc3ae0b47c00'}}
```

### DIFF
--- a/great_ape_safe/ape_api/cow.py
+++ b/great_ape_safe/ape_api/cow.py
@@ -2,9 +2,9 @@ import json
 import os
 import requests
 from decimal import Decimal
-from pprint import pprint
 
-from brownie import Contract, chain, interface, web3
+from brownie import Contract, chain, interface
+from rich.pretty import pprint
 from rich.prompt import Confirm
 
 from great_ape_safe.ape_api.helpers.coingecko import get_cg_price
@@ -55,7 +55,7 @@ class Cow:
 
         r = requests.post(self.api_url + "quote", json=fee_and_quote_payload)
         print("FEE AND QUOTE RESPONSE:")
-        print(r.json())
+        pprint(r.json())
         print("")
 
         r.raise_for_status()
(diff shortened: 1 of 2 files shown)


petrovska-petro (approved):
changes in the function `buy_aura_or_badger` lgtm and order display for current weth sell looks with genuine values ✅

sajanrajdev (approved):
## Review
- Modifies `buy_aura()` function to allow for buying BADGER or AURA
- Allows for an specific amount of an specific ERC20 token to sell to be passed as parameters
- Better printout of errors for Cow.py

Don't see the point on allowing for any ERC20 to be added as custom sell asset given that both `swapWethForBadger` and swapWethForAURA` only allow for WETH in:
- https://etherscan.io/address/0x8ABD28E4D69bD3953b96dd9ED63533765AdB9965#code#L1345
- https://etherscan.io/address/0x8ABD28E4D69bD3953b96dd9ED63533765AdB9965#code#L1356

Nevertheless, when used with WETH (as intended right now) it works and, given the time constraint on the current round, I will approve.

gosuto-inzasheru:
thanks for dd @sajanrajdev you are right, arbitrary address is not needed. will fix in a later push.

PR #1274, opened 2023-05-19
fix: hardcode weth address

gosuto-inzasheru:
addendum to #1272


petrovska-petro (approved):
defaults to always be selling `weth` as we discussed that was the only one erc20 allowed by the processor to do this ops ✅

PR #1293, opened 2023-05-31
chore: git pull votium and add dai

gosuto-inzasheru:


petrovska-petro (approved):
submodule points to latest commit and adds DAI addr ✅

Issue #1298, opened 2023-06-07
migrate treasury balancer positions to new pids


gosuto-inzasheru:
should all be possible to script atomically, except for the ownership transfer:
> - upgrade the avatar to support the new pids
> - (move ownership of avatar from trops to vault)
> - withdraw badgerwbtc and wbtcdiggaura from old pids for both vault and avatar
> - deposit 100% of both positions into the upgraded avatar, which will deposit both into their new pids

ref: https://github.com/Badger-Finance/Badger-Project-Planning/issues/137#issuecomment-1580884476

related to avatar integration ticket: #1207

gosuto-inzasheru:
note: wait for dd on reason for new pids

also not migrate through avatars, stick to 90/10 split

petrovska-petro:
https://etherscan.io/tx/0xcb9ffe61e4dd2c224418e18077c72103a86e87a7b2b11ad20c4da1c8fd981762

PR #1301, opened 2023-06-09
feat: avatar ownership transfer, prep for aura pid migrations and remaining funds into avatars

petrovska-petro:
Tackles #1298 

Ownership transfer:

```
brownie run scripts/avatars/transfer_ownership 
```

Funds migrations to new pids:

```
brownie run scripts/avatars/migrate_pids

brownie run scripts/avatars/migrate_pids sim
```

### DIFF
--- a/scripts/avatars/migrate_pids.py
+++ b/scripts/avatars/migrate_pids.py
@@ -31,6 +31,13 @@ def main(aura_migration=True, convex_migration=True, treasury_wd=True):
     # boosters
     aura_booster = vault.contract(r.aura.booster)
 
+    # aura rewards contracts
+    rewards_aura_20wbtc80badger_gauge = vault.contract(r.aura.aura_20wbtc80badger_gauge)
+    rewards_aura_40wbtc40digg20gravi_gauge = vault.contract(
+        r.aura.aura_40wbtc40digg20gravi_gauge
+    )
+    rewards_aura_50reth50badger_gauge = vault.contract(r.aura.aura_50reth50badger_gauge)
+
     # avatars
     aura_avatar = vault.contract(r.avatars.aura)
     convex_avatar = vault.contract(r.avatars.convex)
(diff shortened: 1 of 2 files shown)


scripts/avatars/migrate_pids.py
```diff
@@ -0,0 +1,109 @@
...
+        kek_id = staking_contract.lockedStakes(private_vault, 0)[0]
+        private_vault.withdrawLocked(kek_id)
+
+        vault.print_snapshot()
+
+    if aura_migration:
+        total_assets = list(aura_avatar.totalAssets())
+
+        pids = list(aura_avatar.getPids())
+        if PID_AURA_EXCEPTION != 0:
```
gosuto-inzasheru, line 68: what is the purpose of this? checking a constant versus a hardcoded value?

petrovska-petro: as touched on our meeting, modify lang and add onchain checks on 2945209862e13fab95358a879da13b7593e38d71

jalbrekt85 (approved):
new pids lgtm, withdraws from old positions and migrates to new aura/convex pids

gosuto-inzasheru (commented):
- withdraw treasury's bpt and wcvx positions to treasury
- withdraw aura avatar's bpt positions to treasury
- deposit all bpt positions to aura avatar
- deposit cvx position to convex avatar

is there a reason you chose to not withdraw the wcvx position from the convex avatar before redepositing? wouldnt it be better to have a single locked position versus two?

petrovska-petro:
> is there a reason you chose to not withdraw the wcvx position from the convex avatar before redepositing? wouldnt it be better to have a single locked position versus two?

@gosuto-inzasheru since the last arg is `True` means that will deposit additional funds in existing kekid. [ref](https://github.com/Badger-Finance/badger-avatars/blob/main/src/convex/ConvexAvatarMultiToken.sol#L325-L331)

PR #1309, opened 2023-06-15
Vault v1.5 deposits pausing helper

sajanrajdev:
Simple script to pause deposits on a v1.5 vault from dev Multisig as needed.

Use as:
```
brownie run scripts/warroom/pause_deposits.py main vault_address
```


gosuto-inzasheru (commented):
lgtm but wouldnt we also want a way to batch these into one tx?

petrovska-petro:
> lgtm but wouldnt we also want a way to batch these into one tx?

agree that it should be batch ideally, maybe add an array of target vaults that we could modify as needed in the future if any. cc: @sajanrajdev

sajanrajdev:
Good point! Will modify to allow for batching @petrovska-petro @gosuto-inzasheru

petrovska-petro (approved):

gosuto-inzasheru (approved):

PR #1319, opened 2023-06-28
chore: incentives claim week 26

gosuto-inzasheru:
we encountered some bugs in prod this week.

this should fix it, if not will push more. draft pr for now.

### DIFF
--- a/scripts/badger/process_bribes_graviaura.py
+++ b/scripts/badger/process_bribes_graviaura.py
@@ -24,9 +24,9 @@ SAFE.init_cow(prod=COW_PROD)
 PROCESSOR = SAFE.contract(r.aura_bribes_processor, from_explorer=True)
 
 # tokens involved during processing
-WETH = interface.IWETH9(r.treasury_tokens.WETH, owner=SAFE.account)
-BADGER = interface.ERC20(r.treasury_tokens.BADGER, owner=SAFE.account)
-AURA = interface.ERC20(r.treasury_tokens.AURA, owner=SAFE.account)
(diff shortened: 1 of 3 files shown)


sajanrajdev (commented):
- Introduced tweaked Ape_safe main library with the Gnosis Safe endpoints adjusted (except for Fantom) and format set to our repo's standards: https://www.diffchecker.com/UwHRbqNI/ LGTM
- Added Votium's round data. LGTM
- Sets the bveCVX slippage for the orders to 2% instead of 4%. LGTM

scripts/badger/process_bribes_graviaura.py
```diff
@@ -24,9 +24,9 @@
 PROCESSOR = SAFE.contract(r.aura_bribes_processor, from_explorer=True)
 
 # tokens involved during processing
-WETH = interface.IWETH9(r.treasury_tokens.WETH, owner=SAFE.account)
```
sajanrajdev, line 27: Why doing this? These tokens are assigned for their usage after the bribes processing but not expected to come in as bribes themselves. It can occur but its is not the intention for the most. Doesn't matter much tbh just curious on the reasons for the change.

jalbrekt85 (approved):
https://www.diffchecker.com/UwHRbqNI/ looks good, diffs are mainly just formatting

scripts/badger/process_bribes_graviaura.py
gosuto-inzasheru: because the last couple of weeks i have needed to add new tokens there to ignore, since the autovoter is picking up dust. it is easier to do so with all of them coming from that claimable list

sajanrajdev (approved):

Issue #1321, opened 2023-06-28
Set keeper on graviAURA to AdressZero


sajanrajdev:
GraviAURA won't be harvested nor earned anymore due to deprecation. Even though these functions have been disabled for it on the Web2 side, it was decided to remove the keeper on it as a precaution to avoid accidental runs of the function (calling these would trigger a relock).

sajanrajdev:
https://etherscan.io/tx/0xc3b3b58dbb264405c1586566e6af3a624894d4292acce41357ed48a6a3881d91

PR #1322, opened 2023-06-28
feat: remove graviAura's keeper

sajanrajdev:
Tackles #1321

### DIFF
--- a/scripts/issue/1321/remove_graviAURA_keeper.py
+++ b/scripts/issue/1321/remove_graviAURA_keeper.py
@@ -1,0 +1,15 @@
+from helpers.addresses import r
+from great_ape_safe import GreatApeSafe
+from helpers.constants import AddressZero
+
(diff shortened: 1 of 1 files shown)


gosuto-inzasheru:
note: merging main to prevent calling old gnosis safe api endpoints

gosuto-inzasheru (approved):
does what it says on the box:

prevents calling `earn` on graviaura via keeper access control. still remains callable by gov

petrovska-petro (approved):
Sets to null addr the keeper on gravi, blocking the action ✅

sajanrajdev:
> does what it says on the box:
> 
> prevents calling `earn` on graviaura via keeper access control. still remains callable by gov

Correct, it is ok for actions to be callable from Multis for this purpose. The fear here is that a earn is triggered accidentally by a misconfiguration on the bot. This action removes that issue.

Issue #1331, opened 2023-07-05
sign wbtc dao txid 3 from techops


gosuto-inzasheru:
> Hi everyone, we’re almost done!
>
> Victor at Kyber has initiated the last txn to move the wBTC from the Big DAO to the distribution Gnosis safe here: https://etherscan.io/tx/0xb70fd02346ddae47f620041610c26d607fcdcf318fde29faf87e9acd4811a1f8
> 
> Please confirm txId 3 from your individual keys. Once this is complete, we can air drop the lost wBTC with 4 signers instead of 8.

gosuto-inzasheru:
ref: https://github.com/WrappedBTC/DAO/issues/17

gosuto-inzasheru:
https://etherscan.io/tx/0xc26e5314d6220204f8e35c528b9274b2738d2133281488f9813f5029033baac6

PR #1332, opened 2023-07-05
feat: confirm txid 3 on wbtcdao msig

gosuto-inzasheru:
solution for #1331


petrovska-petro (approved):
confirms tx to transfer wbtc to the new msig gnosis safe ✅

jalbrekt85 (approved):

Issue #1335, opened 2023-07-07
Adapt Hidden Hands processing scripts to their v2


sajanrajdev:
Hidden Hands just released a [V2](https://mirror.xyz/0xE90c74145245B498fef924fAdC7bb34253c7cF90/uMF43OIe5k3hF-D6ZeLUWkoD-iOAv-HNYzmhgT8kkcg) of their contracts. The flows and scripts for claiming and and processing graviAURA's bribes must be fixed. These are the new addresses: 
```
Mainnet

RewardDistributor | 0xa9b08B4CeEC1EF29EdEC7F9C94583270337D6416
RewardSwapper | 0x4177b6f4C38A2ee94606e10049691393643b80ae
RewardHarvester | 0xd23aa7EdF42CD3Fc4CD391faAbc0c207B1c86542
RewardHarvesterClaim | 0x4666f33E4fbE6C8AA9671bD2AEdc46d641451F85
BribeVault | 0xE00fe722e5bE7ad45b1A16066E431E47Df476CeC
BribeMarket (implementation) | 0x0F7983eb3D146477B6c7d4C5aAB427D3340DB38E
BribeFactory | 0xB49f6D281b84D3B29203BA38360a7cb97780B2f1

BribeMarket for BALANCER | 0x45Bc37b18E73A42A4a826357a8348cDC042cCBBc
BribeMarket for AURA | 0xcBf242F20D183B4116C22Dd5e441b9aE15b0d35A
BribeMarket for FRAX | 0x2694214a4EA6A2C9A9427AA12251705B67e86960
BribeMarket for BUNNI | 0x175b01eF7CEa8615B1D9485BC827378EA5D50C86
BribeMarket for FLOOR_DAO | 0x19302c4ec309b3913e8bE41025DCF66A562BDa42
BribeMarket for RIBBON | 0x97A45f52e5EdD775923a8DE3f5373D41DA5dDa73
BribeMarket for PENDLE | 0xf12e79c95d3C13e2ab4f8222d698DAA53Df17aD8
```

sajanrajdev:
# Impact Analysis on Bribes Processing

## GraviAURA Strategy
- Interaction with HH occurs in the following function: https://github.com/Badger-Finance/vested-aura/blob/f05bfb22cb82ecaa57cd957ce7fe530e900e8384/contracts/MyStrategy.sol#L386
- Function takes the rewards distributor as a parameter (Great because this address changed)
[...]

PR #1341, opened 2023-07-14
Fix HH Bribing and Claiming scripts

sajanrajdev:
Tackles https://github.com/Badger-Finance/badger-multisig/issues/1335


petrovska-petro (commented):
updates addrs lgtm and interfaces, abi changes are minimal in the methods ✅ 

@saj, could we make more flexible the `periods` and `maxTokensPerVote` values instead of hardcoded (by default `_maxTokensPerVote=0` & `_periods=1` 

i think the most value on their upgrade lies into those params. while we may do one per period and accepting whatever amounts of votes/token, i will say that having that dynamic as value accepted on the cli will be useful from the influence strat pov

sajanrajdev:
> updates addrs lgtm and interfaces, abi changes are minimal in the methods ✅
> 
> @saj, could we make more flexible the `periods` and `maxTokensPerVote` values instead of hardcoded (by default `_maxTokensPerVote=0` & `_periods=1`
> 
> i think the most value on their upgrade lies into those params. while we may do one per period and accepting whatever amounts of votes/token, i will say that having that dynamic as value accepted on the cli will be useful from the influence strat pov

Good call, will do so.

sajanrajdev:
@petrovska-petro, introduced the ability to choose the period and bribes per period via function parameters. In addition, I added a check for 0 bribes on BADGER/wBTC for Balancer as it is the case for this round, attempting to bribe with 0 reverts as expected.

petrovska-petro (approved):
allows flex for the case that influence pod decides to be more granular with rounds or expected votes per incentive amts ✅ 

addrs in book are correct ✅ 

checks if when passing concatenated incentives for bal hh market any of them are `0` to skipped them ✅

scripts/badger/bribe_ecosystems.py
```diff
@@ -59,6 +59,8 @@ def main(
     badger_bribe_in_votium=0,
     badger_bribe_in_frax=0,
     badger_bribe_in_bunni=0,  # NOTE: dollar denominated. Badger calculation is done internaly
+    max_tokens_per_vote=0,  # Maximum amount of incentives to be used per round (Hidden Hands V2)
+    periods=1,  # Rounds to be covered by the incentives deposited (Hidden Hands V2)
```
gosuto-inzasheru, line 63: in cases such as these, it is usually better to add new params to the end of the signature, to maintain backwards compability.

gosuto-inzasheru (approved):
confirmed onchain with command `brownie run badger/bribe_ecosystems main 0 0,16000`. runs successfully

Issue #1364, opened 2023-08-14
Buy 200K BADGER with up to 500K USDT


sajanrajdev:
As per TCD 35

petrovska-petro:
1st order filled: https://explorer.cow.fi/orders/0x0b9c16045eddab19bc3f5f588860e1722110652790313b78541768740e047199d0a7a8b98957b9cd3cfb9c0425abe44551158e9e64dbf533?tab=overview

sajanrajdev:
Laddered orders filled: 

![Image](https://github.com/Badger-Finance/badger-multisig/assets/25463788/b5a8c553-003d-472b-a3c5-a5f2a7b93edb)

https://explorer.cow.fi/orders/0xb5ffb11f4dc524aa74e79a90185dafa1a7f1c6b5f47bda82fd7992cd0d39de7cd0a7a8b98957b9cd3cfb9c0425abe44551158e9e64e757d6
https://explorer.cow.fi/orders/0x98fe1dd535ade1ae2445ffb431c597110fe727dc6c899107927e04aec8019e7ad0a7a8b98957b9cd3cfb9c0425abe44551158e9e64e757d8
https://explorer.cow.fi/orders/0xeeb6812b731858b01ad05b22d5ab143e4385c60dbe54835a05d72db067ec5febd0a7a8b98957b9cd3cfb9c0425abe44551158e9e64e757da

sajanrajdev:
Result: 45,596.765 + 47,285.7247 + 47,549.5357 + 64,151.1085 = 204583.1339 BADGER for 450K USDT

PR #1368, opened 2023-08-16
Badger buyback script

sajanrajdev:
Relates to #1364

### DIFF
--- a/scripts/treasury/badger_buyback_ladder_order.py
+++ b/scripts/treasury/badger_buyback_ladder_order.py
@@ -1,0 +1,49 @@
+from decimal import Decimal
+from math import log, floor
+from great_ape_safe import GreatApeSafe
+from helpers.addresses import r
+from great_ape_safe.ape_api.helpers.coingecko import get_cg_price
+
+
+SAFE = GreatApeSafe(r.badger_wallets.treasury_vault_multisig)
+USDT = SAFE.contract(r.treasury_tokens.USDT)
+BADGER = SAFE.contract(r.treasury_tokens.BADGER)
+
+COW_PROD = True
+
+MAX_USDT = 500_000
+MAX_BADGER = 200_000
+LEG_AMOUNT = 100_000
+DEADLINE = 60 * 60 * 24 * 7  # gives one week to sell
+DECIMALS = USDT.decimals()
+
+
+def main(usdt_remaining=(MAX_USDT * 0.7), badger_bought=64151.1085, approve=True):
+    usdt_remaining_int = int(usdt_remaining)
+    SAFE.init_cow(prod=COW_PROD)
+    if approve:
+        amount_to_approve = (
+            int(usdt_remaining_int / LEG_AMOUNT) * LEG_AMOUNT * 10 ** DECIMALS
+        )
+        SAFE.cow.allow_relayer(USDT, amount_to_approve)
+
+    cg_price = get_cg_price(BADGER.address)
+    total_badger = float(badger_bought)
+    for i in range(int(usdt_remaining_int / LEG_AMOUNT)):
+        coefficient = 1.05 + (i * 0.025)
+        total_badger = total_badger + (
+            LEG_AMOUNT / (cg_price / coefficient)
+        )  # Rough estimate, does not take fees into account
+        if total_badger < MAX_BADGER:
+            SAFE.cow.market_sell(
+                asset_sell=USDT,
+                asset_buy=BADGER,
+                mantissa_sell=LEG_AMOUNT * 10 ** DECIMALS,
+                deadline=DEADLINE,
+                coef=coefficient,
+                destination=r.badger_wallets.treasury_vault_multisig,
+            )
+        else:
+            print("Buy amount would exceed buyback total!")
+
+    SAFE.post_safe_tx()


sajanrajdev:
@petrovska-petro @gosuto-inzasheru @rayeaster, tried to introduce a programatic way of limiting the buyback at 200K BADGER but given that it does not take into account fees it may not be as accurate as we need. I will be happy to remove that whole thing and we can handle the accounting manually as needed. Also, this script only applies for buying with chunks of 100K as spec'ed, I think that in practice we may need to do an additional manual purchase to compensate after 1 week.

petrovska-petro (commented):
orders looks good:

```
{
│   'sellToken': '0xdAC17F958D2ee523a2206206994597C13D831ec7',
│   'buyToken': '0x3472A5A71965499acd81997a54BBA8D852C6E53d',
│   'receiver': '0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e',
│   'sellAmount': '100000000000',
│   'buyAmount': '45003235037720961089536',
│   'validTo': 1692798188,
│   'appData': '0x2ab7ad46094d32adf4db9196c6a0a1de32cb030485e47bd5cb7a92b7306a7b99',
│   'feeAmount': '0',
│   'kind': 'sell',
│   'partiallyFillable': False,
│   'sellTokenBalance': 'erc20',
│   'buyTokenBalance': 'erc20',
│   'signingScheme': 'presign',
│   'signature': '0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e',
│   'from': '0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e'
}

{
│   'sellToken': '0xdAC17F958D2ee523a2206206994597C13D831ec7',
│   'buyToken': '0x3472A5A71965499acd81997a54BBA8D852C6E53d',
│   'receiver': '0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e',
│   'sellAmount': '100000000000',
│   'buyAmount': '46075258997682020024320',
│   'validTo': 1692798193,
│   'appData': '0x2ab7ad46094d32adf4db9196c6a0a1de32cb030485e47bd5cb7a92b7306a7b99',
│   'feeAmount': '0',
│   'kind': 'sell',
│   'partiallyFillable': False,
│   'sellTokenBalance': 'erc20',
│   'buyTokenBalance': 'erc20',
│   'signingScheme': 'presign',
│   'signature': '0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e',
│   'from': '0xD0A7A8B98957b9CD3cFB9c0425AbE44551158e9e'
}
```

but is able only to generate and post two orders, meaning two 100k usdt lots, will only used 350k from 500k and bought max of ~152k badger. or i am missing something?

btw, as discussed i think fee could be taken into account via [ref](https://github.com/Badger-Finance/badger-multisig/blob/main/great_ape_safe/ape_api/cow.py#L40), but may complicate accounting

ref to my test orders: https://explorer.cow.fi/address/0xd0a7a8b98957b9cd3cfb9c0425abe44551158e9e. first 2 top ones

sajanrajdev:
@petrovska-petro, seems like if in practice we set the `badger_bought` amount to 0, it will post the three tx of the ladder as shown in the first three txs on the OrderBook. The expected amounts of these three orders plus the already purchased should be slightly below 200K BADGER. I think we should do this, and use the limiting feature for next week given that it is unlikely that the three orders will go through:
image

petrovska-petro:
> @petrovska-petro, seems like if in practice we set the `badger_bought` amount to 0, it will post the three tx of the ladder as shown in the first three txs on the OrderBook. The expected amounts of these three orders plus the already purchased should be slightly below 200K BADGER. I think we should do this, and use the limiting feature for next week given that it is unlikely that the three orders will go through: image

@sajanrajdev yes, that looks more accurate towards the final goal imo re stable spending and badger bought. will expect at least 3 orders (otherwise, misses the one at -10%)

scripts/treasury/badger_buyback_ladder_order.py
```diff
@@ -0,0 +1,49 @@
...
+DEADLINE = 60 * 60 * 24 * 7  # gives one week to sell
+DECIMALS = USDT.decimals()
+
+
+def main(usdt_remaining=(MAX_USDT * 0.7), badger_bought=64151.1085, approve=True):
+    usdt_remaining_int = int(usdt_remaining)
+    SAFE.init_cow(prod=COW_PROD)
+    if approve:
+        amount_to_approve = (
+            int(usdt_remaining_int / LEG_AMOUNT) * LEG_AMOUNT * 10 ** DECIMALS
```
rayeaster, line 26: could use a variable to store the calculation result of `int(usdt_remaining_int / LEG_AMOUNT)` since it is used later again. similar for `10 ** DECIMALS`

scripts/treasury/badger_buyback_ladder_order.py
```diff
@@ -0,0 +1,49 @@
...
+        )
+        SAFE.cow.allow_relayer(USDT, amount_to_approve)
+
+    cg_price = get_cg_price(BADGER.address)
+    total_badger = float(badger_bought)
+    for i in range(int(usdt_remaining_int / LEG_AMOUNT)):
+        coefficient = 1.05 + (i * 0.025)
```
rayeaster, line 33: could use a constant variable for this slippage coefficient?

rayeaster (approved):
LGTM to add a series of market-buy orders for BADGER token via CowSwap with increasing expected bought outputs.

petrovska-petro (approved):
g2g assuming will be running with `purchase_amt=0` ✅

Issue #1277, opened 2023-05-24
Minting upgradability


petrovska-petro:
Currently the minting capabilities are restricted to the controller, which is governed by the `dev_msig`, which under current conditions this could occur atomically and its status presents an opportunity to upgrade the flow considering the following approaches (but not limited to):

- Establish a timelock with at least **x** amount of days as delay behind this action
- Consider enforcing governance to a greater group of `msig` for consensus of proceeding with the action. Example: devmsig, treasury and badger council should be align to queue the tx in the timelock 
- Introduce a mint cap per action, including a setter, which again will be behind timelock. For example: at once no more than **y** tokens can be minted (400k, 500k…y)

This will achieve a greater standard of security around this action matching our practices across-the-board.

petrovska-petro:
Refs:

- [transferOwnership](https://etherscan.io/address/0xeba06639750badcdf5ee100d72f3c93f2d1918a7#code#L177). It could transfer it to timelock
- for `queueTransaction` in the timelock the `admin` could be a *msig* where the owners are the folllowing existing msigs: `dev_msig`, `treasury_msig` and a new one `badger_council_msig`
- cap could be introduce one L above [this](https://etherscan.io/address/0xeba06639750badcdf5ee100d72f3c93f2d1918a7#code#L364) with a `require` statement check

gosuto-inzasheru:
conclusions from discussion with @dapp-whisperer:
- yes, moving minting func to the timelock makes sense. security policy currently in place for the badger system should also apply to the badger token itself
- broadening the scope of what governance is (ie making treasury also a signer on dev msig, or something similar) is a good convo to have but outside of the scope of the ticket. convo should take place though
- introducing a mint cap again is outside the scope. introducing such a cap would imply a consensus on inflation rate, which needs a bip first or some form of agreement between councils.

tldr: scope of this ticket is therefore reduced to:
- [ ] transfer ownership to timelock, so that minting cannot be done atomically

sajanrajdev:
@petrovska-petro, given the reach of the contract in matter, Dapp requested that we include a bit more of testing diligence in the script before executing. Specifically, we wants confirmation via a test that the controller and the BADGER token will continue to work properly after the change. 

Adding a simulation of a few of the actions on both the controller and the token within the same script (scripts/issue/1277/mint_controller_admin_update.py) will be enough.

PR #1395, opened 2023-09-08
feat: add tests after ownership is transfer to assert behaviours

petrovska-petro:
Tackles latest reccoms in #1277

### DIFF
--- a/scripts/issue/1277/mint_controller_admin_update.py
+++ b/scripts/issue/1277/mint_controller_admin_update.py
@@ -1,17 +1,60 @@
+from brownie import accounts
+
 from great_ape_safe import GreatApeSafe
 from helpers.addresses import r
 
 
-def main():
+def main(verify_future_state=True):
     safe = GreatApeSafe(r.badger_wallets.dev_multisig)
 
     # contracts
     controller = safe.contract(r.GatedMiniMeController)
     timelock_gov = safe.contract(r.governance_timelock)
 
+    # tkns
+    badger = safe.contract(r.treasury_tokens.BADGER)
+    dai = safe.contract(r.treasury_tokens.DAI)
+
     controller.transferOwnership(timelock_gov)
 
-    # guardian setup
-    timelock_gov.setGuardian(r.badger_wallets.techops_multisig)
+    if verify_future_state:
+        tl_gov = accounts.at(r.governance_timelock, force=True)
+
+        # mint
+        controller.mint(100e18, {"from": tl_gov})
+
+        assert badger.balanceOf(tl_gov) == 100e18
+
+        # claim tokens
+        bal_in_minime = dai.balanceOf(badger)
+
+        controller.claimTokens(r.treasury_tokens.DAI, {"from": tl_gov})
+
+        assert dai.balanceOf(tl_gov) == bal_in_minime
+
+        # disable minting
+        controller.disableMinting({"from": tl_gov})
+
+        assert controller.mintingEnabled() == False
+
+        # transfer ownership
+        controller.transferOwnership(
+            r.badger_wallets.treasury_vault_multisig, {"from": tl_gov}
+        )
+
+        assert controller.owner() == r.badger_wallets.treasury_vault_multisig
+
+        # erc20 basic samples
+        holder_one = accounts.at(r.badger_wallets.treasury_vault_multisig, force=True)
+        holder_two = accounts.at(r.badger_wallets.treasury_ops_multisig, force=True)
+
+        # check basic transfer and balance updates
+        holder_one_bal_before = badger.balanceOf(holder_one)
+        holder_two_bal_before = badger.balanceOf(holder_two)
+
+        badger.transfer(holder_two, 5e18, {"from": holder_one})
+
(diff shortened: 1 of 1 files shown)


scripts/issue/1277/mint_controller_admin_update.py
```diff
@@ -1,17 +1,60 @@
...
 
-    # guardian setup
-    timelock_gov.setGuardian(r.badger_wallets.techops_multisig)
+    if verify_future_state:
+        tl_gov = accounts.at(r.governance_timelock, force=True)
+
+        # mint
+        controller.mint(100e18, {"from": tl_gov})
+
+        assert badger.balanceOf(tl_gov) == 100e18
```
sajanrajdev, line 26: This actually may have revealed an issue. How would we be able to transfer the minted BADGER out of the [Timelock](https://etherscan.io/address/0x21CF9b77F88Adf8F8C98d7E33Fe601DC57bC0893#code)?  The contract doesn't seem to have a sweep function.

petrovska-petro: that's by design of the controller to send over to the owner, but the timelock could exec on itself a queue and exec of a transfer elsewhere, right?

sajanrajdev: No because the Timelock doesn't own itself and, most importantly, doesn't have a transfer ERC20 method (or sweep to be more specific). This could work if it had a method to perform aleatory actions on itself as we added to the avatars but in this case it is constraint to the methods it was designed with.

petrovska-petro: pushed another piece of test to help visualise the msg im trying to communicate in latest commit

please pull it locally and inspect the timelock (queuing & exec flow), thats how i envision is doable

wdyt?

sajanrajdev: Ah, you are absoluteley right, my bad. Of course the timelock can interact with functions on the Badger token just like any other actor, given that these are permissionless. Great idea, thanks for the POC. I don't think it is an issue but just worth noticing then that any BADGER minting flows will require two Timelock transactions: One for minting and one for transfering.

sajanrajdev (approved):
Tests showed that all governance functions will be executable from the Timelock and that the BADGER token will remain fully functional (as expected).

NOTE: The BADGER minting processes will require two separate timelock operations:
- Minting
- Transfering the minted tokens out of the Timelock and into their destination 

CC @dapp-whisperer, Petro has now pushed the tests for this operation as per your request. Please take a look and note the above.

sajanrajdev:
These tests revealed the need for a redesign of the approach as the Timelock would now become the recipient of any minted BADGER which isn't an intended behavior. Nevertheless, there is value on preserving these tests and I would say we move for a merge of them.

@rayeaster, would you mind taking a look?

scripts/issue/1277/mint_controller_admin_update.py
```diff
@@ -1,17 +1,85 @@
...
+
     controller.transferOwnership(timelock_gov)
 
-    # guardian setup
-    timelock_gov.setGuardian(r.badger_wallets.techops_multisig)
+    if verify_future_state:
+        tl_gov = accounts.at(r.governance_timelock, force=True)
+
+        # mint
+        controller.mint(100e18, {"from": tl_gov})
```
rayeaster, line 26: could use a variable to hold the magic value of `100e18`

scripts/issue/1277/mint_controller_admin_update.py
```diff
@@ -1,17 +1,85 @@
...
+        holder_one = accounts.at(r.badger_wallets.treasury_vault_multisig, force=True)
+        holder_two = accounts.at(r.badger_wallets.treasury_ops_multisig, force=True)
+
+        # check basic transfer and balance updates
+        holder_one_bal_before = badger.balanceOf(holder_one)
+        holder_two_bal_before = badger.balanceOf(holder_two)
+
+        badger.transfer(holder_two, 5e18, {"from": holder_one})
+
+        assert badger.balanceOf(holder_two) == holder_two_bal_before + 5e18
```
rayeaster, line 79: could use a variable to hold the magic value of `5e18`

scripts/issue/1277/mint_controller_admin_update.py
```diff
@@ -1,17 +1,85 @@
...
+        assert badger.balanceOf(tl_gov) == 100e18
+
+        # ensure tokens can be move from tl
+        badgerBalBefore = badger.balanceOf(safe)
+
+        calldata = encode_abi(
+            ["address", "uint256"],
+            [safe.address, int(100e18)],
+        )
+        eta = chain.time() + int(3 * 60 * 60 * 24)
```
rayeaster, line 37: could use a variable to hold the magic value of `3 * 60 * 60 * 24`

rayeaster (approved):
LGTM with minor comment

sajanrajdev:
Good comments with regards the magic values @rayeaster. Given that these tests are more informative at this point and the operation is not planned anymore, I will go ahead and merge as is.

Issue #1325, opened 2023-06-29
week 39: settle budgets onchain


gosuto-inzasheru:
prev #1185

Prereqs:
- Sell influence assets on vault and trops
- Balance stables (sell DAI for USDC)
- Transfer stables from trops to Vault

Operation:
- Transfer BADGER and USDC to Payments and Trops for incentives and expenses as per accounting

petrovska-petro:
dai order for usdc filled: https://explorer.cow.fi/orders/0x6f913c6c80b40a158bc0653b237abbef2ef2e705bfa6a6c9a81a8ab8171a880dd0a7a8b98957b9cd3cfb9c0425abe44551158e9e6516f29f?tab=overview

sajanrajdev:
Stables transfer to Vault: https://etherscan.io/tx/0x2f5783da8b9382b3d8c1f145564f10d7802876529b92b60cea9eff4986e054c4

sajanrajdev:
Settling: https://etherscan.io/tx/0x432aee5f45b179ea0dbd4b36bddce84a1da665726081e225d5f5b0a35ad9f603

PR #1407, opened 2023-09-27
feat: dai -> usdc order to reach budget req

petrovska-petro:
tackles #1325 

```
brownie run scripts/issue/1325/sell_stables_q4_budget
```


rayeaster (approved):
LGTM to have configured DAI sold  for USDC. maybe we could use a bit tight sllipge (like `0.995`?) for this stablecoin swap

sajanrajdev (approved):
Simple enough: sells the required amount of DAI for USDC in agreement with the quarterly stablecoin rebalancing. Agree with @rayeaster on tightening the slippage tolerance when posting as it will likely be routed through stable swaps.

LGTM

PR #1433, opened 2023-10-19
Update AutonomousDripper.sol: Optimized gas costs

Aryan9592:
This PR aims for two main changes:
1. Optimized `for` loop: Well, the way these 'for' loops were structured in this smart contract were quite normal but kind of taking more gas than it needed to execute. Changing `i++` to `++i`, and using the `unchecked` block alone can play a lot of effect. For more info, refer this -> https://gist.github.com/grGred/9bab8b9bad0cd42fc23d4e31e7347144#for-loops-improvement. This convo might help -> https://github.com/lurk-lab/solidity-verifier/pull/31
2. Changes `uint256 count = 0;` to `uint256 count;`: Usually, default values of unsigned integers are 0 itself, thus there's no need to initialize any unsigned integer to 0. This same change takes effect in case of `bool dripped = false;` to `bool dripped;`

### DIFF
--- a/contracts/AutonomousDripper.sol
+++ b/contracts/AutonomousDripper.sol
@@ -64,12 +64,16 @@ contract AutonomousDripper is VestingWallet, KeeperCompatibleInterface, Confirme
      */
     function _getAssetsHeld() internal view returns (address[] memory) {
         address[] memory _assetsHeld = new address[](assetsWatchlist.length);
-        uint256 count = 0;
-        for (uint idx = 0; idx < assetsWatchlist.length; idx++) {
+        uint256 count;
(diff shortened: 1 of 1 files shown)


gosuto-inzasheru:
thanks @Aryan9592! can i ask you to move this pr to this repo: https://github.com/Badger-Finance/badger-infra/blob/main/contracts/refillers/AutonomousDripper.sol ?

it is where the original source of the contract lives :)

Aryan9592:
Okay

sajanrajdev:
Closing as it will be handled within the Dripper's repo.

Issue #1443, opened 2023-10-31
Exeute on BIP 102


sajanrajdev:
Ref: https://snapshot.org/#/badgerdao.eth/proposal/0xa24a37037ea31e34e01e5f8ba5afd8b40ebec5bd9c1ab00085e6a35df287bbed

sajanrajdev:
https://etherscan.io/tx/0x0e8c569b5f0966d76bcdf1403099b6571783535dc9ed6d368c6d77403eab284f

Issue #1380, opened 2023-08-29
Execute on TCD-36


sajanrajdev:
Treasury vlAURA should be divest for wETH as it unlocks. 

Schedule:
![image](https://github.com/Badger-Finance/badger-multisig/assets/25463788/3f215f3e-fd2b-4d5e-99c5-d9da39be253e)

## Execution plan
On a weekly basis, starting now and until November 30th, do the following atomically:
- Call `vlAURA.processExpiredLocks(False)`
- Sell the AURA unlocked for wETH through Cowswap

sajanrajdev:
## 24 August Unlock
- https://etherscan.io/tx/0xac14ef36fb81b12767f19c552fa1e1d440fe53455fb4d2b1febf0ac56a9e9765
- https://explorer.cow.fi/orders/0xdcdca84965afb18c8eaf7d4beed587898052b74f7bc7dba0d74dfee5413d4e69a9ed98b5fb8428d68664f3c5027c62a10d45826b64f23104?tab=overview

sajanrajdev:
## 7 Sept Unlock
- https://etherscan.io/tx/0xdbb399082ccca09988603fbd4532dc7aadebb7a16072fa8746dcd48c554591ae
- https://explorer.cow.fi/orders/0xf009e1042fa0f5b6cad9b77e0e4e7717030c78149c16a208c931628187314569a9ed98b5fb8428d68664f3c5027c62a10d45826b64fb6d82?tab=overview

sajanrajdev:
## 14 Sept Unlock
- https://etherscan.io/tx/0xadd5945edb35800ba82dcc8ffb76be2623987c28b3b7242b3d44f5ab762ca9b2
- https://explorer.cow.fi/orders/0x41beed51585891af585901fb55a9aae278b27440328a5b63be923397a0b4c5a6a9ed98b5fb8428d68664f3c5027c62a10d45826b65047337?tab=overview

sajanrajdev:
## 21 Sept Unlock
- https://etherscan.io/tx/0xde5cd10017aa8a5142047e7332c2d1ed6717639f8143b3ee1f4f5ac91a9af6de
- https://explorer.cow.fi/orders/0xd0d50b05e126a62500308be5dae44d1c97e678ef70a6698d323c3e65c6e9ae0da9ed98b5fb8428d68664f3c5027c62a10d45826b650db11a?tab=overview

sajanrajdev:
[...]

PR #1444, opened 2023-10-31
feat: aura lps divestment

sajanrajdev:
Tackles #1443 and #1380 and TCD1 (withdraw BADGER/wBTC LP)

Atomically:
- Withdraws BADGER and DIGG LP from Aura -> Balancer and unrwaps them
- Doesn't claim rewards from gauges (to be claimed afterwards on TCL strats)
- Sells graviAURA from Digg BPT for wETH as per TCD 36
- Burns DIGG (can't transfer to AddressZero so transfering to address(dead)

Run with
```
brownie run scripts/issue/1443/digg_lp_divestment.py
```

### DIFF
--- a/scripts/issue/1443/digg_lp_divestment.py
+++ b/scripts/issue/1443/digg_lp_divestment.py
@@ -8,6 +8,7 @@ DEADLINE = 60 * 60 * 24
 
 BURN_ADDRESS = "0x000000000000000000000000000000000000dEaD"
 
+
 def main(unwrap="true"):
     vault = GreatApeSafe(r.badger_wallets.treasury_vault_multisig)
     vault.init_aura()
--- a/scripts/issue/1443/digg_lp_divestment.py
+++ b/scripts/issue/1443/digg_lp_divestment.py
@@ -1,6 +1,5 @@
 from great_ape_safe import GreatApeSafe
 from helpers.addresses import r
-from helpers.constants import AddressZero
 
 # slippages
 COEF = 0.95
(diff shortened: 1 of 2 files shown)


scripts/issue/1443/digg_lp_divestment.py
```diff
@@ -0,0 +1,52 @@
...
+    # Selling graviAURA for wETH as per TCD 36
+    graviaura.withdrawAll()
+    vault.cow.market_sell(
+        aura, weth, aura.balanceOf(vault), deadline=DEADLINE, coef=COEF
+    )
+
+    # Burning all Digg holdings by transfering to AddressZero
+    vault_digg_balance = digg.balanceOf(vault)
+    digg.transfer(BURN_ADDRESS, vault_digg_balance)
+    assert digg.balanceOf(vault) == 0
```
rayeaster, line 50: do we need to check the balance of dead address?

scripts/issue/1443/digg_lp_divestment.py
```diff
@@ -0,0 +1,52 @@
...
+        if unwrap == "true":
+            # Not claiming rewards as these will be processed separately on TCL strat
+            vault.balancer.unstake_all_and_withdraw_all(
+                pool=bpt, unstake=False, pool_type="Weighted", claim=False
+            )
+
+    # Selling graviAURA for wETH as per TCD 36
+    graviaura.withdrawAll()
+    vault.cow.market_sell(
+        aura, weth, aura.balanceOf(vault), deadline=DEADLINE, coef=COEF
```
rayeaster, line 44: sell `aura` or `graviAURA`?

sajanrajdev: It is the graviAURA but in the previous line it is withdrawn into AURA. I guess a good fix would be to use the exact amount withdrawn.

scripts/issue/1443/digg_lp_divestment.py
sajanrajdev: addressed

petrovska-petro (approved):
steps in the scripts wraps up the bip decision for digg appropiatly, burning in addr(dead), wd and unwrap badger/wbtc due to treasury decisions and sells aura following divesting tcd

asserts are on place to ensure right accounting ✅ 

only concern i had in the logic is the swap size and the liq profile of aura (due to locks) that the price impact looks to be >3%

Screenshot 2023-11-01 at 11 33 55

simple workarounds could be either split it on multi-orders bins (25% each of size) or accept the price impact

rayeaster (approved):
LGTM

Issue #1437, opened 2023-10-26
Updates: quest v2 & HH liquis support


petrovska-petro:
Integrate:

- [x] quest v2 [0x1B921DBD13A280ee14BA6361c1196EB72aaa094e](https://etherscan.io/address/0x1B921DBD13A280ee14BA6361c1196EB72aaa094e)
- [x] HH liquis [0xA44C4c85eB01f02d8cEa9BC416174Bb13d13Aa72](https://etherscan.io/address/0xA44C4c85eB01f02d8cEa9BC416174Bb13d13Aa72)

petrovska-petro:
finalised in #1397

PR #1449, opened 2023-11-07
feat: include hh liquis support

petrovska-petro:
tackles partially #1437 to enable hidden hands support for liquis incoming epoch

address [ref](https://docs.redacted.finance/resources/contract-addresses/hidden-hand-contracts)

NOTE: bare in mind that if the epoch is off, it may trigger a revert in the tx during your local test, to see whole successful flow either ping a block where epoch was running before timestamp expired or wait till new one is mined


rayeaster (approved):
LGTM

sajanrajdev (approved):
- Confirmed Liquis Briber address
- Sets the use of HH as default for Liquis bribes (Palading can still be used)
- Confirmed valid generated [proposal](https://api.hiddenhand.finance/proposal/liquis); 
image

Script LGTM, we just need BADGER whitelisted as incentive on the briber (on the works).

LGTM

PR #1475, opened 2023-12-21
docs: update readme

sajanrajdev:
Minor updates to the docs


petrovska-petro (approved):
written updates informs better of latest shape of diff msigs and their scope, removed one signer from techops scope ✅

ICEBADGER (changes requested):
I think there is an extra "s" in line 83 after "required"

Otherwise looks good.

ICEBADGER (approved):

petrovska-petro (approved):

Issue #1480, opened 2024-01-09
week 3: incentives Paladin Bunni 


petrovska-petro:
prev #1478

petrovska-petro:
https://docs.google.com/spreadsheets/d/136NQDQP5WrSSGUd3gM0xeaHlCwDPsoamwypmAMqqm1w/edit#gid=1343908694

petrovska-petro:
```
snapshot result for 0x042B32Ac6b453485e357938bdC38e0340d4b9276:
                    balance_before              balance_after                balance_delta
symbol                                                                                    
LIQ     205,513.894202895980276762  60,766.915532762616920424  -144,746.978670133363356338
```

petrovska-petro:
https://dashboard.tenderly.co/testing-ape/project/simulator/02d52d9b-b4fc-4605-a816-4e30d4c58de7

petrovska-petro:
https://dashboard.tenderly.co/testing-ape/project/simulator/02d52d9b-b4fc-4605-a816-4e30d4c58de7

petrovska-petro:
https://etherscan.io/tx/0x789501c6f7b8d1ccfc2f34f7726aa430ea1de5a06cd585dcc6332caa266321eb

PR #1488, opened 2024-01-19
feat: allow $liq incentives via Paladin for Bunni marketplace

petrovska-petro:
allows incentivising Bunni marketplace #1480 via Paladin

in this case it is required to use "blacklisting" vote type, since it requires to blacklist the vlLIQ voter proxy

run with following params: `brownie run scripts/badger/bribe_ecosystems`

```
    badger_bribe_in_aura=0,
    badger_bribe_in_balancer=0,
    badger_bribe_in_votium=0,
    badger_bribe_in_frax=0,
    badger_bribe_in_bunni=1600,  # NOTE: dollar denominated. Badger calculation is done internaly
    max_tokens_per_vote=0,  # Maximum amount of incentives to be used per round (Hidden Hands V2)
    periods=1,  # Rounds to be covered by the incentives deposited (Hidden Hands V2)
    badger_bribe_in_liquis=0,  # NOTE: dollar denominated. Badger calculation is done internaly, the incentive gets process via Paladin
    duration_paladin_quest=1,  # Duration (in number of periods) of the Quest
    reward_per_vote_liquis=0.059,  # Amount of reward per vlLIQ
    liquis_incentive_in_paladin=True,  # Indicates if the incentive is going to be post in paladin or HH
    is_governance_incentive_token=False,  # Indicates if the incentive is going to be governance token ($badger) or different. Applicable only in Paladin.
    aura_proposal_id=None,
    convex_proposal_id=None,
```

### DIFF
--- a/scripts/badger/bribe_ecosystems.py
+++ b/scripts/badger/bribe_ecosystems.py
@@ -42,6 +42,7 @@ MAX_BPS = 10_000
 
 # vote type. ref: https://etherscan.io/address/0x1b921dbd13a280ee14ba6361c1196eb72aaa094e#code#F6#L16
 PALADIN_VOTE_TYPE_WHITELISTING = 2
+PALADIN_VOTE_TYPE_BLACKLISTING = 1
 
 # close type. ref: https://etherscan.io/address/0x1b921dbd13a280ee14ba6361c1196eb72aaa094e#code#F6#L27
 PALADIN_CLOSE_TYPE_ROLLOVER = 1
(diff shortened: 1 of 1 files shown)


sajanrajdev (approved):
- Allows for the creation of Paladin quests for bribing directly to Bunni voters with both BADGER and LIQ. 
- Accomplishes this in a similar fashion than when creating quests for veLIQ voters but blacklisting the LIQ voting proxy from the quest instead.
- Ability to bribe through hidden hand is optionally preserved.

LGTM

PR #1533, opened 2024-05-16
Feat/restitution2.0

sajanrajdev:
- Adds the newly approved user to redeposit into remBADGER along with the other one.
- Removes BADGER seeding from the simulation
- Pushed current upgrade data (expired so will need to be replaced) for reference
- Adds scripts to enable and disable deposits to the production bremBADGER

NOTE: Both users have transfer their BADGER to the [devMultisig](https://app.safe.global/transactions/history?safe=eth:0xB65cef03b9B89f99517643226d76e286ee999e77) so both addresses and amounts can be confirmed from:
- User 1: https://etherscan.io/tx/0x72fe5089064aca2dc9589de758dc88daa5bcc89ea5af0a0c6c8d885eb2d270fd
- User 2; https://etherscan.io/tx/0x20212321883f0435ea2869315f99d0935770f8e14fab36d8d570bbe323c3abd6 and https://etherscan.io/tx/0xd021b82de2159f78be3e7b62c3b8de5fed524404f81920fb02e74675668c3938


wtj2021 (approved):
looks good

mrbasado (approved):

Issue #1558, opened 2024-08-21
Swap Badger TechOps members for new technical contributors


sajanrajdev:
Swap:
- `0xE78e3E1668D42FfCa767e22e57d7d249e02B5F0e` -> `0xE2C5B2008d9cc8F8E1FDa8552f7df63Af1f747f8` 
- `0x8938bf50d1a3736bdA413510688834540858dAEA` -> `0xDA82F543613f90deA718c46D02Ca15e05e88e4aC`

- [ ] Mainnet
- [ ] Arbitrum
- [ ] Polygon
- [ ] Optimism
- [ ] Fantom

sajanrajdev:
Ethereum: https://etherscan.io/tx/0x07d5b791936545d36182e348f87d15f060d2636747f46528e4ce786cc46b0d12

PR #1561, opened 2024-08-29
docs: adapt readme

sajanrajdev:
Relates to #1558


ICEBADGER (approved):

petrovska-petro (approved):
swaps looks correct based on techops tx ✅

